PR #6, opened 2019-09-06
Update media queries

SiAdcock:
A more consistent syntax, with a smaller API surface area:

```js
// before
${mobileLandscape} {

}

// after
${from.mobileLandscape} {

}
```


packages/foundations/media-queries.ts
```diff
@@ -1,62 +1,140 @@
...
+	leftCol: {
+		until: {
+			wide: minWidthMaxWidth(breakpointMap.leftCol, breakpointMap.wide)
+		}
+	}
 };
 
-export { mq };
+// e.g. from.*
+from.mobileMedium.toString = () => minWidth(breakpointMap.mobileMedium);
```
sndrs: ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️

PR #45, opened 2019-10-18
Darken and Lighten

SiAdcock:
Functions that lighten and darken a hex colour by a specified percentage.

This introduces the `utilities` package of functionality that make it easier for web developers to apply some of the design patterns specified by the design system.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,4 @@ node_modules
 *.log
 dist
 .cache
-
+.DS_Store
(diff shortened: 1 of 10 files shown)


SiAdcock:
On reflection this PR opens a Pandora's box that allows designers and developers to mix their own colours, potentially leading to all manner of inconsistency. I'll discuss with Akemi how we can use or adapt the existing palette to accommodate button hover colours.

PR #50, opened 2019-10-22
Added above and below adverbs

oliverlloyd:
This PR introduces 2 new adverbs to `media-queries.ts`.

## `above`
This is semantically similar to saying `from.desktop` but that query returns an object at the moment so the only way to state 'for all widths above desktop' is just `${desktop}`. For me, `${above.desktop}` is more explicit and so more readable, especially for newcomers to the library.

## `below`
The expression `${below.wide}` is identical to `${until.wide}` but I'm including the adverb `below` as I feel there should be a partner for `above` and that having it makes the library more intuitive. This is subjective but I greatly prefer apis that let me do the same thing using interchangeable expressions.

## Other benefits
The real motivation for this PR isn't language choice. I have a function that accepts a prop `breakpoint: 'desktop' | 'leftCol' // | etc.`. I want to use this prop directly in my css but currently when I do this I have:

```
    // The below case
    ${until[breakpoint]}

    // The above case
    ${breakpoint}
```

This works fine for `until` but in the above case I'm just dropping a string into my css, not the actual rule.

I'd much rather have:

```
    // The below case
    ${below[breakpoint]}

    // The above case
    ${above[breakpoint]}
```

### DIFF
--- a/packages/foundations/media-queries.ts
+++ b/packages/foundations/media-queries.ts
@@ -18,8 +18,17 @@ const maxWidth = (until: number): string =>
 const minWidthMaxWidth = (from: number, until: number): string =>
 	`@media (min-width: ${`${from}px`}) and (max-width: ${`${until - 1}px`})`
 
-// e.g. until.*
-const until = {
+const minWidths = {
+	mobileMedium: minWidth(breakpointMap.mobileMedium),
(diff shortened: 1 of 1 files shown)


oliverlloyd:
Closing in favour of #51

PR #236, opened 2020-02-24
REM Space

JamieB-gu:

Adds a `rem` version of the `space` object.

**Note:** I've generated the new object using `reduce` and `as` (there may be a downside to this). Alternatively the new object could be created as a literal, with a type annotation to make sure it has the same keys as `space`. What do you think @SiAdcock?

## What does this change?

- Added a `remSpace` object derived from the `space` object.

### DIFF
--- a/src/core/foundations/src/space.ts
+++ b/src/core/foundations/src/space.ts
@@ -13,9 +13,17 @@ const space = {
 	24: _space[8],
 }
 
-const remSpace = Object.entries(space).reduce(
-	(acc, [k, v]) => ({ ...acc, [k]: `${v / rootPixelFontSize}rem` }),
-	{} as Record<keyof (typeof space), string>,
-)
+const pxToRem = (px: number): string => `${px / rootPixelFontSize}rem`
+
+const remSpace: { [K in keyof (typeof space)]: string } = {
+	1: pxToRem(space[1]),
+	2: pxToRem(space[2]),
+	3: pxToRem(space[3]),
+	5: pxToRem(space[5]),
+	6: pxToRem(space[6]),
+	9: pxToRem(space[9]),
+	12: pxToRem(space[12]),
+	24: pxToRem(space[24]),
+}
 
 export { space, remSpace }
--- a/src/core/foundations/src/space.ts
+++ b/src/core/foundations/src/space.ts
@@ -1,5 +1,7 @@
 import { space as _space } from "./theme"
 
+const rootPixelFontSize = 16
+
 const space = {
 	1: _space[1],
 	2: _space[2],
@@ -11,4 +13,9 @@ const space = {
 	24: _space[8],
 }
 
-export { space }
+const remSpace = Object.entries(space).reduce(
+	(acc, [k, v]) => ({ ...acc, [k]: `${v / rootPixelFontSize}rem` }),
+	{} as Record<keyof (typeof space), string>,
+)
+
+export { space, remSpace }
--- a/src/core/foundations/src/space.test.ts
+++ b/src/core/foundations/src/space.test.ts
@@ -1,0 +1,4 @@
+import { space, remSpace } from "./space"
+
+it("should provide a rem equivalent of the pixel space", () =>
+	expect(remSpace[2]).toBe(`${space[2] / 16}rem`))


src/core/foundations/src/space.ts
```diff
@@ -1,5 +1,7 @@
 import { space as _space } from "./theme"
 
+const rootPixelFontSize = 16
```
webb04: Should use this in the test?

SiAdcock:
Thanks @JamieB-gu it's a decent proposal. There's a few things I think we'll need to figure out. In a way this PR acts as a forcing function to get me to think about them.

The assumption of root pixel font size is one that is unavoidable, as far as I can tell. It would be useful to extract this to somewhere common, to share with other foundations (e.g. [typography](https://github.com/guardian/source/blob/master/src/core/foundations/src/typography/index.ts#L23)), and indeed tests as @webb04 points out.  

Including the unit in the name of the exported object (i.e. `remSpace`) is also a useful convention, when a unit is included. Unfortunately, we're not doing that for font sizes ([example](https://github.com/guardian/source/blob/master/src/core/foundations/src/typography/index.ts#L25-L52)) so we might need to introduce a breaking change there. Nothing to be done about that for now, I'll find out what the impact would be.

I'd like to standardise on a common type for all our scale-based foundations, if possible. There's a [proposal here](https://github.com/system-ui/theme-specification/issues/1#issuecomment-523127428) that could be a good starting point, although I'm not sure how useful it is?

Regarding the use of `reduce` with `as`... An alternative would be to use `Object.entries`, then `map` to iterate and calculate, then [`Object.fromEntries`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries) to go back to an object again? Not sure if we can make this type-safe, but I'm all for avoiding `reduce` wherever possible.

Thanks for your help, I'm up for chatting if it's easier than using comments 😄

JamieB-gu:
> There's a proposal here that could be a good starting point

Yeah that sounds like a good idea!

I wonder whether an approach like this might work:

```ts
const enum Space {
    One,
    Two,
    Three,
    Five,
    Six,
    Nine,
    Twelve,
    TwentyFour,	
}

type Spaces<A> = {
    [S in Space]: A
}

const spaces: Spaces<string> = {
    [Space.One]: '3rem',
    [Space.Two]: '3rem',
    [Space.Three]: '3rem',
    [Space.Five]: '3rem',
    [Space.Six]: '3rem',
    [Space.Nine]: '3rem',
    [Space.Twelve]: '3rem',
    [Space.TwentyFour]: '3rem',
}
```

> Not sure if we can make this type-safe

I just had a quick go and I think you hit similar problems in that you lose the literal type, i.e. it becomes `Record<string, string>`. Perhaps it's better to write it out as a literal and rely on the type checker to keep them in sync:

```ts
const pxToRem = (px: number): string => `${px / rootPixelFontSize}rem`

const remSpace: { [K in (keyof (typeof space))]: string } = {
    1: pxToRem(space[1]),
    2: pxToRem(space[2]),
    3: pxToRem(space[3]),
    5: pxToRem(space[5]),
    6: pxToRem(space[6]),
    9: pxToRem(space[9]),
    12: pxToRem(space[12]),
    24: pxToRem(space[24]),
}
```

JamieB-gu:
> There's a proposal here that could be a good starting point

Yeah that sounds like a good idea!

I wonder whether an approach like this might work:

```ts
const enum Space {
    One,
    Two,
    Three,
    Five,
    Six,
    Nine,
    Twelve,
    TwentyFour,	
}

type Spaces<A> = {
    [S in Space]: A
}

const spaces: Spaces<string> = {
    [Space.One]: '3rem',
    [Space.Two]: '3rem',
    [Space.Three]: '3rem',
    [Space.Five]: '3rem',
    [Space.Six]: '3rem',
    [Space.Nine]: '3rem',
    [Space.Twelve]: '3rem',
    [Space.TwentyFour]: '3rem',
}
```

> Not sure if we can make this type-safe

I just had a quick go and I think you hit similar problems in that you lose the literal type, i.e. it becomes `Record<string, string>`. Perhaps it's better to write it out as a literal and rely on the type checker to keep them in sync:

```ts
const pxToRem = (px: number): string => `${px / rootPixelFontSize}rem`

const remSpace: { [K in (keyof (typeof space))]: string } = {
    1: pxToRem(space[1]),
    2: pxToRem(space[2]),
    3: pxToRem(space[3]),
    5: pxToRem(space[5]),
    6: pxToRem(space[6]),
    9: pxToRem(space[9]),
    12: pxToRem(space[12]),
    24: pxToRem(space[24]),
}
```

SiAdcock:
Great thanks @JamieB-gu. Yes, I think you're right about writing it out as a literal, and relying on the type checker to enforce consistency. It's more human-readable, and just as type-safe.

I'll have a think about the type definition for `Space`, ideally I'd like something generic and flexible that can also be used for stuff like `Size`, `FontSize`, maybe even `Breakpoint`. We can address that outside of this PR.

In the interests of getting this merged, could you make the change you suggested that gets rid of `reduce` and we'll have separate conversations about types and naming?

Thanks so much for leading the way here! 🎉

PR #291, opened 2020-03-23
Simpler Headline Font Name

JamieB-gu:

I noticed a bug that we've introduced in apps-rendering since switching over to the design system typography APIs: our headline fonts are falling back to Georgia. It might be hard to spot this if your machine has the fonts installed locally, as that will mean it never attempts to load the webfonts. To observe this behaviour you can disable the Guardian fonts locally (in Font Book on macOS).

### Why?

Long story short, naming things is hard. We provide the headline font under the name "Guardian Headline" in our `@font-face` declarations, but the design system specifies "GH Guardian Headline" in `font-family` declarations. These don't match, therefore the browser falls back to the next font specified in `font-family`, in this case Georgia.

### Solution

There are a few ways to solve this:

1. Specify "GH Guardian Headline" in apps-rendering
2. Add "Guardian Headline" to the `font-family` declaration in the design system
3. Replace "GH Guardian Headline" with "Guardian Headline" in the `font-family` declaration in the design system

I've gone for (2) because it's less likely to introduce breaking changes for other projects than (3), although (3) would be my preferred option if possible. I haven't gone for (1) because I think the name should be "Guardian Headline". That said, I don't know why the "GH" prefix exists, so I may be wrong and the best solution is (1) because "GH Guardian Headline" is the correct name. @paperboyo what do you think?

cc @webb04 @gtrufitt 

## What does this change?

- Adds a new name for the headline font: "Guardian Headline".

### DIFF
--- a/src/core/foundations/src/theme.ts
+++ b/src/core/foundations/src/theme.ts
@@ -27,7 +27,7 @@ const fontSizes = [12, 15, 17, 20, 24, 28, 34, 42, 50, 70]
 const fonts = {
 	titlepiece: "GT Guardian Titlepiece, Georgia, serif",
 	headlineSerif:
-		"GH Guardian Headline, Guardian Egyptian Web, Georgia, serif",
+		"Guardian Headline, GH Guardian Headline, Guardian Egyptian Web, Georgia, serif",
 	bodySerif:
 		"GuardianTextEgyptian, Guardian Text Egyptian Web, Georgia, serif",
 	bodySans:


paperboyo:
@JamieB-gu Thanks for tagging me! Unfortunately, option 1 is the only correct one:

image

I think the filenames are already synchronised.

JamieB-gu:
> I think the filenames are already synchronised

Based on the meeting we had the other week, did we want to set the filenames to whatever we want now, and stick with that? Or do we say the PostScript names are set in stone, so we should align all the filenames with those?

paperboyo:
> the PostScript names are set in stone, so we should align all the filenames with those

This. And I already sent webfonts too, so all is in place. Fontastique will generate Android-kosher filenames just for this (silly here) platform, otherwise PS names=fielnames all over. Thx!

SiAdcock:
I concur with @paperboyo on this. 

@JamieB-gu is it feasible to change the `font-family` names in apps-rendering?

JamieB-gu:
> @JamieB-gu is it feasible to change the `font-family` names in apps-rendering?

Yep! Just opened a PR to do that, linked above ^.

@paperboyo what does the "GH" refer to?

paperboyo:
Just an innocent pleonasm, I suppose. Was needed for print (coz all there have them) and we wanted print and digital to match.

JamieB-gu:
Fixed in apps-rendering instead, see the above linked PR.

PR #297, opened 2020-03-26
Font Style Normal

### LINKED ISSUE #292: Font Style Normal
https://github.com/guardian/source/blob/d42e9821c4bd1d7a7ac2a01a00be7b5821987117/src/core/foundations/src/typography/fs.ts#L30

As `italic` is an optional parameter for fonts, there are three states it can be in: `true`, `false` and `undefined`. Should `font-style` be set explicitly to `normal` when `italic` is set explicitly to `false`? In essence, the following:

<dl>
  <dt>true</dt>
  <dd><code>font-style: italic;</code></dd>
  <dt>false</dt>
  <dd><code>font-style: normal;</code></dd>
  <dt>undefined</dt>
  <dd><code>""</code></dd>
</dl>

This is useful for times when you want to explicitly set `font-style: normal;`. For example, when overriding the user agent italic styles for [`<address>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/address).

JamieB-gu:

Introduced a way to get `font-style: normal;` by explicitly passing `false` to the `italic` property of font options. Fixes #292.

## What does this change?

- Made default fontStyle `undefined` for all fonts
- Updated 'italic' property to be `true | false | undefined`
- Created function to generate font style
- Added tests for font-style changes

### DIFF
--- a/src/core/foundations/src/typography/api.ts
+++ b/src/core/foundations/src/typography/api.ts
@@ -15,7 +15,7 @@ type TitlepieceFunctions = {
 const titlepieceDefaults = {
 	lineHeight: "tight",
 	fontWeight: "bold",
-	italic: false,
+	italic: undefined,
 	unit: "rem",
 }
 const titlepieceFs = fs("titlepiece")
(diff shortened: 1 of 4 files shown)


JamieB-gu:
After some discussion of API changes, will replace this with an alternative implementation.

Issue #281, opened 2020-03-18
Text input field should have blue outline when active


SiAdcock:
**Describe the bug**
Text input field should have blue outline when active

**Screenshots**

Live example:

<img width="536" alt="Screenshot 2020-03-18 at 21 30 10" src="https://user-images.githubusercontent.com/5931528/77009397-c10ebe80-695f-11ea-8dcc-99bb7299099d.png">

(https://www.theguardian.design/2a1e5182b/p/097455-text-input-field/b/050445)

In Figma:

<img width="468" alt="Screenshot 2020-03-18 at 21 30 20" src="https://user-images.githubusercontent.com/5931528/77009398-c1a75500-695f-11ea-98e1-d6b5a4f2dd9b.png">

(https://www.figma.com/file/qY0C090LN6LDAqdzmV2FsA/Text-inputs?node-id=575%3A164)

PR #298, opened 2020-03-27
add blue border to text input when active

webb04:
Add a blue border colour to text input when active. This shows when the input has been clicked/tapped.

Fixes: https://github.com/guardian/source/issues/281

### Screenshots

**Before**
Screen Shot 2020-03-27 at 10 23 22

**After**
Screen Shot 2020-03-27 at 10 22 42

### DIFF
--- a/src/core/components/text-input/styles.ts
+++ b/src/core/components/text-input/styles.ts
@@ -24,6 +24,10 @@ export const textInput = ({
 	border: 2px solid ${textInput.border};
 	padding: 0 ${space[2]}px;
 
+	&:active {
+		border: 2px solid ${textInput.borderActive};
+	}
+
 	&:focus {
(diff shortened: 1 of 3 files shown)


SiAdcock:
Oh wow, thanks so much for doing this @webb04!

Can I make 2 requests:

1) Rather than using `inputChecked`, could you add another entry to the [default border palette](https://github.com/guardian/source/blob/master/src/core/foundations/src/palette/border/default.ts) called `inputActive`? It's possible that colours for checked and active could diverge, so I think it would be best to pre-empt this rather than create a hidden coupling.
2) Could you create commits so that there are changes to only one package per commit? This makes maintenance a bit eeasier. I should create some better contributer instructions somewhere!

Let me know if you need any more info

PR #301, opened 2020-03-31
Strict Foundations

JamieB-gu:

Applies the [`strict` flag](https://www.typescriptlang.org/docs/handbook/compiler-options.html) to the TS compiler. It's a useful option to make the type checker do more work for you. The principle reason I'm enabling it here is to get `strictNullChecks`, which prevents `null` and `undefined` being part of every type.

## What does this change?

- Added `strict` to the `tsconfig.json` for foundations


JamieB-gu:
Sounds good! We do some of that [here](https://github.com/guardian/apps-rendering/tree/master/config).

PR #302, opened 2020-03-31
Editorial lines

SiAdcock:

The GuardianLines component represents a signature pattern that is used across editorial applications. If only there was somewhere central we could keep components like this so everyone could share them 🤔 

I know! Let's put it in Source!*

I'm putting out this request for comments to seee whether you all think that:

- Source is a good place for this particular component
- components like this could live here in the future
- we'll be happy to install shared editorial components from NPM/Yarn
- the name is meaningful: `@guardian/src-ed-lines`
- Editorial component version numbers should match core component version numbers (currently all version numbers in Source are [kept in sync](https://www.theguardian.design/2a1e5182b/p/876251-getting-started/t/7213f7))
- the directory structure under `src/editorial` looks okay 

There's a few things that are definitely not right:

- The Pillar type will need to be moved to [guardian/types](https://github.com/guardian/types/)
- The `neutralBorder` function should probably live somewhere more shareable
- The new component is not hooked up to the build, CI and scripted release processes
- The stories are added to the core storybook (this might be okay but it would benefit from some further thought)

<sub>*disclaimer: this was actually @andre1050's idea, I don't want any credit or blame when it all goes south</sub>

## What does this change?

-   Adds a Lines component to Source Editorial

### Screenshots

**straight**
Screenshot 2020-03-31 at 22 18 48

**squiggly**
Screenshot 2020-03-31 at 22 18 39

**dotted**
Screenshot 2020-03-31 at 22 18 59


SiAdcock:
Thanks @oliverlloyd!

> But doesn't `@guardian/lines` sound so much sweeter?

It really does! I'm envisioning a world in which we have dozens of core `src-` components, dozens of `src-ed-` components, dozens of `src-rr-` components, maybe even dozens of `src-experimental-` components. I'd like to explore a naming convention that describes the status of a component, its ownership, its place in the design systems constellation.

> I'm not sure it's wise to ask a large group of developers what they think about folder structures

Haha good point! At this point, I'm not looking for consensus, only opinions 😄

package.json
```diff
@@ -26,7 +26,8 @@
 	"private": true,
 	"workspaces": [
 		"src/core/*",
-		"src/core/components/*"
+		"src/core/components/*",
+		"src/editorial/web/components/*"
```
JamieB-gu, line 30: Why does `editorial` have a `web` subset but `core` doesn't?

src/editorial/web/components/lines/.babelrc.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+	presets: [
+		"@babel/preset-env",
```
JamieB-gu, line 3: It's possible that this is documented elsewhere and I've completely missed it, but how does `source` decide on the browser support list?

src/editorial/web/components/lines/styles.ts
```diff
@@ -0,0 +1,41 @@
...
+		to bottom,
+		${neutralBorder(pillar)},
+		${neutralBorder(pillar)} 1px,
+		transparent 1px,
+		transparent 4px
+	);
+	background-repeat: repeat-x;
+	background-position: top;
+	background-size: 1px ${count === 4 ? "13px" : "29px"};
+	height: ${count === 4 ? "13px" : "29px"};
```
JamieB-gu, line 26: Does it make sense to use `rem` here for scaling purposes?

src/editorial/web/index.d.ts
```diff
@@ -0,0 +1,6 @@
+// Pillars are used for styling
+// RealPillars have Pillar palette colours
+// FakePillars allow us to make modifications to style based on rules outside of the pillar of an article
+type RealPillars = "news" | "opinion" | "sport" | "culture" | "lifestyle"
+type FakePillars = "labs"
+type Pillar = RealPillars | FakePillars
```
JamieB-gu, line 6: I'm going to make an argument for the existence of [`ColourScheme`](https://github.com/guardian/types/blob/master/ColourScheme.ts) again here 🙂

nicl:
👍 this is useful for us!

>  I'd like to explore a naming convention that describes the status of a component, its ownership, its place in the design systems constellation.

I'm curious how ownership makes sense here? As surely all teams are owners/users dictate ownership in practical terms?

package.json
SiAdcock: No real reason, just haven't got around to it 😄

src/editorial/web/components/lines/.babelrc.js
SiAdcock: There's nothing explicit yet. At the moment, it's whatever `preset-env` and ES5 gives us.

src/editorial/web/components/lines/styles.ts
SiAdcock: Yeah, happy to tinker with this for sure. Perhaps we can address the implementation details of this component in a future PR?

src/editorial/web/index.d.ts
SiAdcock: I'm going to happily agree with you 🙂

SiAdcock:
@andre1050 

> I'm not sure the pillar should be a mandatory prop, seeing as all pillars will render a neutral border except "labs".

I have to agree with you here, `Pillar` seems to be doing a bit of heavy lifting here that would ideally be moved up to the component tree. This is something that can be fixed with a [theme](https://www.theguardian.design/2a1e5182b/p/42c3a2-theming) or by [overriding the CSS](https://www.theguardian.design/2a1e5182b/p/2619e3-overriding-styles) in another way.

SiAdcock:
@nicl 
> I'm curious how ownership makes sense here? As surely all teams are owners/users dictate ownership in practical terms?

The suggestion is that this component makes sense as an editorial component, and as such should be owned by the on-platform teams (dotcom, apps, editions). Other teams (reader revenue, tools) may want to use it in the future, at which point we could consider making this a core component, owned by the design system team (or all teams, depending on how that plays out!)

Does that make sense? Chat me if you'd like to discuss, I don't wanna be talking at cross purposes here 😄

SiAdcock:
@JamieB-gu 

> It would be interesting to consider GH Packages as well

@oliverlloyd looked into GH Packages and encountered a bit of friction. We decided that we would need to make the move as an organisation, rather than publishing some packages to GH and some to NPM.

> Does this synchronisation indicate compatibility/interoperability (something along the lines of babel's versions sync)?

Exactly this

> Does a requirement for JS interop preclude the use of TS-specific features like enums?

Possibly? I think the limitation with `const enum` is a Babel limitation rather than a blocker for JS interoperability. I've probably missed a ton of context here, so this should go on an agenda for a larger discussion.

src/editorial/web/components/lines/.babelrc.js
JamieB-gu: It's an interesting one because we're targeting approximately three different environments (in the editorial stream):

- Node (10 for DCR, 12 for AR)
- Dotcom client-side (Safari, Firefox, Chrome, Edge, Internet Explorer etc.)
- Apps client-side (recent Safari and Chrome - webviews)

The support requirements for apps and dotcom on the client in particular are quite different. We _have_ to transpile down to the level of IE support for dotcom, but will that result in shipping redundant polyfills to the apps?

src/editorial/web/components/lines/styles.ts
JamieB-gu: Worth tracking in an issue?

src/editorial/web/components/lines/.babelrc.js
SiAdcock: Good points. Happy to pull this into a larger discussion around browser support.

src/editorial/web/components/lines/styles.ts
SiAdcock: Sure thing! Feel free to raise one, maybe after this is merged

JamieB-gu:
> I think the limitation with `const enum` is a Babel limitation rather than a blocker for JS interoperability.

Oh yeah, good point; I had a chat with @gtrufitt about this recently WRT dotcom-rendering. I think the Babel limitation comes from the fact that `preset-typescript` [doesn't do any type-checking](https://babeljs.io/docs/en/babel-plugin-transform-typescript#caveats), it does something like transform it rather than compiling with TSC. I guess this could be solved by running TSC before Babel, or using `ts-loader` when working with Webpack (the approach taken by apps-rendering).

Would it not be a blocker for JS interop? `const enum`s become numbers in the output JavaScript. If you were to do something like:

```ts
const enum FontStyle {
  Normal,
  Italic,
}
```

you'd then consume the API like this in TS:

```ts
body.medium({ fontStyle: FontStyle.Italic })
```

but I think you'd have to do this if you were working in pure JS:

```js
body.medium({ fontStyle: 1 })
```

SiAdcock:
> const enums become numbers in the output JavaScript. 

Oh. Yes. If we're using `const enums` in the public API, this is certainly going to cause a problem. We should probably avoid it for now, but instigate a wider discussion (I'm collecting discussion points, BTW!)

SiAdcock:
I've made the following changes:

- removed dependency on Pillar. The border will always be `neutral[86]` [unless overridden](https://www.theguardian.design/2a1e5182b/p/2619e3-overriding-styles), so we'd have to handle the `labs` case in the consuming application
- replaced the dynamic styling for `count` with a performant composition-based approach
- added more stories

I'll merge this afternoon then we can go to work on polishing it up ready for release and, ultimately, installing in some applications 🎉

PR #303, opened 2020-04-01
Font-Style API - Specify Normal

### LINKED ISSUE #292: Font Style Normal

JamieB-gu:

Introduces a new API for `font-style` via a new `FontStyle` type. If specified, `FontStyle` can be `"normal"` or `"italic"`. If unspecified it won't be set in the resulting CSS.

Closes #292, and introduces a more explicit API than #297.

**Note:** I've added the `Option` type and put it in what I think is probably not the right place. If you would prefer not to create this new type, or would like it to exist in a different place at least, please let me know.

## What does this change?

- Migrated from specifying `italic` as a boolean to custom type `FontStyle`
- Added `Option` type
- Updated tests

### DIFF
--- a/src/core/foundations/src/typography/fs.ts
+++ b/src/core/foundations/src/typography/fs.ts
@@ -7,9 +7,17 @@ import {
 } from "./data"
 import { Fs, FontWeightDefinition, FontStyle, Option } from "./types"
 
+// Exists to support deprecated API, should be removed at some point
+const legacyItalic = (
+	font: FontWeightDefinition | undefined,
+	italic: boolean,
+): Option<FontStyle> =>
+	italic && font !== undefined && font.hasItalic ? "italic" : null
+
 function getFontStyle(
 	font: FontWeightDefinition | undefined,
 	fontStyle: Option<FontStyle>,
+	italic: boolean,
 ): Option<FontStyle> {
 	switch (fontStyle) {
 		case "italic":
@@ -18,13 +26,13 @@ function getFontStyle(
 			return "normal"
 		case null:
 		default:
-			return null
+			return legacyItalic(font, italic)
 	}
 }
 
 export const fs: Fs = category => (
 	level,
-	{ lineHeight, fontWeight, fontStyle, unit },
+	{ lineHeight, fontWeight, italic, fontStyle, unit },
 ) => {
 	const fontFamilyValue = fontMapping[category]
 	const fontSizeValue =
@@ -40,7 +48,7 @@ export const fs: Fs = category => (
 	// font is unavailable
 	const requestedFont = availableFonts[category][fontWeight]
 	const fontWeightValue = requestedFont ? fontWeightMapping[fontWeight] : ""
-	const fontStyleValue = getFontStyle(requestedFont, fontStyle)
+	const fontStyleValue = getFontStyle(requestedFont, fontStyle, italic)
 
 	return Object.assign(
 		{
--- a/src/core/foundations/src/typography/fs.ts
+++ b/src/core/foundations/src/typography/fs.ts
@@ -5,11 +5,26 @@ import {
 	fontWeightMapping,
 	availableFonts,
 } from "./data"
-import { Fs } from "./types"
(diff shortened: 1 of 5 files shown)


src/core/foundations/src/typography/types.ts
```diff
@@ -67,6 +69,6 @@ export type FontScaleFunctionStr = (options?: FontScaleArgs) => string
 export interface FontScaleArgs {
 	lineHeight?: LineHeight
 	fontWeight?: FontWeight
-	italic?: boolean
```
SiAdcock, line 70: Can we keep the old `italic` argument alive for now, to avoid introducing a breaking change?

src/core/foundations/src/typography/fs.ts
```diff
@@ -25,8 +40,7 @@ export const fs: Fs = category => (
 	// font is unavailable
 	const requestedFont = availableFonts[category][fontWeight]
 	const fontWeightValue = requestedFont ? fontWeightMapping[fontWeight] : ""
-	const fontStyleValue =
-		italic && requestedFont && requestedFont.hasItalic ? "italic" : ""
+	const fontStyleValue = getFontStyle(requestedFont, fontStyle)
```
SiAdcock, line 43: I like this. Perhaps all of this `xxxValue` logic should live in separate functions.

src/core/foundations/src/typography/types.ts
JamieB-gu: If both are present does the new argument take priority?

SiAdcock: Yeah, I guess that makes sense! Hopefully nobody will do that if we delete the documentation for the old API 😂

JamieB-gu: Ok, should be updated now

src/core/foundations/src/typography/fs.ts
JamieB-gu: Functions 💪

src/core/foundations/src/typography/types.ts
```diff
@@ -49,12 +51,13 @@ export type Fs = (
 	{
 		lineHeight,
 		fontWeight,
-		italic,
+		fontStyle,
```
SiAdcock, line 54: Do we still need to explicitly include the legacy `italic` in here? Presumably not, since it's compiling okay! It looks weird that it's missing though 🤔

src/core/foundations/src/typography/fs.ts
```diff
@@ -5,11 +5,34 @@ import {
 	fontWeightMapping,
 	availableFonts,
 } from "./data"
-import { Fs } from "./types"
+import { Fs, FontWeightDefinition, FontStyle, Option } from "./types"
+
+// Exists to support deprecated API, should be removed at some point
+const legacyItalic = (
```
SiAdcock, line 11: Great naming, great comment!

src/core/foundations/src/typography/types.ts
JamieB-gu: Oh interesting. I didn't spot this because it's not throwing a type error. I don't think it matters what it's called because this is a type definition... you can call it `arg0` or `arg1` for all TS cares. I think it's just syntactic constraints that mean you have to give it a name 🤷‍♂️

I've added it back in anyway.

PR #312, opened 2020-04-07
CI Excludes master And gh-pages

### LINKED ISSUE #304: Prevent CI action running on both master and gh-pages branches
**Describe the bug**
Currently the CI action runs on both `master` and `gh-pages` branches. This is unnecessary

**Expected behavior**
Only the Deploy to GitHub Pages action should run on `master`. We don't need any actions to run explicitly on `gh-pages`.

JamieB-gu:

Added `master` and `gh-pages` to ignored CI branches. CI will run on all other branches. From the docs [here](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-ignoring-branches-and-tags). Should fix #304.

## What does this change?

- Removes the `branches` positive match
- Adds the `branches-ignore` negative match for `master` and `gh-pages`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,9 @@
 name: CI
 on:
     push:
-        branches:
-            - "*"
+        branches-ignore:
+            - 'master'
+            - 'gh-pages'
 jobs:
     build:
         runs-on: ubuntu-latest


JamieB-gu:
Well I can see CD running against `master`, and CI ran against this branch, but not against `master` or the freshly updated `gh-pages` branch. Looks like it's working?

SiAdcock:
> Looks like it's working?

I would say 😁 👍 

Thanks so much!!

PR #313, opened 2020-04-07
Transaparent backgrounds for tertiary buttons

### LINKED ISSUE #311: Make tertiary buttons transparent

Currently tertiary buttons take an explicitly defined background that happens to be identical to the background colour of the theme in which they ought to be used. For instance, on default theme, the tertiary button has a white background.

Screenshot 2020-04-07 at 13 36 39

Screenshot 2020-04-07 at 13 37 40

The background colour of tertiary buttons ought to be transparent, so it can be used on different background colours.

Ideally, the border radius would be 0 to make the button feel more like a link when tabbing

oliverlloyd:
Fixes #311 

Changes the background colour for tertiary buttons from the theme background colour to `transparent`

## Why
Sometimes we want to land a button on a backgrond that isn't controlled by or even part of a theme. In this scenario we want the text on a `tertiary` button to display gracefully

### Screenshots

**Before**
Screenshot 2020-04-07 at 15 10 42

**After**
Screenshot 2020-04-07 at 15 11 18

## Focus Halo
It's more natural to have square borders when showing focus halos for buttons that only show text

### DIFF
--- a/src/core/components/button/stories.tsx
+++ b/src/core/components/button/stories.tsx
@@ -108,6 +108,22 @@ priorityBlue.story = {
 	},
 }
 
+export const priorityGrey = () => (
+	<ThemeProvider theme={buttonBrand}>
+		<div css={flexStart}>
+			{priorityButtons.map((button, index) => (
+				<div key={index}>{button}</div>
+			))}
+		</div>
+	</ThemeProvider>
+)
+priorityGrey.story = {
+	name: "priority grey",
+	parameters: {
+		backgrounds: [{ name: "grey", value: "lightgrey", default: true }],
+	},
+}
+
 export const priorityYellow = () => (
 	<ThemeProvider theme={buttonBrandAlt}>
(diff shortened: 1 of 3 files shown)


SiAdcock:
Nice one, thanks so much @oliverlloyd! And thanks for adding a story to help demonstrate it 🙏 

Could I request one more change<sup>1</sup> and make another suggestion<sup>2</sup> that we can discuss further if you like.

1. In a new commit, please could you delete [`button.tertiaryBackground`](https://github.com/guardian/source/blob/master/src/core/foundations/src/themes/button.ts) for each theme in `src-foundations`
2. Do you think we should override the [`border-radius`](https://github.com/guardian/source/blob/master/src/core/components/button/styles.ts#L65) to be 0 on tertiary buttons? As you mentioned in chat, it looks a bit odd when you're tabbing through

Screenshot 2020-04-07 at 16 01 13

PR #321, opened 2020-04-09
Develop accordion for Subscriptions Print landing page

lucymonie:
This is a pretty basic roll-out of the accordion to meet the needs of the new Subscriptions print landing page with the iMovo card tab (and accordions 🎉 ). 

## What does this change?
It's a start towards making accordions available in the design system.

### Screenshots
N/A - see Storybook!

### Accessibility
// I don't think these are currently being met 😢 - TODO

### DIFF
--- a/src/core/components/accordion/index.tsx
+++ b/src/core/components/accordion/index.tsx
@@ -7,20 +7,32 @@ import {
 	accordionToggle,
 	showAccordionElement,
 	hideAccordionElement,
+	grey,
+	white,
 } from "./styles"
 import { Props } from "@guardian/src-helpers"
 import { SvgChevronDownSingle, SvgChevronUpSingle } from "@guardian/src-svgs"
+
+type Colours = "grey" | "white"
+
+const backgroundColours = {
+	grey: grey,
+	white: white,
+}
 
 interface AccordionProps extends Props {
 	label: string
 	children: ReactNode
 }
 
-const Accordion = ({ children, label }: AccordionProps) => {
+const Accordion = ({ children, label, backgroundColour }: AccordionProps) => {
 	const [accordionOpen, toggleAccordion] = useState(false)
 
 	return (
-		<div css={accordion} aria-label="tablist">
+		<div
+			css={[accordion, backgroundColours[backgroundColour]]}
+			aria-label="tablist"
+		>
 			<div css={titleRow}>
 				<strong css={labelText}>{label}</strong>
 				<button
--- a/src/core/components/accordion/index.tsx
+++ b/src/core/components/accordion/index.tsx
@@ -1,7 +1,7 @@
 import React, { useState, ReactNode } from "react"
 import {
 	accordion,
-	normalise,
+	button,
 	labelText,
 	titleRow,
 	accordionRow,
@@ -26,14 +26,14 @@ const Accordion = ({ children, label, id }: AccordionProps) => {
 	const toggleToOpen = () => toggleAccordion(true)
 
 	return (
-		<div css={[accordion, normalise]}>
+		<div css={accordion}>
 			<button
 				aria-expanded={accordionOpen}
 				onClick={accordionOpen ? toggleToClosed : toggleToOpen}
-				aria-controls="accordionBody"
-				css={titleRow}
+				aria-controls={id}
+				css={[titleRow, button]}
 			>
-				<h3 css={labelText}>{label}</h3>
+				<strong css={labelText}>{label}</strong>
 				<div css={showHide}>
 					<div
 						css={
@@ -63,8 +63,7 @@ const Accordion = ({ children, label, id }: AccordionProps) => {
 			</button>
 			<div
 				id={id}
-				aria-expanded={accordionOpen}
-				aria-hidden={!accordionOpen}
+				hidden={!accordionOpen}
 				css={
 					accordionOpen ? showAccordionElement : hideAccordionElement
 				}
--- a/src/core/components/accordion/index.tsx
+++ b/src/core/components/accordion/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState, ReactNode } from "react"
+import React, { useState, ReactElement, ReactNode } from "react"
 import {
 	accordion,
 	button,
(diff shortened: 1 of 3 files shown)


src/core/components/accordion/index.tsx
```diff
@@ -1,27 +1,105 @@
...
-			<div>{children}</div>
+		<div
+			css={[accordion, backgroundColours[backgroundColour]]}
+			aria-label="tablist"
+		>
+			<div css={titleRow}>
+				<strong css={labelText}>{label}</strong>
+				<button
+					aria-controls="accordionBody"
+					aria-expanded={accordionOpen.toString()}
```
SiAdcock: I don't think we need to convert this to a string, `aria-expanded` should happily accept a boolean 🤞

lucymonie: All the fooffing for nothing 😆

src/core/components/accordion/index.tsx
```diff
@@ -1,27 +1,105 @@
...
+
 	return (
-		<div>
-			<strong css={label}>{labelText}</strong>
-			<div>{children}</div>
+		<div
+			css={[accordion, backgroundColours[backgroundColour]]}
+			aria-label="tablist"
+		>
+			<div css={titleRow}>
```
SiAdcock, line 37: Should we (can we?) make the entire row a button, so the user doesn't have to click the hide/show button... they can just click anywhere to expand?

src/core/components/accordion/index.tsx
```diff
@@ -1,27 +1,105 @@
...
+const Accordion = ({ children, label, backgroundColour }: AccordionProps) => {
+	const [accordionOpen, toggleAccordion] = useState(false)
+
 	return (
-		<div>
-			<strong css={label}>{labelText}</strong>
-			<div>{children}</div>
+		<div
+			css={[accordion, backgroundColours[backgroundColour]]}
+			aria-label="tablist"
```
SiAdcock, line 35: I don't think this adds much. Can we delete it?

src/core/components/accordion/index.tsx
```diff
@@ -1,27 +1,105 @@
...
+								? hideAccordionElement
+								: showAccordionElement
+						}
+					>
+						<SvgChevronDownSingle />
+					</div>
+				</button>
+			</div>
+			<div
+				id="accordionBody"
```
SiAdcock, line 85: We'll probably have to accept `id` as a prop in case we want to support more than one accordion on one page

src/core/components/accordion/index.tsx
```diff
@@ -1,27 +1,105 @@
...
+					aria-expanded={accordionOpen}
+					css={accordionToggle}
+				>
+					<div
+						css={
+							accordionOpen
+								? showAccordionElement
+								: hideAccordionElement
+						}
+						onClick={() => toggleAccordion(false)}
```
SiAdcock, line 50: We should put the `onClick` on the button element rather than the div

src/core/components/accordion/index.tsx
```diff
@@ -1,27 +1,105 @@
...
-interface AccordionProps extends Props {
-	children: ReactNode
-}
+type Colours = "grey" | "white"
 
-const Accordion = ({ children }: AccordionProps) => {
-	return <div css={accordion}>{children}</div>
+const backgroundColours = {
+	grey: grey,
+	white: white,
```
SiAdcock, line 20: I wonder if we should just support one of these for now to reduce complexity and API surface area? Maybe just the grey background if it serves your purpose?

src/core/components/accordion/index.tsx
```diff
@@ -1,27 +1,105 @@
...
+								: showAccordionElement
+						}
+					>
+						<SvgChevronDownSingle />
+					</div>
+				</button>
+			</div>
+			<div
+				id="accordionBody"
+				aria-expanded={accordionOpen}
```
SiAdcock, line 86: We only need `aria-expanded` [on the button element](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#Associated_ARIA_roles_states_and_properties)

SiAdcock: We probably want to add [the `hidden` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) when the accordion is closed

src/core/components/accordion/styles.ts
```diff
@@ -1,11 +1,83 @@
...
+
+	button {
+		background: none;
+		color: inherit;
+		border: none;
+		padding: 0;
+		font: inherit;
+		cursor: pointer;
+		outline: inherit;
+	}
```
SiAdcock, line 20: Shall we extract this into a separate `const`?

src/core/components/accordion/index.tsx
```diff
@@ -1,27 +1,105 @@
...
+					</div>
+					<div
+						css={
+							accordionOpen
+								? showAccordionElement
+								: hideAccordionElement
+						}
+					>
+						<SvgChevronUpSingle />
+					</div>
```
SiAdcock, line 62: Maybe we can combine the label and SVG into a single div?

```jsx
<div css={...}>
  Hide
  <SvgChevronUpSingle />
</div>

```

src/core/components/accordion/index.tsx
lucymonie: Yes, definitely! That is so much nicer

src/core/components/accordion/index.tsx
lucymonie: Yep 👍

src/core/components/accordion/index.tsx
lucymonie: Good point

src/core/components/accordion/index.tsx
lucymonie: 😳 Oops - have fixed!

src/core/components/accordion/index.tsx
lucymonie: Do you mean that I should just make the background grey and remove this from the API?

src/core/components/accordion/index.tsx
lucymonie: 👍

src/core/components/accordion/styles.ts
lucymonie: Sure thing - I have create a 'normaliser' style for this.

src/core/components/accordion/index.tsx
lucymonie: Yeees... this is undoubtedly my poor css chops, but I am struggling to get these to behave the way I want without quite a few divs to manage size and vertical alignment.

src/core/components/accordion/index.tsx
lucymonie: (I have gone ahead and done that, since keeping the API seems YAGNI with only one colour)

lucymonie: Yes, a transparent background seems like a good solution. I have done that.

PR #322, opened 2020-04-12
Icons Stories

oliverlloyd:
Adds stories for the icons

## Why?
Because I wanted to propose a new icon but fitst needed a story to be able to visualise the svg I want to add

Screenshot 2020-04-12 at 15 56 15

If I've put things in the wrong place, named things wrongly, etc. please don't hold back from telling me. Or just fix it to be the right way, I don't mind

### DIFF
--- a/src/core/svgs/stories.tsx
+++ b/src/core/svgs/stories.tsx
@@ -1,10 +1,8 @@
 import React from "react"
 import { css } from "@emotion/core"
-import { ThemeProvider } from "emotion-theming"
 
 import { size } from "@guardian/src-foundations"
 import { storybookBackgrounds } from "@guardian/src-helpers"
-import { brand } from "@guardian/src-foundations/themes"
 
 import {
 	SvgAlert,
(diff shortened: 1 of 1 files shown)


src/core/svgs/stories.tsx
```diff
@@ -0,0 +1,114 @@
...
+const Default = ({ children }: { children: JSX.Element | JSX.Element[] }) => (
+	<div css={[iconDefault]}>{children}</div>
+)
+
+export default {
+	title: "Icons",
+}
+
+export const SmallIcons = () => (
+	<ThemeProvider theme={brand}>
```
SiAdcock, line 51: SVGs don't use any theme information, so this is not really doing anything

src/core/svgs/stories.tsx
```diff
@@ -0,0 +1,114 @@
+import React from "react"
+import { css } from "@emotion/core"
+import { ThemeProvider } from "emotion-theming"
+
+import { storybookBackgrounds } from "@guardian/src-helpers"
+import { brand } from "@guardian/src-foundations/themes"
+
+import { iconDefault, iconSmall } from "../components/button/styles"
```
SiAdcock, line 8: Could we inline the relevant styles here? It's a pain having cross-package references

oliverlloyd:
> Please could you squash your changes before you merge?

I know this is some git wizardry where I run some commands and then force push to end up with one commit message but I use Github Desktop and try to limit the number of git commands I run at the command line, esp. ones with `--force` in them and esp. when it's not my repo!

Github has a 'Merge and Squash' option you could enable?

SiAdcock:
> Github has a 'Merge and Squash' option you could enable?

Indeed, I think it's already enabled! Can you confirm you see the fancy little dropdown menu next to the Merge button?

Screenshot 2020-04-14 at 15 40 36

oliverlloyd:
> Indeed, I think it's already enabled!

Oof! That's a UX fail. I read the greyed out text to indicate that it was disabled

Squashing now...

PR #323, opened 2020-04-12
GuardianRoundel

oliverlloyd:
To add the `GuardianRoundel` svg

Screenshot 2020-04-12 at 16 14 30

## Why?
Because we need this in DCR to place in the nav bar for [immersive articles](https://www.theguardian.com/film/2019/oct/17/ken-loach-all-his-films-ranked) as a link home but also because it seems like a useful svg to have as a pretty generic representation of the brand

## Problems with this PR that I need help with
There are 2.

1. I've not got the sizing or colours quite right, I'm using `56 56` but it should be`30 30` but if I make it smaller it cuts off the svg
2. The background and foreground colours are fixed but need to be dynamic

Screenshot 2020-04-12 at 16 15 19
Screenshot 2020-04-12 at 16 15 34

Sadly, I have no idea how to fix these problems. Me & Svgs? 🤷‍♂️ 

## Another possible problem
When we use this icon in DCR we want it to be bigger
```
    height: 2.625rem;
    width: 2.625rem;
```
Can we just override it the same as frontend does on [this page](https://www.theguardian.com/film/2019/oct/17/ken-loach-all-his-films-ranked) (at mobile view)?

### Depends on
https://github.com/guardian/source/pull/322


SiAdcock:
This is very cool, thanks very much @oliverlloyd. I shall find out whether we have a 30x30 version, but this icon would certainly a valueable addition.

sndrs:
since it's an SVG, size shouldn't matter too much should it? you get path-snapping problems on non-hidpi screens, but applying resize with CSS should be fine?

SiAdcock:
@sndrs yes, that's a good point. It's important that the ratio of width:height is consistent with other SVGs, which it will be (1:1). 

I'll chat with Akemi tomorrow to ensure this icon is added to Figma, and ask what the default fill colours should be.

oliverlloyd:
I added a component in DCR that uses the same svg as here but with size specifed and wrapped in an `a` tag linking to the home page. Maybe a better approach for Source is not to add the svg but to create the whole component in the same way?

SiAdcock:
Hi @oliverlloyd. I've added the Roundel to the new package introduced in #372 (`@guardian/src-brand`). Please let me know if this looks like it could work for you 🙇

oliverlloyd:
Closing as this svg is now available in the `brand` package

Issue #293, opened 2020-03-25
Add ChromaticQA


SiAdcock:
**Describe the solution you'd like**

DCR has been using [ChromaticQA](https://www.chromaticqa.com/) to excellent effect for cross-browser, multi-breakpoint visual regression testing. Since we now have a standard licence, we can start using it in other codebases too. Source will find a lot of benefit from it, especially cross-browser testing.

oliverlloyd:
I've created a [new project](https://www.chromaticqa.com/builds?appId=5e933197d94d9c002275a15e) and [added the lib](https://github.com/guardian/source/pull/324) to run Chromatic snapshots.

### Next steps
Either configure your CI server to run the snapshot on each push or add a [github action](https://github.com/chromaui/action) to do it. We've been using Github actions over in [DCR](https://github.com/guardian/dotcom-rendering/blob/master/.github/workflows/chromatic.yml) and [Discussion](https://github.com/guardian/discussion-rendering/blob/master/.github/workflows/chromatic.yml) with success. You just need to add the workflow and create github secret for `CHROMATIC_APP_CODE`. If you can't see the app code in Chromatic ask me and I can ping it to you.

SiAdcock:
Fixed in #324, thanks @oliverlloyd 🎉

PR #324, opened 2020-04-12
Chromatic

oliverlloyd:
Adds Chromatic lib to run visual snapshots 

Relates to #293

Chromatic project: https://www.chromaticqa.com/builds?appId=5e933197d94d9c002275a15e

Pending integration with CI: https://docs.chromaticqa.com/setup_ci


package.json
```diff
@@ -1,77 +1,78 @@
...
+    "eslint-config-prettier": "^6.1.0",
+    "eslint-plugin-prettier": "^3.1.0",
+    "eslint-plugin-react": "^7.14.3",
+    "execa": "^4.0.0",
+    "jest": "^25.2.7",
+    "prettier": "^1.18.2",
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6",
+    "regenerator-runtime": "^0.13.3",
+    "storybook-chromatic": "^3.5.2",
```
SiAdcock, line 64: It looks like this line is the substantive change. Do you know why Yarn has replaced all those lovely tabs with spaces?

oliverlloyd: Because I ran an `npx` command which changed some stuff and the `.editorconfig` file doesn't specify tabs for the `package.json` file so it defaulted to spaces

I fixed it 😉

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,18 @@
...
+            - uses: actions/setup-node@v1
+              with:
+                  node-version: 12
+            - run: |
+                  yarn
+                  yarn ci:build
+            - uses: chromaui/action@v1
+              with:
+                  appCode: ${{ secrets.CHROMATIC_APP_CODE }}
+                  token: ${{ secrets.GITHUB_TOKEN }}
```
oliverlloyd, line 18: 👍

package.json
SiAdcock: Right. I figured it would inherit from [`[*]`](https://github.com/guardian/source/blob/master/.editorconfig#L8) but maybe I need to be more explicit?

PR #333, opened 2020-04-20
Include dist dir in svgs package

### LINKED ISSUE #328: @guardian/src-svgs is missing the esm build
**Describe the bug**
`@guardian/src-svgs` `dist` dir only includes a cjs `index.js`, but the package declares:

```json
"main": "dist/index.js",
"module": "dist/index.esm.js",
``` 

Running `yarn build` in `src/core/svgs` _does_ generate the file though.

**To Reproduce**
1. `yarn add @guardian/src-svgs`
2. look in `node_modules/@guardian/src-svgs/dist`

**Expected behavior**
Both files should exist

cc @ripecosta

sndrs:

fixes #328

## What does this change?

publishes all JS files in dist dir

### DIFF
--- a/src/core/svgs/package.json
+++ b/src/core/svgs/package.json
@@ -44,7 +44,8 @@
 		"minus.tsx",
 		"plus.tsx",
 		"paypal.tsx",
-		"*.d.ts"
+		"*.d.ts",
+		"dist/*.js"
 	],
 	"peerDependencies": {
 		"react": "^16.8.6"


sndrs:
does one need adding in here too?

https://github.com/guardian/source/blob/1c16ef7c2bffee2a038d1148913fdbc07c7199e2/src/core/foundations/package.json#L21-L42

SiAdcock:
@sndrs No, we don't need to do this in src-foundations as there is no dist directory. The root directory is essentially the dist directory, and we explicitly include the `foundations.esm.js` file in the files list. Thanks for checking though 👍

PR #388, opened 2020-06-11
Text Area Rows/Height support

liywjl:

Remove `height` css from TextArea and explicitly support `rows` props.

## What does this change?

Allow textArea to use rows to determine height.

## Screenshots

**Before**

Screenshot 2020-06-11 at 10 19 53

**After**
Screenshot 2020-06-11 at 10 18 39

### Accessibility

-   [ ] The component doesn't use only colour to convey meaning

### Responsiveness

### Documentation

-   [x] Examples in Storybook

### DIFF
--- a/src/core/components/text-area/index.tsx
+++ b/src/core/components/text-area/index.tsx
@@ -51,6 +51,7 @@ const TextArea = ({
 					cssOverrides,
 				]}
 				aria-required={!optional}
+				rows={rows}
 				{...props}
 			/>
 		</label>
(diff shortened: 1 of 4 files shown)


src/core/components/text-area/index.tsx
```diff
@@ -31,6 +32,7 @@ const TextArea = ({
 	supporting,
 	error,
 	cssOverrides,
+	rows = 3,
```
SiAdcock, line 35: I think we'll have to explicitly pass `rows` to the `<textarea>` element

PR #403, opened 2020-06-23
Added spacing based on global space tokens

mxdvl:

Uniform spacing in buttons, to match more closely Figma intent.

Co-author: Simon Adcock, @SiAdcock 

## What does this change?

-   Change 1
-   Change 2

## Screenshots

**Before**

image

**After**

image

### Accessibility

### Responsiveness

### Documentation

-   [X] Examples in Storybook

### DIFF
--- a/src/core/components/button/index.tsx
+++ b/src/core/components/button/index.tsx
@@ -101,9 +101,10 @@ const Button = ({
 	children,
 	...props
 }: ButtonProps) => {
-	const buttonContents = [children]
+	const buttonContents = [<div className="baseline-kick">{children}</div>]
 
 	if (iconSvg) {
+		buttonContents.push(<div className="spacer"/>)
 		buttonContents.push(React.cloneElement(iconSvg, { key: "svg" }))
 	}
 
--- a/src/core/components/button/index.tsx
+++ b/src/core/components/button/index.tsx
@@ -104,6 +104,8 @@ const Button = ({
 	const buttonContents = [<div className="baseline-kick">{children}</div>]
 
 	if (iconSvg) {
+        // adding an empty spacer div is simpler than adding a left margin
+        // to icons on the right and a right margin to icons on the left =)
 		buttonContents.push(<div className="spacer"/>)
 		buttonContents.push(React.cloneElement(iconSvg, { key: "svg" }))
 	}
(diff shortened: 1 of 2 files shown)


src/core/components/button/styles.ts
```diff
@@ -135,22 +145,19 @@ export const iconXsmall = css`
...
+	.spacer {
+		width: ${space[1]}px;
+	}
 `
-
+// TODO: fix viewBox in SVGs
 export const iconRight = css`
-	svg {
-		margin: 0 ${-size.medium / 8}px 0 ${size.medium / 4}px;
-	}
```
SiAdcock, line 146: We no longer need this export (or comment), so we can remove it. This will require some refactoring in [index.tsx](https://github.com/guardian/source/blob/master/src/core/components/button/index.tsx). For instance [this](https://github.com/guardian/source/blob/217fcd972e287d28381f4e588eea4d0545acad55/src/core/components/button/index.tsx#L210-L212):

```tsx
iconSvg && (!hideLabel && children)
	? iconSides[iconSide]
	: "",
```
can be simplified to

```tsx
iconSvg && (!hideLabel && children)
	? iconLeft
	: "",
```

and so on...

src/core/components/button/index.tsx
```diff
@@ -101,9 +101,10 @@ const Button = ({
 	children,
 	...props
 }: ButtonProps) => {
-	const buttonContents = [children]
+	const buttonContents = [<div className="baseline-kick">{children}</div>]
 
 	if (iconSvg) {
+		buttonContents.push(<div className="spacer"/>)
```
SiAdcock, line 107:
```suggestion
        // adding an empty spacer div is simpler than adding a left margin
        // to icons on the right and a right margin to icons on the left =)
		buttonContents.push(<div className="spacer"/>)
```

src/core/components/button/index.tsx
```diff
@@ -101,9 +101,10 @@ const Button = ({
 	children,
 	...props
 }: ButtonProps) => {
-	const buttonContents = [children]
+	const buttonContents = [<div className="baseline-kick">{children}</div>]
```
SiAdcock, line 104: We'll need to apply these changes to [LinkButton](https://github.com/guardian/source/blob/master/src/core/components/button/index.tsx#L159) too. I'll look into consolidating these components when the API is mostly aligned.

PR #416, opened 2020-06-26
Reinstate nudge animation in buttons with icons

SiAdcock:

The nudge animation will be unavilable when the `showIcon` API is removed. We should find a way to hold onto this.

## What does this change?

-   add a new `nudgeIcon` prop. When set, the icon moves to the right on hover


paperboyo:
Cool! Just throwing a thought: do we want our own `Allow flashing elements` and/or `prefers-reduced-motion` to be considered in Source for all animating elements? Maybe we can have no-motion-but-colour-change variants?

PR #505, opened 2020-08-24
Prevent SVG repetition

mxdvl:
## What is the purpose of this change?

Reduce the size of SVG elements and make them easier to read.

## What does this change?

- Repetition of the squiggly line as a `xmlns:xlink`.
- Only one dot is repeated.
- horizontal lines are untouched.

## Screenshots

**Before**
squiggle before

dotted before

**After**

squiggle after 16px _(16px)_
squiggle after 12px _(12px)_

dotted after

### Accessibility

-   [ ] [Tested with screen reader](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/78ac51)
-   [ ] [Navigable with keyboard](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/009027)
-   [ ] [Colour contrast passed](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/58dbf2)
-   [ ] [The component doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/61524d)

### Cross browser and device testing

-   [ ] Tested with touch screen device

### Responsiveness

-   [X] Tested at all breakpoints
-   [X] Tested with with long text
-   [X] Stretched to fill a wide container

### Documentation

-   [ ] Full API surface area is documented in the README
-   [X] Examples in Storybook

PR #523, opened 2020-09-04
Eight lines of anything

mxdvl:

Why stick to four when you can have eight? This is a proof of concept to make these SVG. Follow-up of #505 and #302.

## What does this change?

- Allow eight dotted lines.
- Allow eight squiggly lines.
- Removed some of the spaces in the SVG, making for an even smaller source.

## Screenshots

**Before**

image

image

**After**

image

image

### Accessibility

### Responsiveness

### Documentation

-   [X] Full API surface area is documented in the README
-   [X] Examples in Storybook

### Known issues

Sometimes it can appear like there is one line extra or missing. If it does, let your eyes rest for a while, then check again. (Thanks @JamieB-gu / 430ebc4)

### DIFF
--- a/src/editorial/web/components/lines/README.md
+++ b/src/editorial/web/components/lines/README.md
@@ -27,11 +27,10 @@ const Section = () => (
 
 **`"straight" | "dotted" | "squiggly"`** _= "straight"_
 
-The appearance of the lines
+The appearance of the lines.
 
 ### `count`
 
 **`4 | 8`** _= 4_
 
-How many lines appear. Note: this applies to the `straight` effect only.
-Other effects always have 4 lines.
+How many lines appear.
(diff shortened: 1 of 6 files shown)


SiAdcock:
Thanks for this @mxdvl 🙏 I think the consistency here is useful. No breakingchanges, which makes me 😄 

@gtrufitt @JamieB-gu do you have any thoughts on this? If not, I'm happy to merge.

mxdvl:
> Thanks for this @mxdvl 🙏 I think the consistency here is useful. No breaking changes, which makes me 😄

I would not have dared 😉. I’ll let you accept the changes as baseline. I did not manage to get Chromatic to understand that the four lines were an updated component, rather than a new story item…

JamieB-gu:
> Allow eight dotted lines

The screenshot shows seven doesn't it 🤔

> do you have any thoughts on this

Yes: it looks great!

mxdvl:
I can really see eight… worth getting off screens, @JamieB-gu?

gtrufitt:
Haha, came here to check 8 dots, apparently it's Max that needs to step away from the screen 😆 😆 😆

Issue #526, opened 2020-09-09
RFC: Layout components


SiAdcock:
From #488, there was broad agreement in favour of a JSX-based API. I've started spiking a set of layout components.

# Abandoning the Grid components

I have moved away from the existing [Grid component API](https://theguardian.design/2a1e5182b/p/978527--grid-component/b/304251) for a few reasons:

1. The columns / spans abstraction is confusing and hard to remember. It expects the developer to know how many columns are available at each breakpoint. 
2. It provides an unnecessarily high granularity of control, making the API weighty for no good reason.
3. The name "Grid" is confusing and means too many things.
4. It tries (and fails) to be a one-stop solution for all layouts.

# Layout spiking

A number of developers and designers have kindly shared layouts that they are currently working on. Over the last few days I've built a number of them, noting layout similarities and patterns.

- Checkout flow
- Consents flow
- Thank you flow
- Article on web
- Showcase in Editions

As a result of this spike, I've identified 5 new layout components that, when combined, can be used to build (nearly) all of these layouts.

- Container
- Stack
- Columns / Column
- Tiles
- Hide

This work has been greatly inspired by awesome open source component libraries such as [Braid](https://seek-oss.github.io/braid-design-system/foundations/layout) and [Chakra-UI](https://chakra-ui.com/getting-started).

## Container

```tsx
<Container>
  <article></article>
</Container>
```

The Container centres the page content and applies a width that corresponds to the grid at the current breakpoint.

## Stack

```tsx
<Stack>
  <div></div>
  <div></div>
  <div></div>
</Stack>
```

Children will be stacked one on top of the other.

## Columns / Column

```tsx
<Columns>
  <Column width={1/4}></Column>
  <Column width={1/4}></Column>
  <Column></Column>
</Columns>
```

Columns will be arranged side by side on a single row, with the specified width.

`width`: If no value is provided, the column width will be fluid (i.e. take up remaining space, divided between all fluid columns).

## Tiles

```tsx
<Tiles columns={3}>
  <div></div>
  <div></div>
  <div></div>
  <div></div>
</Tiles>
```

Items will be arranged with equal width and spacing with the specified number of columns, wrapping onto a new row as necessary.

## Hide

```tsx
<Hide above="tablet">
  <MobileNavigation />
</Hide>
<Hide below="desktop">
  <DesktopNavigation />
</Hide>
```

Hide a component above or below a certain breakpoint

# Request for comments

[...]

PR #547, opened 2020-09-29
Add an Inline component

amyhughes:
#67  What is the purpose of this change?

This is for elements to sit next to each other with even spacing, for example:

image

This is part of addressing Issue #526

### Responsiveness

### Documentation

-   [x] Examples in Storybook

### DIFF
--- a/src/core/components/layout/components/inline/styles.ts
+++ b/src/core/components/layout/components/inline/styles.ts
@@ -1,0 +1,35 @@
+import { css, SerializedStyles } from "@emotion/core"
+import { space } from "@guardian/src-foundations"
+import { InlineSpace } from "./inline"
+
+export const inline = css`
+	display: flex;
+`
+
+const inlineSpaceStyle = (number: InlineSpace) => css`
+	& > * + * {
+		margin-left: ${space[number]}px;
+	}
+`
+
+export const inlineSpace: {
+	1: SerializedStyles
+	2: SerializedStyles
+	3: SerializedStyles
+	4: SerializedStyles
+	5: SerializedStyles
+	6: SerializedStyles
+	9: SerializedStyles
+	12: SerializedStyles
+	24: SerializedStyles
+} = {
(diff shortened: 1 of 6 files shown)


SiAdcock:
Thanks so much for this @amyhughes, it looks awesome!

I have identified one more decision that we need to make before we roll this out. When there are too many items to fit on a row, do we want to squeeze these items up (as currently happens) or do we want excess items to wrap onto a new row?

**Squeeze**

Screenshot 2020-10-01 at 10 10 22

**Wrap**

Screenshot 2020-10-01 at 10 10 08

I think wrapping is a better default. This is what the [Braid Design System](https://seek-oss.github.io/braid-design-system/components/Inline) does.

We could potentially add the "squeeze" behaviour as a variant: maybe it could accept as prop like `noWrap={true}`. We should consider implementing this only if someone has a use case for it.

I'd be happy for us to tackle this in our next pairing session, so there's no immediate rush to get this out 😊 

What do you think?

amyhughes:
Thanks for the suggestion @SiAdcock, and thanks for taking me through the change!

I've pushed up our changes, the story looks like this:

image

I've added a bottom margin since you last saw it to space out the rows, currently it's the same spacing as the left margin  - let me know what you think.

src/core/components/layout/components/inline/styles.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const inlineSpaceStyle = (number: InlineSpace) => css`
+	& > * {
+		margin-left: ${space[number]}px;
+		margin-bottom: ${space[number]}px;
+	}
+`
+
+export const removeLeftMargin = (number: InlineSpace) => css`
+	margin: 0 -${space[number]}px;
```
SiAdcock, line 18: This style gets applied to the same element as the styles produced by `inlineSpaceStyle`. It even accepts the same argument. Perhaps we can move the negative margin up to `inlineSpaceStyle`?

Issue #549, opened 2020-10-01
Fix lint warnings


SiAdcock:
**Describe the bug**

`typescript-eslint` is currently generating [151 warnings](https://github.com/guardian/source/runs/1182566249?check_suite_focus=true#step:4:293) on each build. The longer we fail to fix them, the more warnings we will accrus.

**To Reproduce**

```sh
$ git clone git@github.com:guardian/source.git
$ cd source
$ yarn
$ yarn validate
```

**Expected behavior**

There should be no warnings

**:warning: Important note**

It is not necessary to fix all these warnings in a single PR 😄

Kingarturs:
Can i work on this? ✌

SiAdcock:
@Kingarturs certainly, thanks so much! Let me know if you need any help 🎉

SiAdcock:
Fixed in #552

PR #552, opened 2020-10-01
Fixed linter warnings

Kingarturs:
I read the warnings and there was basically two types

1. The ones that i think it's just the linter being too strict
2. Some that were somewhat reasonable

## What i skipped
- Warnings that were just because some imported style attributes not being practically "used". (these are type 1)

## What i fixed
- Modified some imports from `./react` like ReactNode, ReactElement, ButtonHTMLAttributes and AnchorHTMLAttributes and replaced with it's `React.ReactNode` version (these are type 2).

**Note: I added a rule to the `.eslintrc.json` so the linter stops warning you for unused vars (Didn't found any var really unused so i think is better this way, but let me know if you prefer it the way it was).**

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -8,7 +8,8 @@
 		"prettier/react"
 	],
 	"rules": {
-		"@typescript-eslint/explicit-module-boundary-types": "off"
+		"@typescript-eslint/explicit-module-boundary-types": "off",
+		"@typescript-eslint/no-unused-vars" : "off"
 	},
 	"parserOptions": {
 		"ecmaFeatures": {
--- a/scripts/build-all.ts
+++ b/scripts/build-all.ts
@@ -1,4 +1,4 @@
-import execa from "execa"
+import execa, { ExecaReturnValue } from "execa"
 import { paths, getComponentPaths } from "./paths"
 
 const build = (dir: string) => {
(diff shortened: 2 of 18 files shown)


SiAdcock:
Hi @Kingarturs, thanks for looking into this!

I think you're right about `@typescript-eslint/no-unused-vars` being too strict. I might go so far as to say it's [a bit broken](https://github.com/typescript-eslint/typescript-eslint/issues/46) at the moment 😷. These types (type 1 and type 2 in your original description) are all being "used", albeit at compile time, so these warnings are all false positives. 

I think the best solution for now would be to disable the rule in the `.eslintrc.json` as you have done. 

There is a new rule called `@typescript-eslint/no-unused-vars-experimental` (undocumented, introduced in typescript-eslint/typescript-eslint#688) which seems to produce the desired result. However, I'm reluctant to use experimental plugins in this codebase for now.

Please could you revert all your _type 2_ fixes (I'm sorry!) but preserve the change to `.eslintrc.json` and I'll be happy to merge this. Thanks a lot for investigating 🙏

Kingarturs:
Thanks a lot for your comments! Reverted the changes, i was really afraid of just set the rule in the `.eslintrc.json` because it wasn't exactly what you asked for 😅, but i'm glad you agree.

Let me know if you need something else! 👍

SiAdcock (approved):
Thanks so much @Kingarturs, this looks great. In this codebase, there's no need to be afraid of raising a speculative PR, I'm always happy to discuss your reasoning ❤️  We really appreciate your contribution, as long as you don't spend too much of your own time 🙂

SiAdcock:
Fixes #549

Issue #551, opened 2020-10-01
Allow choice cards to wrap


SiAdcock:
**Describe the bug**

If choice cards don't fit into their container, they should wrap onto the next line

**To Reproduce**

Constrain the width of the container for a set of choice cards. Observe that the cards break out of their container

**Expected behavior**

The cards should wrap onto the next line. Currently, cards take up `1/n` of the available space, where `n` is the number of choice cards (`flex-grow: 1`). In order to preserve the widths of each choice card, we would need to switch from using flexbox to using [grid](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Mastering_Wrapping_of_Flex_Items#Single-dimensional_layout_explained) to layout the cards.

**Screenshots**

Currently:

![Screenshot 2020-10-01 at 14 48 03](https://user-images.githubusercontent.com/5931528/94818440-0a06fb80-03f6-11eb-8f20-ef26fdb9a402.png)

Ideally:

![Screenshot 2020-10-01 at 14 59 47](https://user-images.githubusercontent.com/5931528/94819091-c3fe6780-03f6-11eb-8386-4abd1d7fdd04.png)

rrrahal:
Hello,

I was looking at this problem and I have a question.
Is there a max size that a choice card should have?
Because when we set something like: `grid-template-columns` you have to set some sort of ceiling for the grid to know when a column ends, otherwise the button will occupy the whole space available (`flex-grow: 1`) 

I got it with `grid-template-columns: repeat(auto-fill, minmax(140px, 1fr));` but I don't think that's 100% correct:


<img width="516" alt="Screen Shot 2020-10-01 at 18 31 41" src="https://user-images.githubusercontent.com/10249593/94843394-8d374a00-0414-11eb-8cad-c676e9960cec.png">
this is without constraining the container size

<img width="499" alt="Screen Shot 2020-10-01 at 18 32 06" src="https://user-images.githubusercontent.com/10249593/94843402-8f99a400-0414-11eb-9ec0-d02129bc2918.png">
this is constraining the container size

SiAdcock:
Hi @rrrahal. Thanks so much for taking a look at this 🙏  

It's such a tricky problem! I suspect that you are correct and that there's no way of doing this without hardcoding either a ceiling or a minimum width for choice cards. However I don't think we should do this. It would be too constraining and would make these components less useful.

I have an alternative proposal that involves expanding the API.

**Default behaviour**

```tsx
[...]

PR #554, opened 2020-10-02
Add columns Prop to ChoiceCardGroup

rrrahal:
Adding a new prop to ChoiceCard Group for it to able to be displayed on a grid layout with a fixed amount of columns.
Related to https://github.com/guardian/source/issues/551

## What does this change?

-   Adds a new container for the component when the columns prop is set
-   Adds margin to the last child of the choice cards.

## Screenshots

Screen Shot 2020-10-02 at 12 34 16
Screen Shot 2020-10-02 at 12 34 31

### Accessibility

### Responsiveness

### Documentation

-   [x] Examples in Storybook

### Known issues

- Adding margin to the last choice card might be a bad choice but if I did not do it the last choice card would always be bigger than the other cards. I'm thinking about other ways to do it.

- The other thing is passing the prop to the css, I see you guys using theme and I don't know if it's ok to pass a prop from the component that is not related to the theme.

- If my changes are Ok, I'm willing to finish to do the screen reader and keyboard testing.


src/core/components/choice-card/index.tsx
```diff
@@ -37,6 +38,7 @@ interface ChoiceCardGroupProps extends Props {
 	supporting?: string
 	multi?: boolean
 	error?: string
+	columns?: number
```
SiAdcock, line 41: It's unlikely this component will ever need an unlimited number of columns. Being strict about the number of allowable columns will let us [avoid writing dynamic styles](https://www.theguardian.design/2a1e5182b/p/77c9d9-contributing/t/902350)

```tsx
type Columns = 2 | 3 | 4 | 5 | undefined
```

src/core/components/choice-card/styles.ts
```diff
@@ -29,6 +29,16 @@ export const flexContainer = css`
...
+export const gridContainer = ({columns} : {columns : number}) => css`
+	width: 100%;
+
+	${from.mobileLandscape} {
+		display: grid;
+		grid-template-columns: repeat(${columns}, 1fr);
+		row-gap: ${space[2]}px;
+	}
+`
+
```
SiAdcock, line 41: To avoid dynamic styles, we could extract the dynamic part into a separate function that populates an object:

```tsx
export const gridContainer = css`
	width: 100%;
	${from.mobileLandscape} {
		display: grid;
		row-gap: ${space[2]}px;
	}
`

const gridColumnsStyle = (columns: Columns) = css`
  	${from.mobileLandscape} {
		grid-template-columns: repeat(${columns}, 1fr);
	}
`

export const gridColumns: {[key in Columns]: SerializedStyles } = {
  2: gridColumnsStyle(2),
  3: gridColumnsStyle(3),
  4: gridColumnsStyle(4),
  5: gridColumnsStyle(5),
}

```

For reference, we've employed this pattern [elsewhere](https://github.com/guardian/source/blob/main/src/core/components/layout/components/stack/styles.ts)

src/core/components/choice-card/styles.ts
```diff
@@ -135,10 +145,6 @@ export const choiceCard = ({
 
 	${from.mobileLandscape} {
 		margin: 0 ${space[2]}px 0 0;
-
-		&:last-child {
-			margin: 0;
-		}
```
SiAdcock, line 141: We shouldn't need a margin if we use [`gap`](https://developer.mozilla.org/en-US/docs/Web/CSS/gap). Perhaps we can eliminate `margin` altogether when the gridContainer is applied

src/core/components/choice-card/stories/single-state-controlled-example.tsx
```diff
@@ -53,6 +53,60 @@ export const singleStateControlled = () => {
...
+						onChange={() => setSelected("yellow")}
+					/>
+				</ChoiceCardGroup>
+			</div>
+			<span css={message}>{selected} is selected</span>
+		</div>
+	)
+}
+
+
```
SiAdcock, line 105: Please could we move this into its own module, perhaps called `with-columns.tsx`?

SiAdcock (changes requested):
This is looking very promising 😍 Let me know if you have any questions regarding these requests. Thanks for all your work so far @rrrahal 🙏

src/core/components/choice-card/styles.ts
```diff
@@ -29,6 +29,16 @@ export const flexContainer = css`
 	}
 `
 
+export const gridContainer = ({columns} : {columns : number}) => css`
+	width: 100%;
+
+	${from.mobileLandscape} {
+		display: grid;
```
SiAdcock, line 36: Maybe we can put this block inside a [`@supports(display: grid) { ... }`](https://developer.mozilla.org/en-US/docs/Web/CSS/@supports) block. This way we can think about how it would look in browsers that don't fully support CSS grid (IE11).

src/core/components/choice-card/styles.ts
rrrahal: That's kind of tricky, I think.
Margin is associated with the ChoiceCard and that component does not know if it's container is flex or grid (and I don't think it should know). 
If I remove the margin then we can't use flex, unless we use something like `justify-content: space-between` (losing the correct spacing of 8px). 
Another solution is setting the default value of columns to 1 and always use gap.

What do you think?

src/core/components/choice-card/README.md
```diff
@@ -104,6 +104,13 @@ Whether choice card is checked. This is necessary when using the [controlled app
 
 **Note:** if you pass the `checked` prop, you **must** also pass an `onChange` handler, or the field will be rendered as read-only.
 
+### `columns`
+
+**`number`**
+
+The number of columns in which the ChoiceCardGroup should be displayed. If this is prop is not set it is going to be displayed without wrapping to the other line.
```
SiAdcock, line 111:
```suggestion
To render of grid of choice cards, specify the number of columns. If this prop is not set, cards will appear on a single line.
```

src/core/components/choice-card/styles.ts
SiAdcock: > Margin is associated with the ChoiceCard and that component does not know if it's container is flex or grid (and I don't think it should know).

Agreed. Can we preserve flex as the default and override the margin in `gridContainer`?

```tsx
export const gridContainer = css`
  // ...

  & > label {
    margin: 0;
  }
`
```

src/core/components/choice-card/styles.ts
```diff
@@ -135,10 +162,10 @@ export const choiceCard = ({
...
 	${from.mobileLandscape} {
 		margin: 0 ${space[2]}px 0 0;
+	}
 
-		&:last-child {
-			margin: 0;
-		}
+	&:last-child {
+		margin: 0;
 	}
```
SiAdcock, line 169: I think this `:last-child` block should be inside the `${from.mobileLandscape}` block. Do you agree?

SiAdcock (approved):
This looks awesome, thanks so much @rrrahal! 💎 👍

PR #557, opened 2020-10-02
Add hide label props

### LINKED ISSUE #555: Allow consumer to visually hide text labels

It is not always necessary to visually show a text label above a user input field such as text input field, text area or select box. An example would be when there is a single user input field on the page and there is enough additional context on the page to inform the user what kind of data is required.

Screenshot_2020-05-28_at_11 41 55

It should be possible for the consumer of these components to visually hide the text label of a user input field. We could achieve this by expanding the API of our user input field components to include a new prop:

```ts
{
  hideLabel: boolean
}
```

The naming here is consistent with the [same feature in the Button component](https://github.com/guardian/source/blob/4637c1142c1b2063e08fa1167354408c5b980173/src/core/components/button/index.tsx#L157).

By default, this prop is set to `false`. If set to true, the text label is visually hidden. The text label will still be read by a screen reader.

This solution should be applied to the following components:

- [ ] [TextInput](https://github.com/guardian/source/tree/main/src/core/components/text-input)
- [ ] [TextArea](https://github.com/guardian/source/tree/main/src/core/components/text-area)
- [ ] [Select](https://github.com/guardian/source/tree/main/src/core/components/select)
- [ ] [RadioGroup](https://github.com/guardian/source/tree/main/src/core/components/radio)
- [ ] [CheckboxGroup](https://github.com/guardian/source/tree/main/src/core/components/checkbox)

Agreon:
## What does this change?

This PR adds the `hideLabel`-prop to diffrent components. Labels hidden by `hideLabel` are not visible, but still readable by screen readers.

### Accessibility

### Responsiveness

### Documentation

-   [x] Examples in Storybook

* Closes #555

### DIFF
--- a/src/core/components/checkbox/README.md
+++ b/src/core/components/checkbox/README.md
@@ -60,6 +60,12 @@ Gets passed as the name attribute for each checkbox
 
 Appears as a legend at the top of the checkbox group
 
+### `hideLabel`
+
+**`boolean`** _= "false"_
+
+Visually hides the label.
+
 ### `supporting`
 
 **`string`**
--- a/src/core/components/checkbox/index.tsx
+++ b/src/core/components/checkbox/index.tsx
@@ -29,6 +29,7 @@ interface CheckboxGroupProps extends Props {
 	id?: string
 	name: string
 	label?: string
+	hideLabel: boolean
 	supporting?: string
 	error?: string
 	children: JSX.Element | JSX.Element[]
@@ -39,13 +40,14 @@ const CheckboxGroup = ({
 	id,
 	name,
 	label,
+	hideLabel,
 	supporting,
 	error,
 	cssOverrides,
 	children,
 	...props
 }: CheckboxGroupProps) => {
-	const legend = label ? <Legend text={label} supporting={supporting} /> : ""
+	const legend = label ? <Legend text={label} supporting={supporting} hideLabel={hideLabel} /> : ""
 
 	const message =
 		typeof error === "string" ? (
@@ -194,8 +196,8 @@ const Checkbox = ({
 					<SupportingText>{supporting}</SupportingText>
 				</div>
 			) : (
-				<LabelText>{labelContent}</LabelText>
-			)}
+					<LabelText>{labelContent}</LabelText>
+				)}
 		</label>
 	)
 
@@ -209,6 +211,11 @@ const checkboxDefaultProps = {
 	error: false,
 }
 
+const checkboxGroupDefaultProps = {
+	hideLabel: false
+}
+
 Checkbox.defaultProps = { ...checkboxDefaultProps }
+CheckboxGroup.defaultProps = { ...checkboxGroupDefaultProps }
 
 export { CheckboxGroup, Checkbox }
--- a/src/core/components/checkbox/stories/checkbox-group/legend.tsx
+++ b/src/core/components/checkbox/stories/checkbox-group/legend.tsx
@@ -49,4 +49,18 @@ legendBlue.story = {
 	},
 }
 
-export { legendLight, legendBlue }
+const legendHideLabelLight = () => (
+	<ThemeProvider theme={checkboxDefault}>
+		<CheckboxGroup name="emails" label="Email newsletters" id="emails" hideLabel={true}>
+			{checkboxes.map((checkbox, index) =>
+				React.cloneElement(checkbox, { key: index }),
+			)}
+		</CheckboxGroup>
+	</ThemeProvider>
+)
+
+legendHideLabelLight.story = {
+	name: "visually hide legend light",
+}
+
+export { legendLight, legendBlue, legendHideLabelLight }
--- a/src/core/components/label/index.tsx
+++ b/src/core/components/label/index.tsx
@@ -1,8 +1,13 @@
 import React, { ReactNode, LabelHTMLAttributes, HTMLAttributes } from "react"
-import { SerializedStyles } from "@emotion/core"
+import { SerializedStyles, css } from "@emotion/core"
 import { legend, labelText, optionalText, supportingText } from "./styles"
 import { Props } from "@guardian/src-helpers"
 export { labelDefault, labelBrand } from "@guardian/src-foundations/themes"
+import { visuallyHidden as _visuallyHidden } from "@guardian/src-foundations/accessibility";
+
+const visuallyHidden = css`
+	${_visuallyHidden}
+`
 
 interface LabelProps extends LabelHTMLAttributes<HTMLLabelElement>, Props {
 	text: string
@@ -15,6 +20,7 @@ interface LegendProps extends HTMLAttributes<HTMLLegendElement>, Props {
 	text: string
 	supporting?: string
 	optional: boolean
+	hideLabel?: boolean
 	cssOverrides?: SerializedStyles | SerializedStyles[]
 }
 
@@ -41,12 +47,13 @@ const Legend = ({
 	text,
 	supporting,
 	optional,
+	hideLabel,
 	cssOverrides,
 	...props
 }: LegendProps) => {
 	return (
 		<>
-			<legend css={[legend, cssOverrides]} {...props}>
+			<legend css={[legend, hideLabel ? visuallyHidden : undefined, cssOverrides]} {...props}>
 				<Text text={text} optional={optional} />
 			</legend>
 			{supporting ? <SupportingText>{supporting}</SupportingText> : ""}
--- a/src/core/components/radio/README.md
+++ b/src/core/components/radio/README.md
@@ -58,6 +58,12 @@ Gets passed as the name attribute for each radio button
 
 Appears as a legend at the top of the radio group
 
+### `hideLabel`
+
+**`boolean`** _= "false"_
+
+Visually hides the label.
+
 ### `supporting`
 
 **`string`**
--- a/src/core/components/radio/stories/radio-group/legend.tsx
+++ b/src/core/components/radio/stories/radio-group/legend.tsx
@@ -55,3 +55,14 @@ export const legendHorizontal = () => (
 legendHorizontal.story = {
 	name: `legend horizontal light`,
 }
+
+export const legendHideLabelLight = () => (
+	<RadioGroup name="colours" label="Select your preferred colour" hideLabel={true}>
+		{radios.map((radio, index) =>
+			React.cloneElement(radio, { key: index }),
+		)}
+	</RadioGroup>
+)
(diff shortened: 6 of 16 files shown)


src/core/components/checkbox/README.md
```diff
@@ -60,6 +60,12 @@ Gets passed as the name attribute for each checkbox
 
 Appears as a legend at the top of the checkbox group
 
+### `hideLabel`
+
+**`boolean`** _= "false_
```
SiAdcock, line 65:
```suggestion
**`boolean`** _= false_
```

src/core/components/checkbox/index.tsx
```diff
@@ -17,18 +17,24 @@ import {
...
+	${_visuallyHidden}
+`
+
 interface CheckboxGroupProps extends Props {
 	// TODO: this is required when an error is passed to correctly set aria-describedby
 	// We should introduce a way of generating element IDs when they are not explicitly passed
 	id?: string
 	name: string
 	label?: string
+	hideLabel?: boolean
```
SiAdcock, line 37: Can we make this mandatory, with a default value of `false`? Here and in other components

src/core/components/checkbox/index.tsx
```diff
@@ -39,13 +45,14 @@ const CheckboxGroup = ({
...
 	label,
+	hideLabel,
 	supporting,
 	error,
 	cssOverrides,
 	children,
 	...props
 }: CheckboxGroupProps) => {
-	const legend = label ? <Legend text={label} supporting={supporting} /> : ""
+	const legend = label ? <Legend text={label} supporting={supporting} cssOverrides={hideLabel ? visuallyHidden : undefined} /> : ""
```
SiAdcock, line 55: Nice work discovering the `cssOverrides` prop. However I think in this case, I'd rather there be a `visuallyHidden: boolean` prop that you pass to the Legend component. This would be tidier, I think, and less prone to error in the future (`cssOverrides` is intended as an escape hatch)

src/core/components/checkbox/stories/checkbox-group/legend.tsx
```diff
@@ -49,4 +49,18 @@ legendBlue.story = {
...
+		<CheckboxGroup name="emails" label="Email newsletters" id="emails" hideLabel={true}>
+			{checkboxes.map((checkbox, index) =>
+				React.cloneElement(checkbox, { key: index }),
+			)}
+		</CheckboxGroup>
+	</ThemeProvider>
+)
+
+legendHideLabelLight.story = {
+	name: "hideLabel light",
```
SiAdcock, line 63:
```suggestion
	name: "visually hide legend light",
```

src/core/components/radio/README.md
```diff
@@ -58,6 +58,12 @@ Gets passed as the name attribute for each radio button
 
 Appears as a legend at the top of the radio group
 
+### `hideLabel`
+
+**`boolean`** _= "false_
```
SiAdcock, line 63:
```suggestion
**`boolean`** _= false_
```

src/core/components/radio/stories/radio-group/legend.tsx
```diff
@@ -55,3 +55,14 @@ export const legendHorizontal = () => (
...
+
+export const legendHideLabelLight = () => (
+	<RadioGroup name="colours" label="Select your preferred colour" hideLabel={true}>
+		{radios.map((radio, index) =>
+			React.cloneElement(radio, { key: index }),
+		)}
+	</RadioGroup>
+)
+legendHideLabelLight.story = {
+	name: `hideLabel light`,
```
SiAdcock, line 67:
```suggestion
	name: `visually hide legend light`,
```

src/core/components/select/README.md
```diff
@@ -47,6 +47,12 @@ Additional text that appears below the label
 
 Adds the word "Optional" after the label
 
+### `hideLabel`
+
+**`boolean`** _= "false_
+
+Hides the label and the `required` text set by the `optional` flag visually.
```
SiAdcock, line 54:
```suggestion
Visually hides the label and the "Optional" text set by the `optional` flag.
```

src/core/components/select/stories/hideLabel.tsx
```diff
@@ -0,0 +1,25 @@
...
+		<Select label="State" hideLabel={true}>
+			<Option value="">Select a state</Option>
+			<Option value="al">Alabama</Option>
+			<Option value="ca">California</Option>
+		</Select>
+	</div>
+)
+
+hideLabelLight.story = {
+	name: "hideLabel light",
```
SiAdcock, line 24:
```suggestion
	name: "visually hide label light",
```

src/core/components/text-area/README.md
```diff
@@ -58,6 +58,12 @@ The contents of the text area. This is necessary when using the [controlled appr
 
 Adds the word "Optional" after the label. Non-optional fields are rendered with the `required` attribute.
 
+### `hideLabel`
+
+**`boolean`** _= "false_
```
SiAdcock, line 63:
```suggestion
**`boolean`** _= false_
```

src/core/components/text-area/README.md
```diff
@@ -58,6 +58,12 @@ The contents of the text area. This is necessary when using the [controlled appr
...
+
+Hides the label and the `required` text set by the `optional` flag visually.
```
SiAdcock, line 65:
```suggestion
Visually hides the label and the "Optional" text set by the `optional` flag.
```

src/core/components/text-area/stories.tsx
```diff
@@ -42,6 +42,16 @@ optionalLight.story = {
...
 }
 
+const hideLabelLight = () => (
+	<div css={wrapperStyles}>
+		<TextArea label="Comments" hideLabel={true} />
+	</div>
+)
+
+hideLabelLight.story = {
+	name: `hideLabel light`,
```
SiAdcock, line 52:
```suggestion
	name: `visually hide label light`,
```

src/core/components/text-input/README.md
```diff
@@ -59,6 +59,12 @@ The contents of the text input field. This is necessary when using the [controll
 
 Adds the word "Optional" after the label. Non-optional fields are rendered with the `required` attribute.
 
+### `hideLabel`
+
+**`boolean`** _= "false_
```
SiAdcock, line 64:
```suggestion
**`boolean`** _= false_
```

src/core/components/text-input/README.md
```diff
@@ -59,6 +59,12 @@ The contents of the text input field. This is necessary when using the [controll
...
+
+Hides the label and the `required` text set by the `optional` flag visually.
```
SiAdcock, line 66:
```suggestion
Visually hides the label and the "Optional" text set by the `optional` flag.
```

src/core/components/text-input/stories.tsx
```diff
@@ -48,6 +48,26 @@ optionalLight.story = {
...
+					value={state}
+					onChange={(event) => setState(event.target.value)}
+				/>
+			</div>
+		</ThemeProvider>
+	)
+}
+
+hideLabel.story = {
+	name: `hideLabel light`,
```
SiAdcock, line 68:
```suggestion
	name: "visually hide label light",
```

SiAdcock (changes requested):
This is a great start @Agreon, many thanks for all the work and research you have put in. I have made mostly superficial change requests, with one larger suggestion to expand the Legend API slightly. Please let me know if you have any questions or would like to discuss any of the requests further 🙂

src/core/components/checkbox/index.tsx
Agreon: Of course, one question just for my understanding: What is the difference/gain? `defaultChecked` from `Checkbox` has no default prop either..

src/core/components/checkbox/index.tsx
```diff
@@ -17,6 +17,7 @@ import {
 	errorCheckbox,
 } from "./styles"
 import { Props } from "@guardian/src-helpers"
+import { visuallyHidden as _visuallyHidden } from "@guardian/src-foundations/accessibility";
```
SiAdcock, line 20: Is this still being used?

src/core/components/label/index.tsx
```diff
@@ -41,12 +47,13 @@ const Legend = ({
...
 	supporting,
 	optional,
+	hideLabel,
 	cssOverrides,
 	...props
 }: LegendProps) => {
 	return (
 		<>
-			<legend css={[legend, cssOverrides]} {...props}>
+			<legend css={[legend, cssOverrides, hideLabel ? visuallyHidden : undefined]} {...props}>
```
SiAdcock, line 56: It's important that `cssOverrides` is the last item in this array. This ensures that the CSS properties in `cssOverrides` override all other CSS properties.

SiAdcock (changes requested):
This is super-close to done! I've added a couple of small requests. Finally, would it be possible to rebase your commits so that [each commit affects a single package](https://www.theguardian.design/2a1e5182b/p/77c9d9-contributing/t/43233e)? Let me know if you need any clarification on that. Thanks so much for your work so far!

src/core/components/checkbox/index.tsx
SiAdcock: The difference is really small, but it's a philosophical point that I try to adhere to: how can we minimise complexity?
If `hideLabel` is a optional, its type is effectively:

```ts
hideLabel: true | false | undefined
```

Whereas if we provide a default value, its type is:

```ts
hideLabel: true | false
```

By providing a stricter type, it reduces the number of possible states the component could be in. This makes our component more predictable. The difference is not very dramatic when you focus on a single prop, but every prop we create adds more state permutations.

If you think this is harmful, misguided or if there's something else I might be missing, please let me know. I'd love to hear additional opinions on this 🙂

SiAdcock: The difference with defaultChecked is that this is overriding a React built-in attribute. In React, [`defaultChecked` is optional](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/7d5565b57a32e6d69bd9b3daa94194effd33f6e0/types/react/index.d.ts#L1748), and I didn't want to mess with the expectation. I hope that makes sense, please let me know if you have any thoughts on this.

Agreon: Thanks for the explanation! Sounds legit :)

SiAdcock (approved):
This is totally awesome, thanks so much @Agreon! 💎  🎆 👍

Agreon:
I'm happy to help :)

PR #567, opened 2020-10-08
feat(button): remove unused props showIcon

### LINKED ISSUE #556: React warning on passing prop to DOM element
**Describe the bug**

In Storybook we're seeing the following error in dev on the [icon-only link button story](http://localhost:9009/?path=/story/button--icon-only-link-buttons) (<-- this link only works when running `yarn storybook` locally)

```
Warning: React does not recognize the `showIcon` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `showicon` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in a (created by Context.Consumer)
    in EmotionCssPropInternal (created by LinkButton)
    in LinkButton
    in div (created by storyFn)
    in div (created by Context.Consumer)
    in EmotionCssPropInternal (created by storyFn)
    in div (created by storyFn)
    in storyFn
    in ErrorBoundary
```

This is because we are passing the (now unused) [showIcon prop](https://github.com/guardian/source/blob/4637c1142c1b2063e08fa1167354408c5b980173/src/core/components/button/index.tsx#L250) to the DOM in the LinkButton component. We could delete this prop entirely.

**Expected behavior**

We should not pass this prop to the underlying DOM element, so this warning would not appear

mathieumure:

Fixes #556

## What does this change?

It seems that the prop `showIcon` is not used and was passed directly to a dom element.
As suggested, I remove this prop.

### Accessibility

### Responsiveness

### Documentation

-   [ ] Examples in Storybook

### DIFF
--- a/src/core/components/button/index.tsx
+++ b/src/core/components/button/index.tsx
@@ -247,7 +247,6 @@ const defaultLinkButtonProps = {
 	size: "default",
 	iconSide: "left",
 	hideLabel: false,
-	showIcon: false,
 }
 
 Button.defaultProps = { ...defaultButtonProps }


SiAdcock (approved):
Awesome, thanks @mathieumure! 👍

Issue #471, opened 2020-07-30
Link components with icon do not wrap correctly


SiAdcock:
**Describe the bug**

In Link and LinkButton, we use `display: inline-flex` to pull icons down to the baseline of the link text description. This means long text descriptions with icons don't wrap correctly and sit on a separate line to the surrounding text.

**Expected behavior**

In Link, at least, the text should wrap correctly.

**Screenshots**

Current behaviour:

![Screenshot 2020-07-30 at 10 44 23](https://user-images.githubusercontent.com/5931528/88908365-ab849c00-d251-11ea-8de1-b08234829d20.png)

Expected behaviour:
![Screenshot 2020-07-30 at 10 47 57](https://user-images.githubusercontent.com/5931528/88908785-3ebdd180-d252-11ea-8d5d-8d2876a90196.png)

(Note: in the above screenshot, the icon should also be underlined)

**Additional context**

The correct behaviour is (I believe) impossible to achieve with LinkButton, since button elements cannot be made to appear inline. The workaround is to pass an `onClick` handler and `role="button"` to a Link component.

PR #571, opened 2020-10-13
Inline link icon story

webb04:
Fixes https://github.com/guardian/source/issues/471 by removing `inline-flex` styles.

## What does this change?
-   Adds story for inline `Link` component with a left and right icon

## Screenshots
| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/core/components/link/styles.ts
+++ b/src/core/components/link/styles.ts
@@ -53,7 +53,12 @@ export const subdued = css`
 export const icon = css`
 	svg {
 		fill: currentColor;
-		vertical-align: text-top;
+		/*
+		TODO: hardcoded bottom margin to vertically align
+		icons with text. This needs to be revisited when
+		the rules of icon spacing have been formalised
+		 */
+		margin-bottom: -3px;
 		width: ${width.iconXsmall}px;
 		height: auto;
 	}
--- a/src/core/components/link/styles.ts
+++ b/src/core/components/link/styles.ts
@@ -53,7 +53,7 @@ export const subdued = css`
 export const icon = css`
 	svg {
 		fill: currentColor;
-		vertical-align: middle;
+		vertical-align: text-top;
 		width: ${width.iconXsmall}px;
 		height: auto;
 	}
--- a/src/core/components/link/styles.ts
+++ b/src/core/components/link/styles.ts
@@ -61,12 +61,13 @@ export const icon = css`
 
 export const iconRight = css`
 	svg {
-		margin-left: ${space[1]}px;
+		margin-left: -${space[5]}px;
 	}
 `
 
 export const iconLeft = css`
 	svg {
+		margin-left: -${space[6]}px;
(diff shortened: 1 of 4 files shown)


SiAdcock:
Thanks for this @webb04. It does work, but with one small change: on hover, contrary to [the design](https://www.figma.com/file/vTlZFrZswOYXNMqLKpvv7r/Links?node-id=6%3A7), the icon is no longer underlined. This is the reason for all the [weird spacing stuff](https://github.com/guardian/source/blob/8d52dfa477bf4da1f4872aff3a64940472e47e9b/src/core/components/link/index.tsx#L55). 

Can you think of a way we can preserve the underlined icon on hover?

src/core/components/link/styles.ts
```diff
@@ -50,27 +51,23 @@ export const subdued = css`
...
 export const icon = css`
-	/*
-	FIXME: This makes icons sit vertically centred, but breaks links with long
-	text descriptions onto a separate line to surrounding text
-	 */
-	display: inline-flex;
 	svg {
 		fill: currentColor;
-		position: absolute;
+		vertical-align: middle;
```
SiAdcock, line 56: With this change, the icons rest on the underline or sometimes slightly below it. I know there is some inconsistency in the spacing in the SVGs themeselves, but it would be reassuring if we could improve the way icons look in the example stories.

**Previously**

Screenshot 2020-10-16 at 15 12 33

**Now**

Screenshot 2020-10-16 at 15 14 33

I wonder if we could add a bit of artificial spacing using negative margings or something?

src/core/components/link/styles.ts
```diff
@@ -50,27 +51,23 @@ export const subdued = css`
...
 export const icon = css`
-	/*
-	FIXME: This makes icons sit vertically centred, but breaks links with long
-	text descriptions onto a separate line to surrounding text
-	 */
-	display: inline-flex;
 	svg {
 		fill: currentColor;
-		position: absolute;
+		vertical-align: text-top;
```
SiAdcock, line 56: It's admirable to try and avoid hardcoding here, but to my eye the icons are riding a little high. I think I'd prefer to fix the alignment in a hacky way in the short term with a view to fixing it properly when we've spent more time thinking about the rules of icon spacing.

```suggestion
		/*
		TODO: hardcoded bottom margin to vertically align
		icons with text. This needs to be revisited when
		the rules of icon spacing have been formalised
		 */
		margin-bottom: -3px;
```

For a visual comparison:

`vertical-align: text-top`

Screenshot 2020-10-19 at 12 03 08

`margin-bottom: -3px`

Screenshot 2020-10-19 at 12 02 14

webb04: That looks better, I agree that consistent icon spacing would make this easier!

SiAdcock (approved):
Thanks so much for this @webb04 😍  👍

PR #600, opened 2020-11-06
Tag prereleases as beta

SiAdcock:

Currently if a prerelease is the latest release and a project attempts to add a Source package as a dependency, by default the project will get the prerelease. This might not be what the project maintainer wants or expects.

We should implement a mechanism that ensures prereleases receive a beta tag so that they will not be installed automatically.

See [related Client Side Infra discussion](https://github.com/orgs/guardian/teams/client-side-infra/discussions/16)

## What does this change?

-   Add a beta tagging script

See [`yarn tag` docs](https://classic.yarnpkg.com/en/docs/cli/tag/)

### DIFF
--- a/scripts/tag-beta-all.ts
+++ b/scripts/tag-beta-all.ts
@@ -1,0 +1,25 @@
+import execa from "execa"
+import { version } from "../package.json"
+import { packageNames, getComponentPackageNames } from "./package-names"
+
+const tag = (packageName: string) => {
+	return execa("yarn", ["tag", "add", `${packageName}@${version}`, "beta"], {
+		stdio: "inherit",
+	})
+}
+
+const { foundations, svgs, icons, brand, helpers } = packageNames
+
+const packages = getComponentPackageNames().then((ps) =>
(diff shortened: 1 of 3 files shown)


scripts/tag-beta-all.ts
```diff
@@ -0,0 +1,28 @@
+// import execa from "execa"
+import { version } from "../package.json"
+import { packageNames, getComponentPackageNames } from "./package-names"
+
+const tag = (packageName: string) => {
+	// return execa("yarn", ["tag", "add", `${packageName}@${version}`, "beta"], {
+	// 	stdio: "inherit",
+	// })
+	console.log(`yarn tag add ${packageName}@${version} beta`)
```
SiAdcock, line 9: FIXME!

SiAdcock:
Let's schedule the 2.6.0 release for Monday and run this script as part of the pre-release to test it works.

SiAdcock:
I _think_ this works, but `yarn` has [some issues related to tagging](https://github.com/yarnpkg/yarn/issues/7823). i'll merge for now, and delete it if it becomes a problem

PR #579, opened 2020-10-20
Filled roundels

SiAdcock:
## What is the purpose of this change?

There is a requirement to ensure the roundel only ever appears in a limited number of colour combinations. They are:

- White G on black - `default` - for general use on light coloured backgrounds 
- White G on blue - `brand` - for general use on light coloured backgrounds
- Black G on white - `inverse` - for use on darker backgrounds (e.g. dark mode)
- Blue G on white - `brandInverse` - for use on brand blue backgrounds

## What does this change?

-   Add fillable Roundel component 
-   Add filled variants of Roundel
-   Add a brand package README
-   Add inverse background support to Storybook

### default

Screenshot 2020-10-20 at 14 33 10

### inverse

Screenshot 2020-10-20 at 14 33 26

### brand

Screenshot 2020-10-20 at 14 33 17

### brandInverse

Screenshot 2020-10-20 at 14 33 33

### Accessibility

-   [x] [Colour contrast passed](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/58dbf2)
-   [x] [The component doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/61524d)

### Documentation

-   [x] Examples in Storybook

PR #605, opened 2020-11-12
Roundels: optimise SVGs

mxdvl:

Optimise the SVG roundels’ code that was introduced in #579.

## What does this change?

- use brand colours
- reduce precision to 2 decimals
- use circle where relevant


src/core/brand/roundel-brand.tsx
```diff
@@ -1,6 +1,7 @@
 import React from "react"
 import { SvgRoundelFilled } from "./roundel-filled"
+import { brand } from "@guardian/src-foundations/palette"
```
SiAdcock, line 3: This is a great idea! Would it be possible for you to add `@guardian/src-foundations` as a `peerDependency` of this package?

mxdvl: Thanks, I’ll have a look at that!

src/core/brand/roundel-filled.tsx
```diff
@@ -9,16 +9,10 @@ export const SvgRoundelFilled = ({
...
+			<circle cx="21" cy="21" r="21" fill={background} />
 			<path
-				fillRule="evenodd"
-				clipRule="evenodd"
-				d="M21.0001 0C9.4022 0 0 9.40185 0 21C0 32.598 9.4022 42 21.0001 42C32.5981 42 42 32.598 42 21C42 9.40185 32.5981 0 21.0001 0Z"
-				fill={background}
-			/>
-			<path
-				fillRule="evenodd"
-				clipRule="evenodd"
```
SiAdcock, line 20: Lazy Q: fillRule and clipRule get included by Figma when exporting SVGs. Are they doing anything useful, or can we just delete them every time?

SiAdcock (approved):
Thanks so much @mxdvl this looks amazing! Please could you squash and merge?

src/core/brand/roundel-filled.tsx
mxdvl: They can be useful in complex cut-out shapes.

PR #634, opened 2020-12-03
Use src-label in text area

SiAdcock:

We now have the reusable `@guardian/src-label` component, meaning the label defined in `@guardian/src-textarea` is redundant.

We should swap out textarea's label for the reusable component

## What does this change?

-   swap out textarea's label for the reusable label component

There should be no visual or API differences


SiAdcock:
Messed up branches, replaced with #636

PR #666, opened 2020-12-23
Disallow pointer events on the checkbox tick to prevent Selenium test issues

### LINKED ISSUE #663: Animation overlay span in Checkbox intercepts Selenium click events aimed at the underlying input
**Describe the bug**
When running our Selenium tests against the Subscriptions checkouts, the tests are failing on Source checkboxes due to the span that provides the tick animation intercepting the click event. An example of a failed test run is here: https://travis-ci.org/github/guardian/support-frontend/builds/750209284

The same behavior appears whether using `clickOn` or the `checkbox(selector).select()` methods from ScalaTest. However the problem is fixed if the offending span has `pointer-events: none` applied, which does not seem to otherwise effect the component in any way.

**To Reproduce**
Clone [support-frontend](https://github.com/guardian/support-frontend/) and rewind to commit 187be8a, then run the Selenium tests with `sbt selenium:test`. The tests will fail when attempting to check out using direct debit, as the checkbox to confirm that a user is the account holder cannot be clicked.

**Expected behavior**
The checkbox should be clickable using Selenium in order to verify the expected behaviour of the checkout as a whole.

The fix for this is really tiny- it's just applying `pointer-events: none` to that span in the Checkbox component- so I'm happy to raise a PR to fix this myself if you'd like!

i-hardy:

This fixes #663. Our Selenium tests on support-frontend were failing on the checkbox component, as the simulated click event on the checkbox input was intercepted by the span element that provides the animated tick. Adding a `pointer-events: none` rule to this span fixes the issue without causing any visual regressions.

The issue doesn't affect the radio buttons, despite also having an animated overlay, as in that component it's provided by a pseudo-element. 

## What does this change?

-   Added a `pointer-events: none` rule to the `tick` styles, and an explanatory comment

### DIFF
--- a/src/core/components/checkbox/styles.ts
+++ b/src/core/components/checkbox/styles.ts
@@ -120,6 +120,11 @@ export const tick = ({
 		*/
 		top: 14px;
 		left: 9px;
+		/*
+			this prevents simulated click events to the checkbox, eg from Selenium tests
+			from being intercepted by the tick
+		*/
+		pointer-events: none;
 
 		/* the checkmark ✓ */
 		&:after,


SiAdcock (approved):
This looks great, thanks so much @i-hardy! 💎 👍 🙏

PR #678, opened 2021-01-04
Add support for optional prop in RadioGroup

liywjl:

Radio inputs in HTML5 support required attribute, meaning the browser can enforce selection instead of needing to use JS validators to check if an option has been selected.

## What does this change?

- Adds support for `optional` prop in `RadioGroup`
- Adds storybook example 

### Required
Screenshot 2021-01-04 at 17 15 33

### Optional
Screenshot 2021-01-04 at 18 08 36

### Accessibility

### Responsiveness

### Documentation

-   [x] Examples in Storybook


src/core/components/radio/index.tsx
```diff
@@ -129,6 +138,8 @@ interface RadioProps extends InputHTMLAttributes<HTMLInputElement>, Props {
 	defaultChecked?: boolean
 	label?: ReactNode
 	supporting?: ReactNode
+	error: boolean
```
SiAdcock, line 141: What does it mean to have an error at the radio button level? We're already capturing errors at the radio group level (for instance, to indicate that the user must select an option). Is it useful to indicate that an individual radio button has been selected erroneously?

src/core/components/radio/index.tsx
```diff
@@ -188,11 +203,16 @@ const Radio = ({
...
 	orientation: "vertical",
+	optional: false,
 	hideLabel: false,
 }
 
 const radioDefaultProps = {
 	disabled: false,
+	type: "radio",
+	optional: false,
+	defaultChecked: false,
```
SiAdcock, line 214: I think having a `defaultChecked` attribute passed by default is going to cause problems when a developer also passes a `checked` attribute. React doesn't allow both attributes to be passed at the same time.

src/core/components/radio/index.tsx
```diff
@@ -188,11 +203,16 @@ const Radio = ({
 
 const radioGroupDefaultProps = {
 	orientation: "vertical",
+	optional: false,
 	hideLabel: false,
 }
 
 const radioDefaultProps = {
 	disabled: false,
+	type: "radio",
```
SiAdcock, line 212: This is already hardcoded in the `Radio` component's `<input>` element. Where do you think is the best place to define this?

src/core/components/radio/index.tsx
```diff
@@ -129,6 +138,8 @@ interface RadioProps extends InputHTMLAttributes<HTMLInputElement>, Props {
 	defaultChecked?: boolean
 	label?: ReactNode
 	supporting?: ReactNode
+	error: boolean
+	optional?: boolean
```
SiAdcock, line 142: Optional is passed as a default, so doesn't need to be marked as optional... Urgh, confusing! To put it another way, I think the question mark is redundant here

src/core/components/radio/stories/radio-group/optional.tsx
```diff
@@ -0,0 +1,30 @@
+import React from "react"
+import { RadioGroup, Radio } from "../../index"
+
+const optional = () => (
+	<form onSubmit={e => e.preventDefault()}>
+		<RadioGroup name="yes-or-no" optional={true}>
```
SiAdcock, line 6: It would be useful if we passed a `label` prop here to make the story feel more realistic and to balance the "Optional" label. Something like: "Do you live in the United Kingdom?"

src/core/components/radio/stories/radio-group/optional.tsx
```diff
@@ -0,0 +1,30 @@
...
+		<button type="submit">Submit</button>
+	</form>
+)
+optional.story = {
+	name: "optional",
+}
+
+const required = () => (
+	<form onSubmit={e => e.preventDefault()}>
+		<RadioGroup name="yes-or-no" optional={false}>
```
SiAdcock, line 19: As above, a label would really help here

SiAdcock:
Thanks so much for this @liywjl. I have rebased your changes on top of the main branch so they are ready for more discussion! 🗣

Along with the comments above, I have two more recommendations that I think will help this along:

1. Give the `RadioGroup` component's `<fieldset>` element a `role` of `radiogroup`. This will announce the element as a "radio group" to screen readers... currently it's announced as a "group", which feels a bit generic
2. Give the `<fieldset>` an `aria-required`, which is set to a negation of the `optional` prop. This announces to screen readers that the fieldset is required. Currently screen reader users will only know that the field is required if they drill into an individual radio button

**Before**

Screenshot 2021-01-07 at 11 12 11

**After**

Screenshot 2021-01-07 at 11 13 39

src/core/components/radio/index.tsx
```diff
@@ -153,6 +166,8 @@ const Radio = ({
 			type="radio"
 			css={(theme) => [radio(theme.radio && theme), cssOverrides]}
 			value={value}
+			aria-invalid={error}
```
SiAdcock, line 169: Related to the [`error` prop comment](https://github.com/guardian/source/pull/678/files#r553256066), this line is giving us a linting error:

```
error  The attribute aria-invalid is not supported by the role radio. This role is implicit on the element input  jsx-a11y/role-supports-aria-props
```

src/core/components/radio/index.tsx
SiAdcock: Actually, maybe not! It doesn't look like TypeScript acknowledges those default props. Perhaps we should move away from default props to use JavaScript default arguments

PR #679, opened 2021-01-04
Checkbox support optional

liywjl:

To show how we can use optional in Checkboxes

## What does this change?

- Document how to use optional prop in checkbox
- Give storybook example

### Optional
Screenshot 2021-01-04 at 18 23 46

### Required
Screenshot 2021-01-04 at 18 23 39

### Accessibility

### Responsiveness

### Documentation

-   [x] Examples in Storybook

### Known issues

We cannot have optional prop for Checkbox Group like we intend to have https://github.com/guardian/source/pull/678

## Why?

HTML5 does not support required for checkbox groups like it does with radio inputs.

### Links
- https://www.w3.org/html/wg/tracker/issues/111?changelog
- https://www.w3.org/Bugs/Public/show_bug.cgi?id=9160#c1


src/core/components/checkbox/README.md
```diff
@@ -94,6 +94,12 @@ If label is omitted, supporting text will not appear either.
 
 Additional text or a component that appears below the label
 
+### `optional`
+
+**`boolean`** _= "false"_
+
+Determine if the user must select this checkbox in order to proceed with submission.
+
```
SiAdcock, line 102: We can probably remove this for now. It's likely that we won't add this feature to the `<CheckboxGroup>` component until we have identified a use case for it.

src/core/components/checkbox/stories/checkbox/required.tsx
```diff
@@ -0,0 +1,44 @@
...
+				/>
+			</CheckboxGroup>
+		</ThemeProvider>
+		<button type="submit">Submit</button>
+	</form>
+)
+
+optional.story = {
+	name: "optional",
+}
```
SiAdcock, line 42: Since optional is the default, we probably don't need another story demonstrating it. Can we remove it?

SiAdcock (commented):
As this stands, it's a useful demo for how devs can enforce a mandatory checkbox. I've made a couple of requests to slim down the PR a bit, but nothing substantial. Thanks for putting this together 👍  💎  ✅

PR #707, opened 2021-02-04
Use new JSX transform for button

jamie-lynch:

This PR updates source to make use of the [new JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) available in React 17.

## What does this change?

- Update `babelrc` files* to set the runtime to `automatic`
- Update rollup config files* to include `react/jsx-runtime` as an external
- Update storybook config to set the runtime to `automatic`
- Update eslint config so that it doesn't flag the lack of `React` import
- Remove `import React from "react"` everywhere

**\* This has only been done for the button component so far**

### Accessibility

### Responsiveness

### Documentation

-   [ ] Examples in Storybook

### DIFF
--- a/src/core/components/button/rollup.config.js
+++ b/src/core/components/button/rollup.config.js
@@ -23,6 +23,7 @@ module.exports = {
 	],
 	external: [
 		"react",
+		"react/jsx-runtime",
 		"@emotion/react",
 		"@emotion/css",
 		"@guardian/src-foundations",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,7 +12,9 @@
 		"prettier/prettier": "error",
 		"@typescript-eslint/explicit-module-boundary-types": "off",
 		"@typescript-eslint/no-unused-vars": "off",
-		"jsx-a11y/label-has-for": "off"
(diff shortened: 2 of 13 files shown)


src/core/components/button/rollup.config.js
```diff
@@ -23,6 +23,7 @@ module.exports = {
 	],
 	external: [
 		"react",
+		"react/jsx-runtime",
```
jamie-lynch, line 26: This is added as otherwise there is an `Unresolved dependencies` warning when building. 

There is currently also a warning for `jsxs,Fragment,jsx imported from external module 'react/jsx-runtime' but never used` whether this is included or not.

jamie-lynch:
In [button/index.tsx](https://github.com/guardian/source/blob/new-jsx-transform/src/core/components/button/index.tsx), `React.cloneElement` is used. When I update the imports to be `import {cloneElement} from "react"` I get a `ReferenceError: React is not defined` in storybook for the `icon-only-buttons` and `icon-only-link-buttons` stories. Will do some more digging into why this is only happening for those cases.

jamie-lynch:
Superseded by #737

PR #713, opened 2021-02-05
Improvements to the Accordion and AccordionRow component

i-hardy:

This tackles a number of small issues encountered recently when trying to utilise the Accordion- the lack of an ability to track clicks on rows, the lack of an ability to pass style overrides, and the AccordionRow button causing issues when the component is used inside a form.

## What does this change?

-   Adds an optional `onClick` callback prop to the AccordionRow component, which receives the click event
-   Adds a `type="button"` attribute to the AccordionRow button, to prevent submit behaviour when used in a form.
- 	Makes the `cssOverrides` prop functional on both the Accordion and AccordionRow components, to enable overriding the default styles
- 
### Documentation

-   [x] Examples in Storybook


src/core/components/accordion/index.tsx
```diff
@@ -48,16 +52,25 @@ interface AccordionRowProps extends Props {
...
 	hideToggleLabel?: boolean;
 	children: ReactNode;
+	onClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
 }
 
 const NoJsRow = ({
 	label,
 	hideToggleLabel = false,
 	children,
+	onClick,
```
SiAdcock, line 62: I don't think we need this prop on the NoJsRow component – it doesn't look like it's being used.

i-hardy: I picked it out as otherwise there was a Typescript error from spreading out the remaining props onto the div (as the onClick type signature is expecting a click event from a button element). If there's a better way to handle this in TS then happy to make any changes!

SiAdcock: Good point! Let's roll with this for now 👍

SiAdcock (approved):
Amazing thanks @i-hardy!

PR #714, opened 2021-02-05
remove vscode settings from source control

sndrs:

to enable personalised setup of vscode

## What does this change?

- removes vscode settings from source control


SiAdcock:
Out of interest are you finding these settings difficult to work with, or is it more like you don't believe the project should dictate settings?

sndrs:
yeah i found the `"files.autoSave": "onFocusChange",` setting v confusing at first, had no idea what was going on :)

it was at the point i thought if you found that useful, you should be able to use it, and i should be able to disable it. only way i could think of was to remove that file from source control?

SiAdcock (dismissed):
👍  Settings freedom for all!

SiAdcock (approved):

sndrs:
interesting... https://github.com/guardian/frontend/blame/842026bc09aa6b86d9786f3484adc71f8108379a/.gitignore#L86

PR #718, opened 2021-02-09
always check visual regressions

sndrs:

always check for visual regressions

## What does this change?

-   runs check whether file appears to be related or not


SiAdcock:
For context, this was to reduce the Chromatic quota back when we had the cheapest possible plan! Maybe less cause for concern these days 💸

PR #719, opened 2021-02-11
One package

sndrs:

to introduce `@guardian/source` and `@guardian/editorial` packages that will replace the more atomic ones. 

## What does this change?

_note that this leaves the existing `src`/`@guardian/src-*` stuff/code/scripts etc in place, incase we need to do a quick release_

-   creates `packages` workspace
-   copies the necessary source code from `src/core` to `packages/source` on demand, at build time
  - updates import paths to use local files, rather than packages
-   copies the necessary source code from `src/editorial` to `packages/editorial` in source control
  - it's simpler and easier to do
- adds a build task to build both packages

almost all the additions are just copying files over, and making minimal updates to them. 

## Is this ready to publish?

nope!

it isn't tested and not release-ready.

~~also, any changes to `src` that need to go out before this is ready to be the main release mechanism will need to be duplicated.~~

it's probably easier to go over this f2f2f in an editor than as a diff :)


packages/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "types": ["@emotion/react/types/css-prop", "@types/jest"],
+    "paths": {
+      "@guardian/src-foundations": ["source/foundations/src"],
+      "@guardian/src-foundations/*": ["source/foundations/src/*"],
+      "@guardian/src-foundations/types/themes": [
+        "source/foundations/types/themes"
+      ],
+      "@guardian/src-helpers": ["source/helpers"],
+      "@guardian/src-brand": ["source/brand"],
+      "@guardian/src-components/*": ["source/components/*"],
```
SiAdcock, line 21: Not sure if we are using this?

packages/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "baseUrl": ".",
+    "allowSyntheticDefaultImports": true,
+    // "isolatedModules": true,
+    "jsx": "preserve",
+    "module": "ES2020",
+    "moduleResolution": "node",
+    "noUnusedLocals": true,
+    "strict": true,
+    "target": "ES2020",
+    "types": ["@emotion/react/types/css-prop", "@types/jest"],
```
SiAdcock, line 12: Is this it necessary to specify the `css-prop` types? We are already adding this reference to the top of each component index file:

```
///<reference types="@emotion/react/types/css-prop" />
```

This ensures consuming apps don't need to add the `css-prop` type to their `tsconfig`

rollup.config.js
```diff
@@ -0,0 +1,37 @@
...
+				],
+			],
+			plugins: ['@babel/plugin-transform-runtime'],
+		}),
+	],
+	external: [
+		'@emotion/react',
+		'@emotion/react/jsx-runtime',
+		'@guardian/src-foundations',
+		'@guardian/src-foundations/palette',
```
SiAdcock, line 34: I don't think we need to specify `@guardian/src-foundations` as an external dependency anymore.

SiAdcock:
Heads up: I'm getting changes to `yarn.lock` after running `yarn`.

rollup.config.js
```diff
@@ -0,0 +1,37 @@
...
+	input: {
+		editorial: 'packages/editorial/index.ts',
+		source: 'packages/source/index.ts',
+	},
+	output: {
+		format: 'esm',
+		sourcemap: true,
+		dir: 'dist',
+		entryFileNames: '[name]/index.js',
+	},
```
SiAdcock, line 14: Are we dropping support for CommonJS?

packages/tsconfig.json
sndrs: ah nice, i didn't realise that. how does that work when it's bundled?

rollup.config.js
sndrs: yeah it's still being used in editorial currently

rollup.config.js
sndrs: do we need to support it? we can add it, i wondered about it. doesn't anyone use it? ought they to be if they are?

SiAdcock:
This looks awesome, @sndrs, I love the new structure and your approach to migrating. Duplication in the short term is probably the best way of doing it. 

I time-boxed this review, so there might be other things we'll have to fix along the way.

I think we should publish a pre-release and regression test this against [source-shakedown](https://github.com/guardian/source-shakedown) to make sure the old ways still work. We should then test the new API against either source-shakedown or somewhere else, open to suggestions 😄 

Begun,  [The Flattening](https://github.com/guardian/source/discussions/688#discussioncomment-278764) has.

packages/tsconfig.json
sndrs: > Triple-slash directives are only valid at the top of their containing file.

should we make sure we prepend it to the build files?

SiAdcock: > how does that work when it's bundled
> should we make sure we prepend it to the build files

I don't think it matter when it's bundled, because the output is JS. It's only useful in TS files

rollup.config.js
SiAdcock: Providing a CJS endpoint is one of the ways we can get Jest to pick up Source components in consuming applications' test suite. We document this here:

https://theguardian.design/2a1e5182b/p/863906-faqs/t/488531

There might be a better way of doing it (e.g. transpiling a subset of `node_modules` through `babel-jest`).

There were a couple of purely node projects about a year ago that wanted to consume Source, but maybe they've been put out to pasture.

sndrs:
updated how this works, will update the description

sndrs:
i think this was a second misstep, but it's getting there. another version is coming...

PR #721, opened 2021-02-12
fix type name clashes that apply in flat land

sndrs:

prevent type name clashes when we have a flat package

## What does this change?

- rename instances of types with same names


SiAdcock (approved):
Good catch 👍

PR #726, opened 2021-02-15
Support responsive column widths

### LINKED ISSUE #701: Responsive column widths

Column widths are not responsive. i.e. it's not possible for a column to take up 1/4 of the container width at desktop, but 1/2 the width at tablet

It would be useful (and indeed 🆒) to create an API that handles that

 e.g. `<Column width={[1, 1/2]}>`

SiAdcock:

It should be possible to specify the width of a column at various breakpoints, rather than the width being static across viewport widths.

It should also be possible to completely hide a column at certain breakpoints.

Fixes #701 

## What does this change?

-   Extend width API to accept an array of widths
-   A zero-width column should be hidden from the page

## Screenshots

**Responsive widths**

```tsx
<Columns>
	<Column width={[1 / 2, 1 / 4]}>
		<p>50% at mobile, 25% above tablet</p>
	</Column>
	<Column width={[1 / 2, 3 / 4]}>
		<p>50% width at mobile, 75% above tablet</p>
	</Column>
</Columns>
```

**Responsive hiding**

```tsx
<Columns>
	<Column width={[0, 1 / 4]}>
		<p>Not visible at mobile, 25% above tablet</p>
	</Column>
	<Column width={[1, 3 / 4]}>
		<p>100% at mobile, 75% above tablet</p>
	</Column>
</Columns>
```

### Accessibility

### Responsiveness

### Documentation

-   [x] Examples in Storybook


tomrf1 (approved):
looks great!

src/core/components/layout/components/columns/styles.ts
```diff
@@ -1,18 +1,18 @@
...
-	& > * + * {
+	& > div:nth-of-type(n + 2) {
 		margin-left: ${space[5]}px;
 	}
 `;
 
 const collapseBelowSpacing = css`
 	display: block;
-	& > * + * {
+	& > div:nth-of-type(n + 2) {
```
jamie-lynch, line 15: I don't quite follow this change. Would you mind explaining what the difference is?

jamie-lynch (approved):

src/core/components/layout/components/columns/styles.ts
SiAdcock: Oh yes, I was messing around with something, but I didn't need it in the end. I'll restore the old way. Thanks for asking!

src/core/components/layout/components/columns/columns.tsx
```diff
@@ -69,17 +69,12 @@ const Columns = ({
...
 	children: ReactNode;
 }
 
-const Column = ({
-	width = 0,
-	cssOverrides,
-	children,
-	...props
-}: ColumnProps) => {
+const Column = ({ width, cssOverrides, children, ...props }: ColumnProps) => {
```
sndrs, line 77: if `width` no longer has a default value of `0`, is this a breaking change?

jamie-lynch: If you were relying on the default before then I think you still get the same behaviour (due to [these changes](https://github.com/guardian/source/pull/726/files#diff-61a31ebdb74d46042d5af57ee18f291987ca2be4bea6ebc25d0ef613c5678e48R141-R142)) but if you're explicitly passing 0 you will now get different behaviour so maybe it's a breaking change? Also, if you pass an empty array I don't think it's applying any styling. Maybe that should be the same as not passing a width?

SiAdcock: @jamie-lynch is correct, if you were explicitly passing `width={0}` before, the behaviour would have been the same as if you didn't pass a width (i.e. the column's width would be the remainder of container width shared between all zero-width columns). This behaviour was not documented. Arguably the behaviour is bizarre and unexpected.

The new behaviour is to hide a column that has a 0 width. I think this is more intuitive.

Although this is technically a breaking change, I'm inclined not to treat it as such from a semver / release perspective. Nobody is currently relying on this behaviour. Let's close this loophole before people start relying on it.

SiAdcock: Good catch with the empty array 🙏  

I've updated the logic to treat an empty array the same as passing not passing a width. I hope that makes sense!

PR #716, opened 2021-02-05
WIP – flat package

sndrs:
## What is the purpose of this change?

try to reshape the repo around the structure that will be in place when we only publish `@guardian/source` and `@guardian/editorial`

## What does this change?

-   renames `./src` to `./packages`
-   renames `./src/core` to `./packages/source`

PR #727, opened 2021-02-15
Flatpack

sndrs:

to create a `@guardian/source` package, hopefully superseding #719 and #716 

## What does this change?

- adds a `@guardian/source` package at `src/core`
- adds a build step that creates node-10 compatible commonjs version, an ES2020 esm version, and typescript types
- updates some source file to explicitly export types of things they were exporting

### Known issues

- this doesn't include publishing or versioning `@guardian/source` – to do in next steps
- it's not part of the toolchain for existing `@guardian/src-*` packages
  - this is intentional though – `np`/`lerna`/`semantic-release` will be a better option
  - its simpler to add those as we remove `@guardian/src-*` or after (one step at a time)

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   ],
   "scripts": {
     "build-storybook": "build-storybook -o dist",
-    "build:all": "ts-node ./scripts/build-all",
+    "build:all": "ts-node ./scripts/build-all && yarn workspace @guardian/source build",
     "ci:build": "NODE_ENV=production build-storybook -o dist",
     "clean": "rm -rf dist",
     "clean:all": "ts-node ./scripts/clean-all",
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "license": "Apache-2.0",
   "workspaces": [
+    "src/*",
     "src/core/*",
     "src/core/components/*",
     "src/editorial/web/*",
--- a/src/core/components/accordion/index.tsx
+++ b/src/core/components/accordion/index.tsx
@@ -25,7 +25,7 @@ const visuallyHidden = css`
 	${_visuallyHidden}
 `;
 
-interface AccordionProps extends Props {
+export interface AccordionProps extends Props {
 	hideToggleLabel?: boolean;
 	children: ReactElement[];
 }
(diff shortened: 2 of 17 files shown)


package.json
```diff
@@ -4,14 +4,15 @@
...
+    "src/*",
     "src/core/*",
     "src/core/components/*",
     "src/editorial/web/*",
     "src/editorial/web/components/*"
   ],
   "scripts": {
     "build-storybook": "build-storybook -o dist",
-    "build:all": "ts-node ./scripts/build-all",
+    "build:all": "ts-node ./scripts/build-all && yarn workspace @guardian/source build",
```
SiAdcock, line 15: Oooh that `yarn workspace` command is a nice touch

src/core/index.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * @guardian/source
+ */
+
+// we need to be explicit about the index file while these directories are also packages,
+// otherwise the node require algorithm uses the package.json
```
SiAdcock, line 6:
```suggestion
// otherwise the node require algorithm uses the package.json
// TODO: we can drop the index file when we remove the package.jsons
```

SiAdcock (dismissed):
This is a great first step, thanks @sndrs. Since it won't impact anything currently consuming Source, we can get away with shipping this, then try publishing and testing in a repo that's already using Source v3 (e.g. [gateway](https://github.com/guardian/gateway))

src/core/components/accordion/index.tsx
```diff
@@ -25,7 +25,7 @@ const visuallyHidden = css`
 	${_visuallyHidden}
 `;
 
-interface AccordionProps extends Props {
+export interface AccordionProps extends Props {
```
SiAdcock, line 28: Do we need to export these Props interfaces in order for consuming projects to understand what props are needed for each component?

sndrs: i think so yeah

sndrs: typescript was complaining, i guess because the rollup build was combining all the exports in one place?

SiAdcock: Right. I guess if TypeScript is complaining, it should prevent us from accidentally removing that `export` keyword in the future. Just thinking that this looks like a Chesterton's Fence in waiting!

SiAdcock (approved):

sndrs:
superseded by #731

PR #729, opened 2021-02-16
Supporting hiding above and below breakpoints

SiAdcock:

Currently it's only possible to hide a component above a certain breakpoint or below a certain breakpoint. It's not possible to do both at the same time.

## What does this change?

-   Compose below and above css to support both props at the same time

### DIFF
--- a/src/core/components/layout/stories/hide/default.tsx
+++ b/src/core/components/layout/stories/hide/default.tsx
@@ -33,3 +33,17 @@ export const above = () => (
 above.story = {
 	name: 'above',
 };
+
+export const between = () => (
+	<Container>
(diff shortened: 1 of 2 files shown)


tomrf1 (dismissed):
thanks!

src/core/components/layout/stories/hide/default.tsx
```diff
@@ -33,3 +33,17 @@ export const above = () => (
 above.story = {
 	name: 'above',
 };
+
+export const aboveAndBelow = () => (
```
sndrs, line 37: is `between` simpler?

sndrs (approved):
lovely

PR #735, opened 2021-02-19
update root tsconfig

sndrs:

set TS to build for ES2015

## What does this change?

- updates the root `tsconfig` to output `es2015`, matching the current babel setup
- make sure it doesn't output anything running `yarn tsc` from root (just check types)


sndrs:
no

PR #736, opened 2021-02-19
TS only icons

sndrs:

build `src-icons` using just `tsc`

## What does this change?

- removes babel and rollup in build steps for icons


sndrs:
supplanted by #737

PR #737, opened 2021-02-19
3.3 TSC

sndrs:

builds each package using `tsc`

## What does this change?

- removes rollup/babel toolchain
- uses `tsc` only
- ensures all packages are their own typescript project (not all were)
- removes unused `react` imports

I tried to do this piecemeal, but the compiler complained unless every project was setup correctly, with references etc. 

This doesn't include any fixes etc or other changes to the source code. Afaict, running `yarn tsc` from root now checks everything. 

Using `tsc` to compile has the added benefit that you cannot build if type-checking fails.


package.json
```diff
@@ -25,7 +25,7 @@
     "release:patch:all": "yarn build:all && yarn validate && yarn verbump:patch:all && yarn publish:release",
     "release:prerelease-bump:all": "yarn build:all && yarn validate && yarn verbump:prerelease:all && yarn publish:release",
     "release:prerelease:all": "yarn build:all && yarn validate && yarn publish:release",
-    "storybook": "yarn watch:foundations & yarn watch:icons & start-storybook",
+    "storybook": "start-storybook",
```
SiAdcock, line 28: The reason this is like this is because all components depend on `@guardian/src-foundations`, and some depend on `@guardian/src-icons`. This means changes to foundations are not picked up by Storybook, because the components depend on the the `dist` code under src/core/foundations. Does that make sense? Maybe there's some way of fixing this with path remapping somewhere.

src/core/components/user-feedback/package.json
```diff
@@ -30,20 +34,11 @@
...
-    "rollup-plugin-babel": "^4.3.3",
-    "rollup-plugin-commonjs": "^10.0.2",
-    "rollup-plugin-node-resolve": "^5.2.0"
+    "npm-run-all": "^4.1.5",
+    "typescript": "^4.1.3"
   },
   "peerDependencies": {
     "@emotion/react": "^11.1.2",
-    "@guardian/src-foundations": "^3.3.0-rc.0",
-    "react": "^17.0.1"
```
SiAdcock, line 47: Just to clarify, are we saying that consumers no longer need to bring React in order to get Source working?

SiAdcock:
This is pretty awesome! 💯 

As I discovered, and @jamie-lynch before me, removing `React` from the scope of some modules causes problems because we're sometimes referencing parts of the React API from the React import e.g. `React.Children` or `React.cloneElement`. I also found that Fragments don't work if React is not in scope, maybe there's some tsconfig that can take care of that.

PR #760, opened 2021-03-04
Responsive columns based on number of grid columns spanned

rhystmills:

This PR enables developers to specify the number of columns they want a `Column` component to span. They can also specify an array of spans for different breakpoints.

The aim is to simplify conformance to our [conceptual 'grid'](https://theguardian.design/2a1e5182b/p/41be19-grids), where we use progressively fewer columns as the screensize falls, for example:

| Wide | Mobile |
| --- | --- |
| image |  image|

## What does this change?

- Adds a `span` prop to our Column component. This can be specified as a single value or an array covering different breakpoints. The result can be seen below.

- Under the hood, a single `span` value adds multiple media queries to an element, because each column has a different meaning at different breakpoints, for example 4 columns are ~100% width at _Mobile_ size but only ~ 25% width at _Wide_ size. In terms of total columns, the significant breakpoints are:

| Breakpoint | Total columns |
| --- | --- |
| Wide | 16 |
| LeftCol | 14 |
| Desktop & Tablet | 12 |
| <= Phablet | 4 |

- Specifying an array of values adds additional overrides for breakpoints as is currently done for the `width` prop for Columns.

- I've also added a story to demonstrate the `span` prop.

## Checklist

_I plan to do some more accessibility testing and test the component with longer text, but wanted to get some PR feedback at this stage. Documentation would also be sensible as the interface is non-obvious._

### Accessibility

-   [X] [The component doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/61524d)

### Responsiveness

### Documentation

-   [X] Examples in Storybook


src/core/components/layout/components/columns/styles.ts
```diff
@@ -135,7 +137,62 @@ const generateWidthCSS = (width: number | number[]) => {
...
+
+	return css`
+		box-sizing: border-box;
+		flex: ${span ? '0 0 auto' : 1};
+		${columnBreakpointCss}
+	`;
+};
+
+export const column = (width?: number | number[], span?: number | number[]) => {
+	if (!width && (span || span === 0)) {
```
SiAdcock, line 192: I'd be inclined to split this function up and push all this logic into the component. What do you think about:

```ts
// always gets applied
export const column = css`
	box-sizing: border-box;
`;

// get applied when no width or span is specified
export const flexGrow = css`
	flex: 1;
`;

// width is specified
export const width = (value: number | number[]) => css`
	flex: 0 0 auto;
	${generateWidthCSS(value)};
`;

// span is specified
export const span = (value: number | number[]) => css`
	flex: 0 0 auto;
	${generateSpanCSS(value)};
`;
```

src/core/components/layout/components/columns/styles.ts
```diff
@@ -135,7 +137,62 @@ const generateWidthCSS = (width: number | number[]) => {
...
+];
+
+const calculateSpan = (span: number) => {
+	const columnBreakpointCss = columnBreakpoints.reduce(
+		(acc: string, cur: ColumnBreakpoint) => {
+			if (span === 0) return `display: none;`;
+			if (span >= cur.totalColumns) return `width: 100%; display: block;`;
+
+			const inferredWidth = span / cur.totalColumns;
+			const cssForBreakpoint = `{ width: calc((100% + ${space[5]}px) * ${inferredWidth} - ${space[5]}px); display: block; }\n`;
```
SiAdcock, line 177: Not sure, but is it possible to use the existing [`calculateWidth`](https://github.com/guardian/source/blob/237dcb06cb549af69036f194560dc224288ef2ec/src/core/components/layout/components/columns/styles.ts#L93) function to do this?

src/core/components/layout/stories/columns/with-span.tsx
```diff
@@ -0,0 +1,150 @@
...
+			</p>
+		</Container>
+
+		<Container>
+			<Columns>
+				<Column span={2} width={3 / 4}>
+					<div css={contents}>span: 2, width: 1/4</div>
+				</Column>
+				<Column width={1 / 4}>
+					<div css={contents}>width: 3/4</div>
```
SiAdcock, line 139:
```suggestion
					<div css={contents}>width: 1/4</div>
```
```suggestion
					<div css={contents}>width: 3/4</div>
```

src/core/components/layout/stories/columns/with-span.tsx
```diff
@@ -0,0 +1,150 @@
...
+			<p>
+				<code css={code}>span</code> is overruled by
+				<code css={code}>width</code> prop:
+			</p>
+		</Container>
+
+		<Container>
+			<Columns>
+				<Column span={2} width={3 / 4}>
+					<div css={contents}>span: 2, width: 1/4</div>
```
SiAdcock, line 136:
```suggestion
					<div css={contents}>span: 2, width: 3/4</div>
```
```suggestion
					<div css={contents}>span: 2, width: 1/4</div>
```

src/core/components/layout/components/columns/styles.ts
rhystmills: Great idea, I seem to have made it work.

src/core/components/layout/components/columns/styles.ts
rhystmills: That looks much cleaner! I've added it now.

src/core/components/layout/components/columns/styles.ts
```diff
@@ -135,23 +137,84 @@ const generateWidthCSS = (width: number | number[]) => {
...
+	{ totalColumns: 16, rule: from.wide },
+];
+
+const calculateSpan = (span: number) => {
+	const columnBreakpointCss = columnBreakpoints.reduce(
+		(acc, cur: ColumnBreakpoint) => {
+			if (span === 0)
+				return css`
+					display: none;
+				`;
```
SiAdcock, line 176: I think this is taken care of in [`calculateWidth`](https://github.com/guardian/source/blob/bd5a9c0486eeddbdbf258c164f1d4c5173acf0d5/src/core/components/layout/components/columns/styles.ts#L94-L104). Are you able to confirm?

src/core/components/layout/components/columns/styles.ts
```diff
@@ -135,23 +137,84 @@ const generateWidthCSS = (width: number | number[]) => {
...
+`;
+
+// get applied when no width or span is specified
+export const flexGrow = css`
+	flex: 1;
+`;
+
+// width is specified
+export const setWidth = (value: number | number[]) => css`
+	flex: 0 0 auto;
```
SiAdcock, line 212: I know you didn't introduce this, but can it be optimised to `flex: 0`?

src/core/components/layout/components/columns/styles.ts
```diff
@@ -135,23 +137,84 @@ const generateWidthCSS = (width: number | number[]) => {
...
+			`;
+		},
+		css``,
+	);
+
 	return css`
 		box-sizing: border-box;
-		flex: ${flex};
-		${widthCSS};
+		flex: ${span ? '0 0 auto' : 1};
```
SiAdcock, line 195: Do we still need to specify `box-sizing` and `flex` properties here? I expect they're taken care of by other style snippets.

src/core/components/layout/components/columns/styles.ts
rhystmills: I found that it was generating some unexpected behaviour - specifically generating some blank space in the page at lower breakpoints (see below). I assumed this was some quirk of how `display: none` and `visibility: hidden` work - do you have any idea?
image

rhystmills: This also highlights that these changes overall are generating absurdly long class names which is bad (especially when multiple breakpoints are being specified) -  and would probably have a performance impact if it were used for lots of elements on the page.

src/core/components/layout/components/columns/styles.ts
rhystmills: `flex: 0` caused some problems but `flex: auto` seemed to work - though I'd be a bit reluctant to change it in this PR without doing some proper testing around it.

This did make me realise that the `flex` property in setSpan below is redundant.

src/core/components/layout/components/columns/styles.ts
rhystmills: Good point! I've removed them now.

src/core/components/layout/components/columns/styles.ts
SiAdcock: I've merged [a fix](https://github.com/guardian/source/pull/772) to use `display: none` instead of `visibility: hidden`. This removes the blank space and some complexity. 

I think we're going to hit issues with the `margin-right` being added to all but the last column. For instance what do we do if the last column is hidden? We'll need to find a way of removing the `margin-right` from the last _visible_ column 🤔

src/core/components/layout/components/columns/styles.ts
```diff
@@ -117,35 +122,99 @@ const generateWidthCSS = (width: number | number[]) => {
...
+			const inferredWidth = span / cur.totalColumns;
+			const cappedWidth = inferredWidth < 1 ? inferredWidth : 1;
+			const cssForBreakpoint = calculateWidth(cappedWidth);
+
+			return `
+				${acc};
+				${cur.rule} {
+					${cssForBreakpoint};
+					display: block;
+				}
```
SiAdcock, line 190: I don't think we need the media query here as it's already being generated in [`generateSpan`](https://github.com/guardian/source/pull/760/files#diff-61a31ebdb74d46042d5af57ee18f291987ca2be4bea6ebc25d0ef613c5678e48R156). Does `calculateSpan()` need to receive the `totalColumns`?

SiAdcock (changes requested):
Thanks for your work on this @rhystmills, I think we're getting close. I've left a couple of requests and there's also a question mark around what happens when the final column in a row is hidden... how do we remove the `margin-right` from the final _visible_ column.

src/core/components/layout/components/columns/styles.ts
rhystmills: (For the sake of any PR reviewers - Simon and I added a negative margin to the container so that we didn't need to identify non-visible final columns)

src/core/components/layout/components/columns/styles.ts
rhystmills: If we're talking about the same thing (`cur.rule` above)- I think we may still need this one - what this does is reflect the fact that _n_ columns have a different meaning for the width at different breakpoints. E.g. '2' columns means 50% of the screen at mobile size (where there are 4 columns total) but only ~16% at desktop size (where there are 12 total columns). 

Essentially a rule saying that an element should span '2' columns is really generating 4 sets of rules - one for each column-relevant breakpoint ( <tablet, tablet & desktop, leftCol & wide). Without `cur-rule` - '2' columns would be treated the same at every breakpoint.

e.g. removing `cur.rule`:
https://media.giphy.com/media/M2pTPdvNNDopNI4ZBP/giphy.gif

SiAdcock: I see. Isn't that what [this media query](https://github.com/guardian/source/pull/760/files#diff-61a31ebdb74d46042d5af57ee18f291987ca2be4bea6ebc25d0ef613c5678e48R148) ought to be doing? At the moment it looks like the media query in `generateSpan` is wrapping the one in `calculateSpan`.

Currently if you `console.log()` the output of `generateSpanCSS([0, 0, 2])`, you get CSS that looks like this:

```css
@media (min-width: 320px) {
	@media (max-width: 739px) {
		display: none;
	}
	@media (min-width: 740px) and (max-width: 1139px) {
		display: none;
	}
	@media (min-width: 1140px) and (max-width: 1299px) {
		display: none;
	}
	@media (min-width: 1300px) {
		display: none;
	}
}
```

I guess I'm wondering if this output could be streamlined?

rhystmills: Good spot Simon - I may have to do some more thinking about that.

I think the recently removed - 
``` 
if (span === 0)
  return `
    display: none;
  `; 
```

- might have exacerbated this particular case (I'm not totally sure without testing though). But there would probably be unnecessary queries in other cases too, e.g. wide breakpoint rules for a mobile breakpoint specified in an array of breakpoint values, e.g. `[0, 0, 2]`

I think streamlining this will be quite complicated - I'll have a look early next week.

SiAdcock: Okay, thanks @rhystmills. If you don't have time or it looks like too big a task, we can put a pin in it for now and circle back around. I think making the API available is more important than implementing perf optimisations right now 🙂

rhystmills: Re-adding the `return 'display: none'` does help with that particular case (multiple nested `display: none` rules), and reduces the amount of generated css overall - so I have added it back in.

I think there are definitely other cases where smarter logic could help remove redundant rules. 

```css
@media (min-width: 320px) {
display: none;
}

@media (min-width: 740px) {
display: none;
}

@media (min-width: 980px) {
    @media (max-width: 739px) {
      width: calc((100%) * 0.5 - 20px);
      /* Reset value that might have been set at a lower breakpoint */
      display: block;
    }

    @media (min-width: 740px) and (max-width: 1139px) {
      width: calc((100%) * 0.16666666666666666 - 20px);
      /* Reset value that might have been set at a lower breakpoint */
      display: block;
    }

    @media (min-width: 1140px) and (max-width: 1299px) {
      width: calc((100%) * 0.14285714285714285 - 20px);
      /* Reset value that might have been set at a lower breakpoint */
      display: block;
    }

    @media (min-width: 1300px) {
      width: calc((100%) * 0.125 - 20px);
      /* Reset value that might have been set at a lower breakpoint */
      display: block;
    }
}
```
For example above we can see:
- Two consecutive `display: none` breakpoints that could be merged into one
- A `@max-width: 739px` media query that can never have an effect because it's within a `@min-width: 980px` media query.

Also - console logging the css strings shows they have a lot of whitespace in it (as a consequence of multiline css strings in the code), it might be nice to make more readable outputted css by changing how we do this in the future.

rhystmills: Perhaps those performance improvements would be good for a future PR?

src/core/components/layout/components/columns/styles.ts
```diff
@@ -1,22 +1,25 @@
...
 	& > * {
 		margin-right: ${space[5]}px;
 	}
 `;
 
 const collapseBelowSpacing = css`
 	display: block;
+	margin-right: 0;
 	& > * {
-		margin-right: 0;
```
SiAdcock, line 19: I think we want to keep this zero margin for columns when collapsed

src/core/components/layout/components/columns/styles.ts
SiAdcock: Thanks for investigating, solid work 💪 

I agree we should park these optimisations for now and decouple them from this PR. We're already covering a lot of ground.

src/core/components/layout/components/columns/styles.ts
rhystmills: Sorted!

SiAdcock (approved):
Thanks so much for sticking with this and getting it over the line @rhystmills! I really appreciate the dedication and the end result is awesome! 💎  👍  🥇

PR #776, opened 2021-03-19
Configurable column spacing

jamie-lynch:

It's possible to collapse columns into a stack below a certain breakpoint using the collapseBelow prop. This PR also makes the vertical space between collapsed columns configurable, similar to how the Stack component works.

## What does this change?

-   Add the `spaceY` prop to the `Columns` component
-   Updating the styling to provide the requested spacing

## Screenshots

**Before**

**After**

### Accessibility

### Responsiveness

### Documentation

-   [ ] Examples in Storybook

The docs site will need to be updated to add the new Prop

### DIFF
--- a/src/core/components/layout/components/columns/columns.tsx
+++ b/src/core/components/layout/components/columns/columns.tsx
@@ -25,10 +25,13 @@ type CollapseBreakpoint = Extract<
 	'tablet' | 'desktop' | 'leftCol' | 'wide'
 >;
 
+export type ColumnsSpaceY = 1 | 2 | 3 | 4 | 5 | 6 | 9 | 12 | 24;
+
 interface ColumnsProps extends HTMLAttributes<HTMLDivElement>, Props {
 	collapseBelow?: CollapseBreakpoint;
 	cssOverrides?: SerializedStyles | SerializedStyles[];
 	children: ReactNode;
+	spaceY?: ColumnsSpaceY;
 }
 
 const collapseBelowMap: { [key in CollapseBreakpoint]: SerializedStyles } = {
@@ -39,7 +42,7 @@ const collapseBelowMap: { [key in CollapseBreakpoint]: SerializedStyles } = {
 };
 
 const collapseBelowColumnsMap: {
-	[key in CollapseBreakpoint]: SerializedStyles;
+	[key in CollapseBreakpoint]: (number: ColumnsSpaceY) => SerializedStyles;
 } = {
 	tablet: collapseBelowTabletColumns,
 	desktop: collapseBelowDesktopColumns,
@@ -51,13 +54,16 @@ const Columns = ({
 	collapseBelow,
 	cssOverrides,
 	children,
+	spaceY,
 	...props
 }: ColumnsProps) => {
 	return (
 		<div
 			css={[
 				columns,
-				collapseBelow ? collapseBelowColumnsMap[collapseBelow] : '',
+				collapseBelow
+					? collapseBelowColumnsMap[collapseBelow](spaceY || 5)
+					: '',
 				collapseBelow ? collapseBelowMap[collapseBelow] : '',
 				cssOverrides,
 			]}
--- a/src/core/components/layout/components/columns/columns.tsx
+++ b/src/core/components/layout/components/columns/columns.tsx
@@ -42,7 +42,7 @@ const collapseBelowMap: { [key in CollapseBreakpoint]: SerializedStyles } = {
 };
 
 const collapseBelowColumnsMap: {
-	[key in CollapseBreakpoint]: (spaceY: ColumnsSpaceY) => SerializedStyles;
+	[key in CollapseBreakpoint]: (spaceY?: ColumnsSpaceY) => SerializedStyles;
 } = {
 	tablet: collapseBelowTabletColumns,
 	desktop: collapseBelowDesktopColumns,
@@ -54,7 +54,7 @@ const Columns = ({
 	collapseBelow,
 	cssOverrides,
 	children,
-	spaceY = 5,
+	spaceY,
 	...props
 }: ColumnsProps) => {
 	return (
--- a/src/core/components/layout/components/columns/columns.tsx
+++ b/src/core/components/layout/components/columns/columns.tsx
@@ -54,7 +54,7 @@ const Columns = ({
 	collapseBelow,
 	cssOverrides,
 	children,
-	spaceY,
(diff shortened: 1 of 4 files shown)


src/core/components/layout/components/columns/columns.tsx
```diff
@@ -51,13 +54,16 @@ const Columns = ({
 	collapseBelow,
 	cssOverrides,
 	children,
+	spaceY = 5,
```
jamie-lynch, line 57: This preserves the current behaviour by default but means that the default behaviour is different to the `Stack` component. I'm not sure which is preferable.

src/core/components/layout/components/columns/styles.ts
```diff
@@ -13,32 +14,32 @@ export const columns = css`
...
 	${until.leftCol} {
-		${collapseBelowSpacing}
+		${collapseBelowSpacing(spaceY)}
 	}
 `;
-export const collapseBelowWideColumns = css`
+export const collapseBelowWideColumns = (spaceY: ColumnsSpaceY) => css`
 	${until.wide} {
-		${collapseBelowSpacing}
+		${collapseBelowSpacing(spaceY)}
```
jamie-lynch, line 42: It feels like there's a nicer way of doing this but I'm not really sure what it is.

src/core/components/layout/stories/columns/collapse-below.tsx
```diff
@@ -28,9 +36,97 @@ export const collapseBelowTablet = () => (
...
+	</Container>
+);
+
+export const collapseBelowTabletWithSpace1 = () =>
+	collapseBelowTabletWithSpace(1);
+
+collapseBelowTabletWithSpace1.story = {
+	...defaultStoryOptions,
+	name: 'collapse below tablet with space 1',
+};
```
jamie-lynch, line 68: Not sure if there's a better way, with less copying-and-pasting. Also, this clutters up the stories for columns a bit. Any thoughts on how else this could work?

src/core/components/layout/components/columns/columns.tsx
SiAdcock: I think we ought to replicate Stack behaviour here for consistency. We can afford to make a "breaking" change because nobody is using this behaviour yet. Arguably zero space between rows is the correct and expected behaviour.

jamie-lynch: Cool, agreed. 👍

src/core/components/layout/components/columns/styles.ts
jamie-lynch: I've refactored this in https://github.com/guardian/source/pull/776/commits/1a6b4394b939865971a7de60a28fa942661c56f8 so that there's not one function which takes the breakpoint and the `spaceY` value. Something similar could work with the `collapseBelowMap` if this is a pattern that we prefer?

src/core/components/layout/stories/columns/collapse-below.tsx
jamie-lynch: I've tidied this up slightly so that the `no space` story follows the same pattern. Maybe this is fine as-is now that it's consistent?

src/core/components/layout/components/columns/styles.ts
```diff
@@ -16,33 +17,21 @@ export const columns = css`
...
-`;
-export const collapseBelowWideColumns = css`
-	${until.wide} {
-		${collapseBelowSpacing}
+export const collapseBelowColumns = (
+	breakpoint: CollapseBreakpoint,
+	spaceY?: ColumnsSpaceY,
+) => css`
+	${until[breakpoint]} {
+		${collapseBelowSpacing(spaceY)}
```
SiAdcock, line 34: I like to avoid using dynamic styles if possible. Dynamic styles can't be determined at build time and therefore will always require a runtime to be shipped to the client. In the future, I'd love for us to move away from Emotion to a solution like [Treat](https://seek-oss.github.io/treat/) or [Linaria](https://github.com/callstack/linaria) that doesn't require a runtime, and doesn't support dynamic styles.

To achieve this, we can use a pattern similar to that in the [Stack styles](https://github.com/guardian/source/blob/4c2c1baaa9e2f34576d1f7d206baf6df7a26a826/src/core/components/layout/components/stack/styles.ts#L17-L37): we expose a map (or just a plain object) with breakpoints for keys and calls to a function that generates the CSS. This keeps duplication low, but means the styles can (theoretically at least) be determined at build time.

```ts
const collapseBelowColumnsCSS = (breakpoint: Breakpoint) => css`
  ${until[breakpoint]} {
    display: block;
    margin-right: 0;
    & > * {
      margin-right: 0;
    }
  }
` ;

export const collapseBelowColumns = {
  tablet: collapseBelowColumnsCSS('tablet'),
  //...
};
```

What about spacing? We can almost lift this straight out the Stack styles

```ts
const collapseBelowSpaceYCSS (spaceY: ColumnsSpaceY) => css`
  & > * {
    margin-bottom: ${space[spaceY]}px;
  }
`;

export const collapseBelowSpaceY: {
	1: SerializedStyles;
	2: SerializedStyles;
} = {
	1: collapseBelowSpaceYCSS(1),
	2: collapseBelowSpaceYCSS(2),
    //...
};
```

But won't this apply the spaceY margin at every breakpoint? Uhhh... yeah, good point! 😅  Maybe we can correct for this also applying a negative bottom margin to the column component?

```ts
const collapseBelowSpaceYCSS (spaceY: ColumnsSpaceY) => css`
  margin-bottom: ${-space[spaceY]}px;
  & > * {
    margin-bottom: ${space[spaceY]}px;
  }
`;
```

Let me know what you think 😄

src/core/components/layout/stories/columns/collapse-below.tsx
SiAdcock: This is 👌  Thanks for taking the time to keep this tidy ❤️

src/core/components/layout/components/columns/styles.ts
jamie-lynch: That makes sense, and thanks for the detailed comment. 👍 

Hopefully https://github.com/guardian/source/pull/776/commits/a3fbea5aa5eb5a682e63536363d8842b344c5934 addresses this.

SiAdcock (approved):
Wow this is awesome, thanks so much @jamie-lynch! 💎 

I think those Chromatic diffs are an artefact and can be ignored. Please feel free to squash and merge

PR #782, opened 2021-03-25
Dashed lines (for labs)

mxdvl:

Implement dashed lines, which will help render Labs, with its mix of straight and dashed lines.

## What does this change?

- Adds dashed lines
- Extends LineCount to accept `1`
- Reorder stories to make it easier to parse.

## Screenshots

storybook lines

### Accessibility

-   [X] [The component doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/61524d)

### Responsiveness

-   [X] Tested with with long text

### Documentation

-   [X] Full API surface area is documented in the README
-   [X] Examples in Storybook

### Known issues

There is a possibility that the dotted line does not perfectly match up with either side of the dashed line.

mismatch in line start


oliverlloyd (commented):
Code looks great. Not sure about the api though. If we were being consistent you'd have:
```
effect = `feature` | `sport` | `news' | `labs`
```
But it doesn't work because there are _other_ themes...

Another issue is we don't actually use 'labs' the string anymore, it's `Special.Labs` which is an enum so, sure, we could transform that in the code but it's another level of weird mapping to get your head around.

I appreciate though there's not a great way to name this style other than labs.

Maybe there's another way to approach this. `Lines` is not a declarative component, it's imperative, so perhaps we can have `count={1}` and then combine that with `effect='dashed'`. And then the parent component can stack them?

```
<Lines count={1} effect='straight' />
<Lines count={1} effect='dashed' />
```

?

mxdvl:
Good suggestion, @oliverlloyd. The consumer on Labs would then have to handle the spacing between them.

oliverlloyd:
> Good suggestion, @oliverlloyd. The consumer on Labs would then have to handle the spacing between them.

I can imagine we'd create a parent component that acccepts a `LineType[]` prop which might look like

```
lines={
  [{
    count: 1,
    effect: 'straight',
  },
  {
    count: 1,
    effect: 'dotted',
  }]
}
```

and then that parent would manage spacing. That way we have a reusable abstracton for all places we use lines on the site.

Maybe this abstract could also live in Source? Or maybe that api (eg. `LineType[]`) could be what `Lines` exposes?

oliverlloyd (approved):

oliverlloyd:
I approved the code but deny'd one of the stories (novel approach, I know it was born of doing things in the wrong order!)

Looks like the dashed line story isn't really dashed?

SiAdcock (approved):
This is a thing of beauty, thanks @mxdvl and @oliverlloyd! 💎 

I'm inclined to say let's allow single lines for all effects, even if this is not actually needed right now. Max, would you be able to squash and merge? 🙏

SiAdcock:
@oliverlloyd 

> I can imagine we'd create a parent component that accepts a `LineType[]` prop

I really like this idea! Would you like to have a go at getting it into our Editorial components in this repo?

SiAdcock:
@mxdvl Apologies, one small request before you merge, if possible... Can you update the [README](https://github.com/guardian/source/blob/main/src/editorial/web/components/lines/README.md) for this component? Thanks so much! 🙏

src/editorial/web/components/lines/dashed.ts
```diff
@@ -0,0 +1,28 @@
+import { space } from '@guardian/src-foundations';
+import { line } from '@guardian/src-foundations/palette';
+import { LineCount } from '.';
+
+const thickness = 1;
+const gapHorizontal = 1;
+const viewHeight = space[1];
+const width = thickness + gapHorizontal;
+// const viewHeight = thickness + gapVertical;
```
SiAdcock, line 9: Can we delete this line?

mxdvl: We shall!
```suggestion
```

PR #786, opened 2021-03-31
Coloured Lines

oliverlloyd:
Adds the `colour` prop to the `Lines` component so that we can change the colour of lines from the default grey `line.primary` to something more interesting

## Why?
Because sometimes, like for labs or special reports, that `line.primary` grey gets washed out against the light grey of the article background.

Also, there are some new designs coming out that have even wilder background colours that we want `Lines` to sit on top of and for that we need to set a custom colour to get that nice line pop.

Screenshot 2021-03-31 at 09 15 24

### DIFF
--- a/src/editorial/web/components/lines/styles.ts
+++ b/src/editorial/web/components/lines/styles.ts
@@ -7,12 +7,12 @@ import { LineCount } from '.';
 
 const lineGap = remSpace[1];
 
-export const straightLines = (count: LineCount, colour: string) => {
+export const straightLines = (count: LineCount, color: string) => {
 	const baseStyles = css`
 		background-image: repeating-linear-gradient(
 			to bottom,
-			${colour},
-			${colour} 1px,
+			${color},
+			${color} 1px,
 			transparent 1px,
 			transparent ${lineGap}
 		);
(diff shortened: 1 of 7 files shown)


oliverlloyd:
I made a few refactors ahead of implementing `colour` so if you want to avoid the noise it might be easier to review commit by commit.

oliverlloyd:
[Coloured Lines](https://trello.com/c/GrfqYozT/2770-coloured-lines)

SiAdcock:
Thanks @oliverlloyd this is great 🙏  I've a couple of questions:

- I've been standardising on the US spelling of `color` for variable names, because that's what CSS uses. It's annoying having to switch between `colour` and `color` depending on whether you're using application code or browser internals. Waddyathink?
- Do you think we should limit which colours are available? Currently you can give these lines any hex value under the sun, so it's not providing any accessibility guarantees or design constraints. Do you have any opinions on this @akemitakagi @HarryFischer?

SiAdcock:
Briefly caught up with Akemi, she's fine with this as long as developers pinky promise to follow the design 🤞

oliverlloyd:
> * I've been standardising on the US spelling of `color` for variable names, because that's what CSS uses. It's annoying having to switch between `colour` and `color` depending on whether you're using application code or browser internals. Waddyathink?

👍 

> 
> * Do you think we should limit which colours are available? Currently you can give these lines any hex value under the sun, so it's not providing any accessibility guarantees or design constraints. Do you have any opinions on this @akemitakagi @HarryFischer?

I dicussed this briefly with @HarryFischer and we both thought it would be better to keep the ability to send any colour. The reason for this is because you can set any background colour you like under the sun (this is done by the consumer) so we need to be able to set the correct line colour in relation to this to maintain accessibility. If we constrain the allowed colours then I suspect we will be blocked behind a chain of PRs for each case where the background colour changes 😐 

That or we start to offer pairs or groups of colours that compliment each other? But that seems like a larger discussion.

SiAdcock (approved):
Looks great 👍 💯  Please could you squashy-merge?

src/editorial/web/components/lines/styles.ts
```diff
@@ -2,53 +2,60 @@ import { css } from '@emotion/react';
...
+				${baseStyles};
+				background-size: 1px calc(${lineGap} * 3 + 1px);
+				height: calc(${lineGap} * 3 + 1px);
+			`;
+		case 8:
+			return css`
+				${baseStyles};
+				background-size: 1px calc(${lineGap} * 7 + 1px);
+				height: calc(${lineGap} * 7 + 1px);
+			`;
```
mxdvl, line 41: This could be simplified to handle any number other than 1.

```ts
		default:
			return css`
				${baseStyles};
				background-size: 1px calc(${lineGap} * ${count-1} + 1px);
				height: calc(${lineGap} * ${count-1} + 1px);
			`;
```

oliverlloyd: Oh nice! I would love to see that improvement to the api.

Previously, we only thought of Lines as coming in batches of 4 or 8 but in truth it would be a lot easier to have a more open api - we actually use single 'lines' in lots of places (the article meta is a good example). We get around this by using borders but it'd be great to move away from that.

mxdvl: https://github.com/guardian/source/pull/790 ?

PR #800, opened 2021-04-14
`darkModeCss` Helper

JamieB-gu:

It introduces the `darkModeCss` helper, currently found in [`image-rendering`](https://github.com/guardian/image-rendering/blob/8e15e909cae8654a6e111abcb88dcc3986a54b81/src/lib.ts), to `src-helpers`. It's a variant of the standard Emotion `css` function, and can be used to wrap template string styles in a dark mode media query. The signature is as follows:

```ts
(supportsDarkMode: boolean) => (styles: TemplateStringsArray, ...placeholders: string[]) => SerializedStyles
```

The first parameter allows the caller to choose whether they want these dark mode styles included - if set to `false` they won't be. The rest of the parameters turned this into a [template literal "tag"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates), which allows it to operate on template strings in the same way as the `css` function.

### Example

```tsx
import { Props, darkModeCss } from '@guardian/src-helpers';

const styles = (supportsDarkMode: boolean): SerializedStyles => css`
  background-color: white;
  color: black;

  ${darkModeCss(supportsDarkMode)`
    background-color: black;
    color: white;
  `}
`;

interface HelloProps extends Props {
  name: string;
}

const HelloProps = ({ name, supportsDarkMode = false }) => (
  <p css={styles(supportsDarkMode)}>Hello {name}</p>
);
```

## What does this change?

- Added `darkModeCss` function
- Added `@emotion/react` as peer dep
- Added `supportsDarkMode` as optional Source-wide prop
- Added docs to `src-helpers` README

### Documentation

-   [ ] Examples in Storybook (not sure if this is needed since this isn't a component?)

May need to include this documentation in the main docs site?

### DIFF
--- a/src/core/helpers/README.md
+++ b/src/core/helpers/README.md
@@ -1,3 +1,37 @@
 # Helpers
 
 Shared helper functions and types used across all components.
+
+## Dark Mode
+
+This package exports a `darkModeCss` function, which is a variant of the standard Emotion `css` function. It can be used to wrap template string styles in a dark mode media query. The signature is as follows:
+
+```ts
+(supportsDarkMode: boolean) => (styles: TemplateStringsArray, ...placeholders: string[]) => SerializedStyles
+```
+
+The first parameter allows the caller to choose whether they want these dark mode styles included - if set to `false` they won't be. The rest of the parameters turned this into a [template literal "tag"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates), which allows it to operate on template strings in the same way as the `css` function.
+
+### Example
+
+```tsx
+import { Props, darkModeCss } from '@guardian/src-helpers';
+
+const styles = (supportsDarkMode: boolean): SerializedStyles => css`
+  background-color: white;
+  color: black;
+
+  ${darkModeCss(supportsDarkMode)`
+    background-color: black;
+    color: white;
+  `}
+`;
+
+interface HelloProps extends Props {
+  name: string;
+}
+
+const HelloProps = ({ name, supportsDarkMode = false }) => (
+  <p css={styles(supportsDarkMode)}>Hello {name}</p>
+);
+```
--- a/src/core/helpers/darkMode.ts
+++ b/src/core/helpers/darkMode.ts
@@ -1,0 +1,26 @@
+// ----- Imports ----- //
+
+import type { SerializedStyles } from '@emotion/react';
+import { css } from '@emotion/react';
+
+// ----- Functions ----- //
+
+const darkModeCss = (supportsDarkMode: boolean) => (
+  styles: TemplateStringsArray,
+  ...placeholders: string[]
+): SerializedStyles =>
+  supportsDarkMode
+    ? css`
+        @media (prefers-color-scheme: dark) {
+          ${styles
+            .map(
+              (style, i) => `${style}${placeholders[i] ? placeholders[i] : ""}`
+            )
+            .join("")}
+        }
+      `
+    : css``;
+
+// ----- Exports ----- //
+
+export { darkModeCss };
--- a/src/core/helpers/index.ts
+++ b/src/core/helpers/index.ts
@@ -1,2 +1,3 @@
 export { storybookBackgrounds } from './storybook-bg';
 export * from './types';
+export { darkModeCss } from './darkMode';
--- a/src/core/helpers/package.json
+++ b/src/core/helpers/package.json
@@ -35,5 +35,8 @@
   "devDependencies": {
     "npm-run-all": "^4.1.5",
     "typescript": "^4.1.3"
+  },
+  "peerDependencies": {
+    "@emotion/react": "^11.1.2"
   }
 }
--- a/src/core/helpers/types.ts
+++ b/src/core/helpers/types.ts
@@ -5,4 +5,5 @@ export type ThemeName = 'default' | 'brand' | 'brandAlt';
 export interface Props {
 	className?: string;
 	cssOverrides?: SerializedStyles | SerializedStyles[];
+	supportsDarkMode?: boolean;
 }


SiAdcock:
Thanks for proposing this @JamieB-gu, it's a really tidy and ergonomic API. 

My only concern is the use of dynamic styles which locks us in to having a CSS-in-JS runtime. If we wanted to move away from Emotion in the future, and use a runtime-less solution like [Linaria](https://linaria.dev/) or [Treat](https://seek-oss.github.io/treat/), we would first have to refactor these styles, both here and in all codebases that adopt it.

An approach for supporting dark mode without dynamic styles would be:

```tsx
import { darkMode } = '@guardian/src-foundations/utils'; // "@media (prefers-color-scheme: dark)"

const styles = css`
  background-color: white;
  color: black;
`;

const darkModeStyles = css`
  ${darkMode} {
    background-color: black;
    color: white;
  }
`;

const HelloProps = ({ name, supportsDarkMode = false }) => (
  <p css={[styles, supportsDarkMode ? darkModeStyles : '']}>Hello {name}</p>
);
```

The main drawbacks to this approach over the dynamic version are:

- physical separation of the dark mode styles from the light mode styles
- repetition of the `supportsDarkMode` check in the `css` prop

I'd be interested to hear what you think and anyone else, especially those who have not started supporting dark mode yet, and will need to consider an approach at some point in the future.

gtrufitt:
I will say that I may have my view coloured by this article https://epicreact.dev/css-variables/ but @oliverlloyd did raise this and I think I'm in agreement that the most future-proof approach would be to look at a css variables solution to dark mode?

SiAdcock:
@gtrufitt some compelling thoughts in there for theming, thanks for sharing! The examples there don't seem to leverage the `prefers-color-scheme` media query though 🤔

gtrufitt:
> @gtrufitt some compelling thoughts in there for theming, thanks for sharing! The examples there don't seem to leverage the `prefers-color-scheme` media query though 🤔

If I understand, you could do that like:

```css
:root {
  --main-bg-color: coral; 
}

 @media (prefers-color-scheme: dark){
 	:root {
  		--main-bg-color: red; 
	}
}
```

but there's probably some thought around having global variables like this. I've not thought too much about implementation, I think it'd need to percolate, but I would worry implementing something that ties us to dynamic styles as you mention and would rather leverage the browser platform if feasible.

SiAdcock:
Oooh I really like that! Design tokens could help with the naming conventions and prevent clashes. It's handy that [CSS custom properties](https://caniuse.com/?search=css%20custom%20properties) are supported by the same browsers that support [`prefers-color-scheme`](https://caniuse.com/?search=prefers-color-scheme). The question becomes should components provide a mechanism to fallback to "light mode" in browsers without CSS custom properties? I feel that they should?

```css
:root {
  --main-bg-color: coral; 
}

 @media (prefers-color-scheme: dark){
 	:root {
  		--main-bg-color: red; 
	}
}

.button {
  color: coral;
  color: --main-bg-color;
}
```

JamieB-gu:
Happy to go with whatever solution works if we can find a nice API 😄. I have a few questions - I don't think I've fully understood some of the things discussed 😬.

> don't seem to leverage the `prefers-color-scheme` media query though

I think we'd definitely want to use this media query if possible.

> My only concern is the use of dynamic styles

> An approach for supporting dark mode without dynamic styles would be:

I think I may be missing something, I'm not sure what is meant by "dynamic styles" here. I've had a read through the link @gtrufitt sent and my interpretation is that toggling dark mode in JS is problematic for performance because of all the re-renders? But AFAIK the `darkModeCss` function here would _not_ toggle dark mode in JS - the `supportsDarkMode` prop is set once by the platform and never changed. The dark mode toggling is handled in pure CSS via the media query. From what I can tell, the `darkModeCss` function achieves the same thing as the CSS variables approach, but without the need for those variables?

Again, I think I may have misunderstood something - let me know what you think!

gtrufitt:
Yeah, that's a fair with this implementation when talking specifically about prefers colour scheme. But I have a couple of points I'd be interested in your opinion on.

- If you opt in you're shipping double the CSS to the user, for all users? One of the core tenants of DCR was to not ship CSS to users that they don't use, so that's not ideal but obviously only applies if we 'opt-in' to dark mode.
- On re-rendering, this applies to a switcher, which I think would need an implementation for web because we're not going to be able to rely on 'prefers colour scheme' alone, IMO. So if you do this client-side with a prop, with dynamic-styles you require a full re-render, but not with css vars.

i.e, think Github, they have an opt-in:

image

But I guess we could de-scope this to only listen to prefers-colour-scheme on all platforms...

- Using prefers-colour-scheme, I don't think we can implement a server-side solution (which would solve the 'not shipping unused css' and 'full re-render on switcher change' problems), so we're tied into a full-re-render if that changes, maybe fine.
- CSS Vars seem to fit with our 'Editorial Colour scheme' approach at little more closely? https://github.com/guardian/dotcom-rendering/blob/main/src/web/lib/decidePalette.ts#L722-L805
- CSS Vars do feel more future proof

SiAdcock:
> I'm not sure what is meant by "dynamic styles" here

[Dynamic styles](https://emotion.sh/docs/styled#composing-dynamic-styles) are styles that are expressed as functions and therefore can only be generated at runtime. They cannot reliably be determined in advance or extracted at build time. 

In the following examples, I want you to imagine a hypothetical build tool that could extract CSS passed to the `css` tag and output a CSS file. 

**Static styles**

This style is static and therefore extractable. By leveraging this hypothetical build tool, I could ship the React component and the CSS that gets generated at build time, without needing to ship Emotion's runtime:

```tsx
const staticStyle = css`
    background-color: white;
`
```

**Dynamic styles**

This CSS is generated dynamically. The full CSS can only be generated at runtime. It therefore always requires us to ship a CSS-in-JS runtime library, assuming the component is rendered on the client:

```tsx
const dynamicStyles = (color: string) => css`
    background-color: ${color};
`
```

Our [contributing guidelines](https://theguardian.design/2a1e5182b/p/77c9d9-contributing/t/902350) recommend avoiding dynamic styles, where possible.

**Why is this important?**

But... Simon. Emotion always requires us to ship a runtime anyway. So why can't I use dynamic styles, if that's what Emotion supports?

The answer is to reduce the burden of a future migration.

I'm betting that in the future we will want to stop supporting dynamic styles, or support them in only limited contexts. Why?

1. Shipping a CSS-in-JS runtime to clients is needlessly expensive. The library itself is around 5KB and the styles expressed as JavaScript are redundant if we are shipping styles as CSS anyway. Do we _need_ to ship styles as both CSS and JS? I don't think so.
2. Injecting CSS at runtime requires us to [add `unsafe-inline` to the `style-src` directive](https://docs.google.com/document/d/1Lhy06ea9hg3h-tPPgvy3q5fIlLTN-i_3kT78aouXpEo), which opens us up to all sorts of [interesting attack vectors](https://www.mike-gualtieri.com/posts/stealing-data-with-css-attack-and-defense).

It's possible that in future we could use a runtimeless CSS-in-JS library such as [Treat](https://seek-oss.github.io/treat/) or [Linaria](https://linaria.dev/). Such as change will be a significant undertaking. It's important now for us to think of the burden we're creating for ourselves in the future.

JamieB-gu:
I will do my best to answer your questions @gtrufitt 💪

> * If you opt in you're shipping double the CSS to the user, for all users?

Yes, but I think that's inevitable if you want to support users switching between dark and light mode? The only alternative I can see is some combination of:

- Detecting the user's preference on page load and either dynamically pulling in the styles, or splitting the cache
- Listening for change to the user's preference in JavaScript (`window.matchMedia('(prefers-color-scheme: dark)').addEventListener(...)`) and dynamically pulling in styles, which will have a performance hit (also a problem for Apps/Editions if they're offline)

I'd argue that with the CSS variable approach you're still shipping all the variables for both modes to every user too, unless you take one of the above approaches.

> * On re-rendering, this applies to a switcher, which I think would need an implementation for web because we're not going to be able to rely on 'prefers colour scheme' alone, IMO

Why wouldn't you be able to rely on `prefers-color-scheme` out of curiosity?

> * Using prefers-colour-scheme, I don't think we can implement a server-side solution

How come? The AR implementation is fully server-side atm. Is this a consequence of hydration client-side?

---

On to @SiAdcock's points:

> I want you to imagine a hypothetical build tool that could extract CSS passed to the `css` tag and output a CSS file.

Like Emotion when used with `extractCritical`? Example code [here](https://github.com/guardian/apps-rendering/blob/9f10e270a88525074104dc82247d4051c3f98b1b/src/server/page.tsx#L131) in AR and an [example of an AR page](https://mobile.guardianapis.com/rendered-items/environment/2021/apr/16/marks-spencer-honeybee-project-threat-biodiversity-conservationists-aoe) with all styles extracted to a tag in the `head` (second `style` tag, the first is fonts) and hashed in the CSP.

I'm not sure whether this is an example of what you mean 🤔.

> assuming the component is rendered on the client

Yeah this is an interesting point. In AR's case this is mostly _not_ true; there's not as yet been a need for hydration, so most things are rendered server-side.

> I'm betting that in the future we will want to stop supporting dynamic styles, or support them in only limited contexts.

Yeah this seems sensible to me. I've done some experiments with this in https://github.com/guardian/apps-rendering/pull/1236 but it's not an ideal solution. I think I need to understand the subtleties better; for example, I don't understand why your example for "some styles expressed as functions can be generated at build time" doesn't fall foul of this issue.

> This style is static and therefore extractable.

Assuming the dynamic parameter that's at fault here is `supportsDarkMode`, that property is known at build time because it's platform-specific. If we could figure out a way of baking that into the build, then the `darkModeCss` function becomes static again. For example:

```ts
const dcrDarkModeCss = darkModeCss(false);

const arDarkModeCss = darkModeCss(true);
```

SiAdcock:
>  that property is known at build time because it's platform-specific. If we could figure out a way of baking that into the build, then the darkModeCss function becomes static again.

I like that framing, I hadn't thought about that. However there's no way Source can guarantee it. It's just a convention that we trust platforms adopt. In the [`Stack` example](https://github.com/guardian/source/blob/7e2c3d431f2da0103b9eff43eff3f8f5a8dc460d/src/core/components/layout/components/stack/styles.ts#L11-L37), we see functions that could be pre-executed at _Source's build time_ and optimised away to static styles.

Please let me know if this makes no sense 😅  I'd also be interested to hear the opinions of others. @sndrs @mxdvl @faresite @i-hardy

JamieB-gu:
> do we still want this?

Good question. I guess that depends on whether you think it would be useful outside of DCR/AR? The migration of image-rendering to the mono-repo means that dotcom, apps and editions all now have access to this directly.

PR #801, opened 2021-04-15
`src ed button`

oliverlloyd:
## What?
Adds a new `Button` component to the `editorial` design system.

This editorial `Button` component extends the core `Button` component adding a single additional prop along the way. This new prop is `format`, taken from https://github.com/guardian/types/

`format` is then used to override the css props `color`, `background-color`, `border` and, sometimes, `:hover` based on `priority`

--------

### `primary`
Screenshot 2021-04-15 at 12 08 45

### `secondary`
Screenshot 2021-04-15 at 12 08 45

### `tertiary`
Screenshot 2021-04-15 at 12 09 08

### `subdued`
Screenshot 2021-04-15 at 12 09 31

## Why?
This component [already exists in Discussion](https://github.com/guardian/discussion-rendering/blob/fc14c26db73bfec8a04ff7a503ed9f90f1a1a8ad/src/components/PillarButton/PillarButton.tsx#L86) and is generic enough to warrent a position higher up, where it can be imported into multiple codebases

## What about `secondary`?
Right now, this is returning the same values as for `primary`. When `PillarButton` was first built in discussion I don't think `secondary` existed as an option so it wasn't styled for and I can't think of an obvious way to go. @HarryFischer Akemi, any thoughts?

## `ButtonProps` are now being exported
This PR exports `ButtonProps` from the `core` `Button` component so that we can extend from it in the `editorial` version, allowing us to pass in those props in addition to `format`

### DIFF
--- a/src/editorial/web/components/button/index.tsx
+++ b/src/editorial/web/components/button/index.tsx
@@ -1,0 +1,237 @@
+import { css } from '@emotion/react';
+
+import { Design, Display, Format, Pillar, Special } from '@guardian/types';
+import {
+	Button as CoreButton,
+	ButtonProps,
+	ButtonPriority,
+} from '@guardian/src-button';
+import {
+	culture,
+	labs,
+	lifestyle,
+	news,
+	opinion,
+	specialReport,
+	sport,
+	neutral,
+} from '@guardian/src-foundations';
+
+export interface Props extends ButtonProps {
+	format?: Format;
+	priority?: ButtonPriority;
+	children: React.ReactNode;
+}
+
+const defaultFormat = {
+	display: Display.Standard,
+	design: Design.Article,
+	theme: Pillar.News,
+};
+
+const WHITE = neutral[100];
+
+const decideBackground = (format: Format, priority: ButtonPriority) => {
+	switch (priority) {
+		case 'primary':
+		case 'secondary':
+			switch (format.theme) {
+				case Pillar.News:
+					return css`
+						background-color: ${news[300]};
+						:hover {
+							background-color: ${news[400]};
+							border: 1px solid ${news[400]};
+						}
+					`;
+				case Pillar.Culture:
+					return css`
+						background-color: ${culture[300]};
+						:hover {
+							background-color: ${culture[400]};
+							border: 1px solid ${culture[400]};
+						}
+					`;
+				case Pillar.Lifestyle:
+					return css`
+						background-color: ${lifestyle[300]};
+						:hover {
+							background-color: ${lifestyle[400]};
+							border: 1px solid ${lifestyle[400]};
+						}
+					`;
+				case Pillar.Sport:
+					return css`
+						background-color: ${sport[300]};
+						:hover {
+							background-color: ${sport[400]};
+							border: 1px solid ${sport[400]};
+						}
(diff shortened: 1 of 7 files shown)


liywjl (approved):
Cool stuff 👍

oliverlloyd:
[Theme Button](https://trello.com/c/jy6Ux3ff/2815-theme-button)

src/editorial/web/components/button/index.tsx
```diff
@@ -0,0 +1,266 @@
...
+	opinion,
+	specialReport,
+	sport,
+	neutral,
+} from '@guardian/src-foundations';
+
+export interface ButtonProps extends CoreButtonProps {
+	format?: Format;
+	priority?: ButtonPriority;
+	children: React.ReactNode;
```
SiAdcock, line 25: Do we need to specify `priority` and `children` if they're already in `CoreButtonProps`?

src/editorial/web/components/button/index.tsx
```diff
@@ -0,0 +1,266 @@
...
+const WHITE = neutral[100];
+
+const decideBackground = (format: Format, priority: ButtonPriority) => {
+	switch (priority) {
+		case 'primary':
+		case 'secondary':
+			switch (format.theme) {
+				case Pillar.News:
+					return css`
+						background-color: ${news[300]};
```
SiAdcock, line 49: I would like to see these colours expressed in a more context-aware way (e.g. `background.ctaPrimaryNews` as demonstrated in the [Tokens documentation](https://theguardian.design/2a1e5182b/p/1377a6-tokens/b/293ddb)). I know we are still talking about how the Editorial palette will be structured, so perhaps we can come back to this when the structure is clearer.

src/editorial/web/components/button/index.tsx
```diff
@@ -0,0 +1,266 @@
...
+						border: 1px solid ${opinion[400]};
+					`;
+				case Special.Labs:
+					return css`
+						border: 1px solid ${labs[400]};
+					`;
+				case Special.SpecialReport:
+					return css`
+						border: 1px solid ${specialReport[400]};
+					`;
```
SiAdcock, line 175: We can probably use `border: 1px solid currentColor` for all the `tertiary` buttons

src/editorial/web/components/button/index.tsx
oliverlloyd: In DCR, we use the `news`, `opinion`, etc. objects everywhere. We vertually never use any of the tokens. This is for two reasons

1. The tokens often don't work for our use cases, they seem very banner/callout orientated, but also mainly because
2. The design team specifically gives use `news[300]` etc., as the required colours to use 🤷

src/editorial/web/components/button/index.tsx
oliverlloyd: Oh nice idea!

src/editorial/web/components/button/index.tsx
oliverlloyd: Good spot, shalll remove

SiAdcock (approved):
This looks great, thanks @oliverlloyd 💎 Please can you squashymerge?

PR #804, opened 2021-04-16
`src-ed-quote-icon`

oliverlloyd:
## What?
Adds the `QuoteIcon` component as part of the `src-ed-quote-icon` package.

Screenshot 2021-04-16 at 15 22 43

## Why?
The Guardian quote icon is used in many places and is a strong part of our brand identity. We should have one cannonical definition for it.

## You vomit?
It's [about aerial yoga](https://www.theguardian.com/lifeandstyle/2019/nov/12/aerial-yoga-i-look-like-a-buffoon-i-feel-incredible-and-then-i-vomit) 🤷

### DIFF
--- a/src/editorial/web/components/quote-icon/index.tsx
+++ b/src/editorial/web/components/quote-icon/index.tsx
@@ -1,0 +1,136 @@
+import { css } from '@emotion/react';
+
+import { Format, Pillar, Special } from '@guardian/types';
+import { until } from '@guardian/src-foundations/mq';
+import {
+	culture,
+	labs,
+	lifestyle,
+	news,
+	opinion,
+	specialReport,
+	sport,
+} from '@guardian/src-foundations';
+
+const quoteStyles = css`
+	height: 17px;
+	width: 9px;
+	margin-right: 12px;
+	transform: translateY(-1px);
+	overflow: visible;
+`;
+
+const quoteColor = (format: Format) => {
+	switch (format.theme) {
+		case Pillar.News: {
+			return css`
+				fill: ${news[400]};
+			`;
+		}
+		case Pillar.Opinion: {
+			return css`
+				fill: ${opinion[400]};
+			`;
+		}
+		case Pillar.Culture: {
+			return css`
+				fill: ${culture[400]};
+			`;
+		}
+		case Pillar.Sport: {
+			return css`
+				fill: ${sport[400]};
+			`;
+		}
+		case Pillar.Lifestyle: {
+			return css`
+				fill: ${lifestyle[400]};
+			`;
+		}
+		case Special.SpecialReport: {
+			return css`
+				fill: ${specialReport[400]};
+			`;
+		}
+		case Special.Labs: {
+			return css`
+				fill: ${labs[400]};
+			`;
+		}
+	}
+};
+
+type HeadlineSize = 'tiny' | 'small' | 'medium' | 'large';
+
+const sizeStyles = (size: HeadlineSize) => {
+	const tinySvg = css`
+		svg {
+			height: 13px;
+			width: 7px;
+		}
+	`;
+	const smallSvg = css`
+		svg {
+			height: 16px;
+			width: 8px;
+		}
+	`;
+	const mediumSvg = css`
+		margin-right: 4px;
+		svg {
+			height: 20px;
+			width: 11px;
+		}
+	`;
+	const largeSvg = css`
+		margin-right: 8px;
+		svg {
+			height: 24px;
+			width: 13px;
+		}
+	`;
+
+	switch (size) {
+		case 'tiny':
+			return css`
+				${tinySvg}
+			`;
+		case 'small':
+			return css`
+				${smallSvg}
+			`;
+		case 'medium':
+			return css`
+				${mediumSvg}
+				${until.desktop} {
+					${smallSvg}
+				}
+			`;
+		case 'large':
+			return css`
+				${largeSvg}
+				${until.desktop} {
+					${mediumSvg}
+				}
+			`;
+	}
+};
+
+type Props = {
+	format: Format;
+	size: HeadlineSize;
+};
+
+export const QuoteIcon = ({ format, size }: Props) => (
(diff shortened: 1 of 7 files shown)


src/editorial/web/components/quote-icon/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+			width="70"
+			height="49"
+			viewBox="0 0 35 25"
+			css={[quoteStyles, quoteColor(format)]}
+			data-testid="quote-icon"
+		>
+			<path d="M69.587.9c-1.842 15.556-3.89 31.316-4.708 48.1H37.043c3.07-16.784 8.391-32.544 17.602-48.1h14.942zM32.949.9c-2.047 15.556-4.094 31.316-4.912 48.1H.2C3.066 32.216 8.592 16.456 17.598.9h15.35z" />
+		</svg>
+	</span>
+);
```
liywjl, line 136: I feel like this could maybe be moved to Icons folder https://github.com/guardian/source/blob/main/src/core/icons

src/editorial/web/package.json
```diff
@@ -25,6 +25,9 @@
     "verbump:preminor": "yarn version --preminor --preid rc --no-git-tag-version",
     "verbump:prerelease": "yarn version --prerelease --preid rc --no-git-tag-version"
   },
+  "dependencies": {
+    "@guardian/types": "^6.0.0"
+  },
```
liywjl, line 30: do we want to bundle types? or should we use `peerDependencies` instead?

oliverlloyd: Good question. Expecting platforms to use types seems very fair. @SiAdcock , what think you?

src/editorial/web/components/quote-icon/index.tsx
oliverlloyd: Yeah, I agree. @mxdvl apparently when anyone talks about svgs we ping you. Shout if that's starting to get old.

How can I best bring the svg in from `/icons` but still set the fill color?

oliverlloyd:
[QuoteIcon in Source](https://trello.com/c/SFzlCUNs/2822-quoteicon-in-source)

src/editorial/web/components/quote-icon/index.tsx
mxdvl: SVGs never get old.

Either you export the icon as a react component, or you just don't specify any fill colour and use css styles on the parent div to style it, like so:

```ts
const style = css`
  svg {
    fill: ${brandColor}
  }
```

oliverlloyd: Oh, so actually these two approaches are how I've been styling svgs for years. I just never realised they were established patterns, great to know!

mxdvl: I find it liberating to treat inline SVGs as just another HTML element.

CSS applies according to [specificity rules][], so if a `fill` attribute is specified on the element, it will take precedence over a rule defined further up. Worth remembering, too, that if you have `styles` defined in a standalone SVG image that becomes inlined, they will apply to the surrounding document.

[specificity rules]: https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

mxdvl: On a slightly different note, you may want to:

- reduce the precision of that `path` down to 1 or 2 decimals, as it seems a bit too precise.
- Consider why the `width`/`height` ratio is different from the `viewBox`?

src/editorial/web/package.json
SiAdcock: Hmm. Would it not be a `devDependency`? Types are a compile-time concern rather than a runtime one.

SiAdcock (commented):
I think this is almost ready to go, thanks for your hard work @oliverlloyd . I'd just like to see it use the existing quote icon from Source, if possible. Let me know if you have any questions.

src/editorial/web/components/quote-icon/index.tsx
oliverlloyd: In the end I imported the existing `SvgQuote` (I missed that!)

oliverlloyd:
> use the existing quote icon from Source

@SiAdcock Thanks for pointing out that icon already existed, this is now done 👍

src/editorial/web/components/quote-icon/stories.tsx
```diff
@@ -0,0 +1,422 @@
...
+			<QuoteIcon
+				size="tiny"
+				format={{
+					display: Display.Standard,
+					design: Design.Article,
+					theme: Pillar.News,
+				}}
+			/>
+			<span>
+				I look life a buffoon. I feel incredible. And then I vomit
```
mxdvl, line 60: I think this is sound from an SVG perspective.

```suggestion
				I look like a buffoon. I feel incredible. And then I vomit
```

mxdvl (approved):

src/editorial/web/components/quote-icon/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+			return css`
+				svg {
+					fill: ${labs[400]};
+				}
+			`;
+		}
+	}
+};
+
+type HeadlineSize = 'tiny' | 'small' | 'medium' | 'large';
```
SiAdcock, line 69: A point on conventions. Source uses `xsmall` as the next size down from `small` in places like [typography](https://theguardian.design/2a1e5182b/p/95d5d0-code/t/858844) and the [Button component](https://theguardian.design/2a1e5182b/p/435225-button/b/86f344/t/59cc6b). I wonder if in the name of consistency and predictability, we should continue this convention here?

oliverlloyd: `teensyweensy`? 🙏

oliverlloyd: Just in case there's no agreement on teensyweensy I pushed a change to `xsmall`

SiAdcock (approved):
Thanks @oliverlloyd 🙏

PR #805, opened 2021-04-19
Editorial palette

oliverlloyd:
## Editorial Palette
This PR adds an implementation of an editorial palette. An editorial palette is a concept that has long been discussed and is something used in isolation on at least two projects. By creating a versioin of it here we aim to centralise this logic so that it can be shared.

### `decidePalette`
A function that takes one argument `format` and returns an object `Palette` which contains keys to the colours that should be used for a piece of content

## Why now?
Because `palette` is a dependency for a lot of components that we would like to hoist into Source but we need the `Palette` type to be shared before we can do that. I considered just putting `Palette` into `@guardian/types` but this felt wrong so I went with taking the bolder step of actually implementing an editorial palette.

This function is without doubt going to need to change in order to meet the requirements for other projects but by having it exist we start the process of creating one that will.

### DIFF
--- a/src/editorial/web/index.tsx
+++ b/src/editorial/web/index.tsx
@@ -1,1 +1,4 @@
 export { Lines } from './components/lines';
+
+export type { Palette } from './libs/decidePalette';
+export { decidePalette } from './libs/decidePalette';
--- a/src/editorial/web/libs/decidePalette/index.ts
+++ b/src/editorial/web/libs/decidePalette/index.ts
@@ -1,0 +1,888 @@
+// ----- Imports ----- //
+import {
+	border,
+	brand,
+	brandAlt,
+	brandAltBackground,
+	culture,
+	labs,
+	lifestyle,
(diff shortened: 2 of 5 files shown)


oliverlloyd:
[Editorial Palette](https://trello.com/c/Tz1Mf4se/2830-editorial-palette)

SiAdcock (commented):
This looks fine to my naive eyes. I know you've been using this for a while and I agree that this concept should be hoisted up to the Editorial Design System.

I'd like to get some @guardian/mss-developers eyes on it if possible to make sure we're not coding ourselves into any corners. Do you have opinions on this @JamieB-gu @faresite? You've probably already discussed it elsewhere, I'm guessing

oliverlloyd:
I'm closing this for now as there's not a route open to bring this code down into DCR (Emotion upgrade issues) nor much appetite to use it outside DCR so it makes sense to keep it local for now

PR #820, opened 2021-05-18
`Container` as a `Section`

oliverlloyd:
This PR enhances the `Container` layout component to:

- Allow the background colour to be specified (full width)
- Adds a `section` tag
- Allows side and top borders to be set separately
- Allows the colour of the borders to be set

### Before
```ts
interface Container extends HTMLAttributes<HTMLDivElement>, Props {
	border?: boolean;
	cssOverrides?: SerializedStyles | SerializedStyles[];
	children: ReactNode;
}
```

### After
```ts
interface Container extends HTMLAttributes<HTMLDivElement>, Props {
	sideBorders?: boolean;
	topBorder?: boolean;
	backgroundColor?: string;
	borderColor?: string;
	cssOverrides?: SerializedStyles | SerializedStyles[];
	children: ReactNode;
}
```

### DIFF
--- a/src/core/components/layout/components/container/container.tsx
+++ b/src/core/components/layout/components/container/container.tsx
@@ -1,27 +1,49 @@
 import { HTMLAttributes, ReactNode } from 'react';
 import { SerializedStyles } from '@emotion/react';
-import { container, containerBorder } from './styles';
+import {
+	container,
+	containerSideBorders,
+	containerTopBorder,
+	containerBackground,
+} from './styles';
 import { Props } from '@guardian/src-helpers';
+import { border } from '@guardian/src-foundations/palette';
 
 interface Container extends HTMLAttributes<HTMLDivElement>, Props {
-	border?: boolean;
+	sideBorders?: boolean;
+	topBorder?: boolean;
+	backgroundColor?: string;
+	borderColor?: string;
 	cssOverrides?: SerializedStyles | SerializedStyles[];
 	children: ReactNode;
 }
 
 const Container = ({
-	border = false,
+	sideBorders = false,
+	topBorder = false,
+	backgroundColor,
+	borderColor = border.secondary,
 	cssOverrides,
 	children,
 	...props
 }: Container) => {
 	return (
-		<div
-			css={[container, border ? containerBorder : '', cssOverrides]}
-			{...props}
+		<section
+			css={[backgroundColor && containerBackground(backgroundColor)]}
 		>
-			{children}
-		</div>
+			<div
+				css={[
+					container,
+					backgroundColor && containerBackground(backgroundColor),
+					topBorder && containerTopBorder(borderColor),
+					sideBorders && containerSideBorders(borderColor),
+					cssOverrides,
+				]}
+				{...props}
+			>
+				{children}
+			</div>
+		</section>
 	);
 };
 const defaultProps = {};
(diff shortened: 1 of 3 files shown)


src/core/components/layout/components/container/container.tsx
```diff
@@ -1,27 +1,49 @@
...
+	container,
+	containerSideBorders,
+	containerTopBorder,
+	containerBackground,
+} from './styles';
 import { Props } from '@guardian/src-helpers';
+import { border } from '@guardian/src-foundations/palette';
 
 interface Container extends HTMLAttributes<HTMLDivElement>, Props {
-	border?: boolean;
```
SiAdcock, line 7: I would suggest leaving the `border` prop in place for backward compatibility for now. We can remove it from the documentation to ensure people don't use it in the future.

src/core/components/layout/components/container/container.tsx
```diff
@@ -1,27 +1,49 @@
...
-		</div>
+			<div
+				css={[
+					container,
+					backgroundColor && containerBackground(backgroundColor),
+					topBorder && containerTopBorder(borderColor),
+					sideBorders && containerSideBorders(borderColor),
+					cssOverrides,
+				]}
+				{...props}
```
SiAdcock, line 42: It has been the convention in other components to spread props over the parent element of a component. Is there a reason we shouldn't spread the props over the `section` element?

src/core/components/layout/components/container/styles.ts
```diff
@@ -25,14 +25,18 @@ export const container = css`
...
 
-export const containerBorder = css`
-	border-style: solid;
-	border-color: ${border.secondary};
-	border-width: 0;
+export const containerSideBorders = (color?: string) => css`
+	border-left-style: solid;
+	border-right-style: solid;
+	border-left-color: ${color || border.secondary};
+	border-right-color: ${color || border.secondary};
```
SiAdcock, line 32: I'd like to avoid, or reduce the scope of, dynamic styles as much as possible. Would  you be able to extract the styles that rely on `color` into a separate function and call it only when a `borderColor` is specified in the props?

src/core/components/layout/components/container/container.tsx
```diff
@@ -1,27 +1,49 @@
...
 	cssOverrides?: SerializedStyles | SerializedStyles[];
 	children: ReactNode;
 }
 
 const Container = ({
-	border = false,
+	sideBorders = false,
+	topBorder = false,
+	backgroundColor,
+	borderColor = border.secondary,
```
SiAdcock, line 25: Following on from the previous comment, I'd prefer `border.secondary` to be specified in the styles rather than here. This may be overridden by further styles if `borderColor` is explicitly passed as a prop.

SiAdcock:
> Do I need to update the docs shown on the Source website?

No, there's no way for you to do that unfortunately. I'll do it manually after this is merged.

src/core/components/layout/components/container/container.tsx
oliverlloyd: Happy to follow convention

oliverlloyd:
@SiAdcock Thanks for the great review comments (as ever). I've made those changes and pushed the updates for your PR reviewing pleasure.

SiAdcock (approved):
Lovely thanks @oliverlloyd 💎

PR #834, opened 2021-06-18
Create a workflow triggered by Dependabot PRs

SiAdcock:

As of March 2021, workflows triggered by a `push` event [don't have access to GitHub secrets](https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/), if that event originates from a fork of the repo. As a result, PRs raised by Dependabot are not able to trigger a Chromatic run.

## What does this change?

Here I am creating [a special workflow](https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-852541544) that completes successfully if the user raising the PR is Dependabot.

When this completes, the Chromatic workflow is triggered by the `workflow_run` event. Workflows triggered by this event have access to secrets, so can successfully run Chromatic. 

~I have given the new workflow [`write` permission to the `GITHUB_TOKEN`](scope(https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token)) within the `checks` scope. I believe this is needed to write the result of the checks (in this case Chromatic) to the PR in GitHub. However, I'm not 100% on whether this is necessary.~


.github/workflows/chromatic-dependabot.yml
```diff
@@ -0,0 +1,23 @@
+name: Dependabot Chromatic
+on:
+  workflow_run:
+    workflows: ['Dependabot PR Check']
+    types:
+      - completed
+
+permissions:
+  checks: write
```
jamie-lynch, line 9: This permission is for the `GITHUB_TOKEN`, right? Does chromatic need write permissions or does it just need read to be able to checkout the branch?

SiAdcock: You're probably right. I'll remove this and see how it gets on without it.

jamie-lynch:
As @SiAdcock pointed out offline, the issue with this is that the Chromatic workflow and the special Dependabot Chromatic workflow will both run when Depenabot opens a PR, but the original Chromatic one will fail.

We could add `if: ${{ github.actor != 'dependabot[bot]' }}` to the original workflow so that only one of the two checks runs depending on the PR author but then we can't make both checks required which isn't ideal.

Instead maybe the condition to the original Chromatic workflow could be something like below to catch both cases?

```
on:
  workflow_run:
    workflows: ['Dependabot PR Check']
    types:
      - completed
  push:
    branches-ignore:
      - 'gh-pages'
      - '[0-9]+.[0-9]+'
      - 'dependabot/.*' // Or whatever the correct patter is here
```

SiAdcock:
We are blocked on the Chromatic GitHub action not currently supporting `workflow_run`

Example of a failing run: https://github.com/guardian/source/actions/runs/949527467

```
workflow_run event is not supported in this action
```

See also chromaui/chromatic-cli#314

We should keep an eye on this issue to see if it gets resolved.

Issue #623, opened 2020-11-25
RadioGroup allow `supporting` prop to be able to use `ReactNode`


liywjl:
**Is your feature request related to a problem? Please describe.**

In some quizzes on frontend we have a supporting image next to the question (example: https://www.theguardian.com/football/that-1980s-sports-blog/2020/jun/12/sports-quiz-football-in-the-1980s look at question 4).

The best way to render the supporting image would be as part of the `RadioGroup` component. However there is no way to have text followed by supporting media in the current design.

**Describe the solution you'd like**

We could change `supporting` prop on `RadioGroup` to allow `text | ReactNode` allowing us to have images or videos as supporting content.

**Describe alternatives you've considered**

- We could add another prop (suggestion: `supportingMedia` as a name) to `RadioGroup`
- We could allow `RadioGroup` to allows other components other than Radio (not ideal)

SiAdcock:
Hi @liywjl, thanks for requesting this. We'd be totally up for supporting it. Would you be able to raise a PR?

SiAdcock:
Fixed in #839

PR #839, opened 2021-06-24
623: add JSX.Element type to supporting prop definition

joecowton1:

As per https://github.com/guardian/source/issues/623 this PR addresses the request for a `supportingMedia` prop on the `RadioGroup` component.

This is preferred to adding an additional type to the `supporting` prop ie `string | ReactNode` as `supporting` is used as a text input to other components where it makes no sense to pass a React component. 

**I've tried to stick to Source conventions where I've noticed them, but apologies if I've missed anything! Please pull me up on anything like that.

## What does this change?

-   add `supportingMedia` prop
-   add corresponding stories

## Screenshots

Screenshot 2021-06-24 at 16 57 25

### Accessibility

### Responsiveness

### Documentation

-   [x] Examples in Storybook


SiAdcock:
Thanks so much @joecowton1 

On this:

> This is preferred to adding an additional type to the supporting prop ie string | ReactNode as supporting is used as a text input to other components where it makes no sense to pass a React component.

I think you're referring to the `supporting` prop on the `Legend` component. Do you not think the Legend component's supporting prop should also be a `string | ReactNode`?

joecowton1:
> Thanks so much @joecowton1
> 
> On this:
> 
> > This is preferred to adding an additional type to the supporting prop ie string | ReactNode as supporting is used as a text input to other components where it makes no sense to pass a React component.
> 
> I think you're referring to the `supporting` prop on the `Legend` component. Do you not think the Legend component's supporting prop should also be a `string | ReactNode`?

@SiAdcock I think you're right on this, I've changed the approach just to use the `supporting` prop for both text and media components.

SiAdcock:
Thanks @joecowton1 this is looking great! Can I ask for a couple of things?

- Add a story for the change to the Legend component
- Update the Legend docs to say that it now supports any JSX element
- Split the changes to Legend and the changes to RadioGroup into separate commits. This helps makes the changelog more coherent!

thanks so much for your help so far 🙏

src/core/components/radio/index.tsx
```diff
@@ -127,7 +127,7 @@ interface RadioProps extends InputHTMLAttributes<HTMLInputElement>, Props {
 	checked?: boolean;
 	defaultChecked?: boolean;
 	label?: ReactNode;
-	supporting?: ReactNode;
+	supporting?: ReactNode | JSX.Element;
```
SiAdcock, line 130: Is this intentional? I wonder what is the difference between these types, in a practical sense?

joecowton1: No that's just me not paying attention. Reverted :)

joecowton1:
@SiAdcock I _think_ I've covered all those tasks now, please say if I've missed anything!

SiAdcock (approved):
Thanks so much @joecowton1!

PR #843, opened 2021-07-02
Add clock icon

tjmw:

We'd like to use the clock icon in a component on the website (alongside email newsletter frequency).

## What does this change?

Adds clock icon `SvgClock`.

## Screenshots

Screenshot 2021-07-02 at 17 00 15

### Accessibility

### Responsiveness

### Documentation

-   [ ] Examples in Storybook


src/core/icons/clock.tsx
```diff
@@ -0,0 +1,11 @@
+export const SvgClock = () => {
+	return (
+		<svg viewBox="0 0 30 30" xmlns="http://www.w3.org/2000/svg">
+			<path
+				fillRule="evenodd"
+				clipRule="evenodd"
```
mxdvl, line 6: `clip-rule` [does not apply in this case](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/clip-rule). There’s no `clip-path`.

```suggestion
```

mxdvl (approved):
Looking good from an SVG perspective 📐

src/core/icons/clock.tsx
tjmw: Thanks Max, I've removed 👍 (and squashed into the same commit)

SiAdcock (approved):

PR #844, opened 2021-07-05
Update font-face urls to `assets.guim` domain

jamie-lynch:
## What does this change?

This PR updates the `font-face` tags in `.storybook/preview-head.html` to load the fonts from the `assets.guim` domain as specified in https://github.com/guardian/fonts

## How to test

Run storybook and verify that components still render correctly

## How can we measure success?

Fonts are loaded from the recommended location


SiAdcock (approved):
👏

PR #865, opened 2021-07-16
Update husky version in package.json

SiAdcock:

Somehow the Husky version has become confused in the yarn.lock. Running `yarn` changes the version from 7.0.1 to 7.0.0

## What does this change?

~Commits the change that is produced by running `yarn`~

Edit: Update to latest version of Husky in the `package.json`

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "eslint-plugin-prettier": "^3.3.1",
     "eslint-plugin-react": "^7.23.1",
     "execa": "^5.0.0",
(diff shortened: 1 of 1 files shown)


jamie-lynch:
Sorry, that's probably me getting muddled between commits. `7.0.1` is out so maybe we want to bump the `package.json` to that instead?

jamie-lynch (approved):

SiAdcock:
@jamie-lynch oh yes, good call! Done 👍

PR #875, opened 2021-07-21
Create generate-component-list script

jamie-lynch:

This PR adds a script which generates a package list and component list (per package) from the contents of the repository. This is added to support the work to monitor component usage across the organisation. 

## What does this change?

1) Add the `@typescript-eslint/typescript-estree` package to parse file contents
1) Add a new script which:
    1) Collates the available packages and a directory path for each
    1) Looks for either an `index.tsx` file or a series of `.tsx` files (excluding `stories.tsx`)
    1) Use `@typescript-eslint/typescript-estree` to parse the file contents and extract any non-type export that starts with a capital letter


SiAdcock:
I love this as a concept and as a piece of implementation. I think this script should replace the existing (and less comprehensive [`scripts/package-names.ts`](https://github.com/guardian/source/blob/main/scripts/package-names.ts)

jamie-lynch:
Given the work done in https://github.com/guardian/actions-source-analysis, we won't need this script for doing source analysis.

PR #876, opened 2021-07-21
Export `ButtonProps` as type

jamie-lynch:

This PR updates the export from the `@guardian/src-button` package so that the `*Props` type objects are exported as types. This change has been driven by the work in #875 to generate a list of packages and corresponding components for each.

## What does this change?

-   Export the `ButtonProps` and `LinkButtonProps` types as types (i.e. using `export type`)


SiAdcock (approved):
Oh, good one! Thanks @jamie-lynch

PR #878, opened 2021-07-22
Revert "Create a workflow triggered by Dependabot PRs"

jamie-lynch:
This PR reverts the changes made in guardian/source#834 which configured a special workflow that was run if the PR author was Dependabot. This in turn triggered chromatic via the `workflow_run` trigger. This technique proved unsuccessful as Chromatic does not currently support the `workflow_run` event. 

Support for this event is planned for [Chromatic's Q3 improvements](https://github.com/chromaui/chromatic-cli/milestone/2) at which point we will re-implement this solution. In the meantime, the chromatic check will fail for dependabot PRs and have to be re-run manually.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,14 +1,9 @@
 name: Chromatic
 on:
-  workflow_run:
-    workflows: ['Dependabot PR Check']
-    types:
-      - completed
   push:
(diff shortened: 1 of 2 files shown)


SiAdcock (approved):
Great, thanks @jamie-lynch 🙏

PR #883, opened 2021-07-27
Show packages in storybook nav

sndrs:

- breaks out stories by package in the nav
- provides a `Demo` story for accordion docs
- fills out stories for all possible accordion prop states for chromatic to watch

## Screenshots

**Before**

image

**After**

image

### DIFF
--- a/src/core/components/accordion/index.stories.tsx
+++ b/src/core/components/accordion/index.stories.tsx
@@ -1,7 +1,7 @@
 import { Accordion, AccordionProps, AccordionRow } from './index';
 
 export default {
-	title: 'Accordion',
+	title: 'Source/src-accordion/Accordion',
 	component: Accordion,
 	subcomponents: { AccordionRow },
 };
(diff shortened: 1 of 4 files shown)


src/core/components/accordion/index.stories.tsx
```diff
@@ -1,30 +1,39 @@
...
 
-Default.args = {
+Demo.args = {
 	hideToggleLabel: false,
 };
 
-Default.story = { name: 'default' };
+export const WithCTALabels = (args: AccordionProps) => (
+	<Accordion {...args} hideToggleLabel={false}>
+		<AccordionRow label="" />
```
SiAdcock, line 29: Do you think stories for Chromatic should have some dummy data too? I think it would be useful to confirm that the data is rendered correctly.

sndrs: ok :+1:

sndrs: fixed in 82d9147

SiAdcock (approved):
Love it! Thanks @sndrs 💎

PR #884, opened 2021-07-27
Bump dependencies (27/07)

jamie-lynch:
* Bump @babel/preset-env from 7.14.7 to 7.14.9
* Bump chromatic from 5.6.1 to 5.9.2
* Bump @emotion/babel-preset-css-prop from 11.0.0 to 11.2.0
* Bump @babel/core from 7.12.10 to 7.14.8
* ~Bump @emotion/react from 11.1.2 to 11.4.0~

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
   "devDependencies": {
     "@babel/core": "^7.14.8",
     "@babel/plugin-proposal-class-properties": "^7.14.5",
-    "@babel/preset-env": "^7.14.7",
+    "@babel/preset-env": "^7.14.9",
     "@babel/preset-typescript": "^7.14.5",
     "@emotion/babel-preset-css-prop": "^11.2.0",
     "@emotion/react": "^11.1.2",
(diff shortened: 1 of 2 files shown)


SiAdcock:
> Bump @emotion/react from 11.1.2 to 11.4.0

I'm expecting this won't change much, since it's only used for Storybook at the workspace root. However, if the components require Emotion 11.1.2 and Storybook is using 11.4.0, is it possible that the components in Storybook will look / behave differently from the components consumed in a production application using 11.1.2?

jamie-lynch:
> > Bump @emotion/react from 11.1.2 to 11.4.0
> 
> I'm expecting this won't change much, since it's only used for Storybook at the workspace root. However, if the components require Emotion 11.1.2 and Storybook is using 11.4.0, is it possible that the components in Storybook will look / behave differently from the components consumed in a production application using 11.1.2?

Ah yes, that's a good point! I'll drop that commit from this PR and we can consider what we want to do with it separately.

jamie-lynch:
Closing to merge bumps individually as there are lots of merge conflicts when combining together

PR #890, opened 2021-07-30
Add TextInput docs to storybook

jamie-lynch:

Include the API docs for the text input package in Storybook to make them more visible to users

## What does this change?

- Add value prop to interface so that doc string can be added

## Screenshots

image

## TODO


src/core/components/text-input/index.tsx
```diff
@@ -29,23 +29,63 @@ const widths: {
...
+	 * Additional text that appears below the label
+	 */
 	supporting?: string;
+	/**
+	 * Determines the width of a fixed-length field, based on the specified character length of the user input. If unspecified,
+	 * the width will default to 100% of the field's container.
+	 */
 	width?: Width;
+	/**
+	 * Whether error styling should apply to this text input. The string appears as an inline error message.
```
SiAdcock, line 58:
```suggestion
	 * The string appears as an inline error message.
```

src/core/components/text-input/index.tsx
```diff
@@ -29,23 +29,63 @@ const widths: {
...
+	 */
 	error?: string;
+	/**
+	 * Whether success styling should apply to this text input. The string appears as an inline success message.
```
SiAdcock, line 62:
```suggestion
	 * The string appears as an inline success message.
```

SiAdcock (approved):
I've made some non-binding suggestions, but this looks awesome! 👍

src/core/components/text-input/stories.tsx
```diff
@@ -17,13 +19,13 @@ const constrainedWith = css`
 	}
 `;
 
-const defaultLight = () => {
+export const Demo = (args: TextInputProps) => {
```
SiAdcock, line 22: Do you think `props` rather than `args` is more React-idiomatic? Or is this some Storybook magic?

jamie-lynch: Good question. I copied this from the accordion example without thinking much about it but maybe props would be nicer... Having said that we're using `Demo.args = {...}` to set the defaults so maybe args is better to be consistent with that?

SiAdcock: I noticed that, so it's probably time to display my ignorance for everyone's amusement 😅  Is `Demo.args = {...}` a Storybook thing, or a TypeScript thing?

jamie-lynch: (AFAIK) It's a storybook thing. It's how we tell storybook what the initial values are so that (as well as passing them to the component) it can set the controls (AKA knobs).

You can see the difference of [using `Demo.args = {...}`](https://guardian.github.io/source/?path=/story/source-src-accordion-accordion--demo) vs [hard coding them in the story](https://guardian.github.io/source/?path=/story/source-src-accordion-accordion--with-cta-labels) if you look at the `hideToggleLabel` control on the accordion stories.

My approach has been to use `Demo.args = {...}` for the Demo story as it's the one that we expect users to interact with. As the other stories are predominately for visual regression testing, I've left the values hard coded. The downside of this is the control don't work for those stories so maybe it's something that we want to adopt everywhere?

SiAdcock: I see, thanks so much for explaining! I don't rally have a strong opinion about any of this, I expect I'll develop feelings over time. Let's leave it the way it is (args for the demo, hardcoded for other stories) and see if a case emerges for args to be the only way of doing things (e.g. it's more consistent, you don't have to think, you know exactly where to look to find the args, etc).

PR #891, opened 2021-07-30
Add TextArea docs to storybook

jamie-lynch:

Include the API docs for the text area package in Storybook to make them more visible to users

## What does this change?

- Remove `cssOverrides` and `className` from props as they are inherited from the `Props` interface already

## Screenshots

image

## TODO


src/core/components/text-area/index.tsx
```diff
@@ -10,26 +9,55 @@ import {
...
+	/**
+	 * Visually hides the label and the "Optional" text set by the `optional` flag.
+	 */
 	hideLabel?: boolean;
+	/**
+	 * Additional text that appears below the label
+	 */
 	supporting?: string;
+	/**
+	 * Whether error styling should apply to this text area. The string appears as an inline error message.
```
SiAdcock, line 39:
```suggestion
	 * Appears as an inline error message.
```

src/core/components/text-area/index.tsx
```diff
@@ -10,26 +9,55 @@ import {
...
+	 */
 	error?: string;
```
SiAdcock, line 41: I think we also need a `success` prop. Maybe we should card this up separately

SiAdcock: I've created #898

SiAdcock (approved):

PR #899, opened 2021-08-02
Add Select API docs to storybook

jamie-lynch:

Include the API docs for the select package in Storybook to make them more visible to users

## What does this change?

- Split the `Select` and `Option` components into their own files

## Screenshots

image

## TODO


src/core/components/select/Select.tsx
```diff
@@ -0,0 +1,122 @@
...
+	supporting?: string;
+	/**
+	 * Whether error styling should apply to this select box. The string appears as an inline error message.
+	 */
+	error?: string;
+	/**
+	 * Whether success styling should apply to this select box. The string appears as an inline success message. This prop should not have a value set at the same time as the error prop. In the event that both are set, errors take precedence.
+	 */
+	success?: string;
+	children: JSX.Element | JSX.Element[];
```
SiAdcock, line 49: It's interesting that we sometimes go with `children: JSX.Element | JSX.Element[]` and sometimes go with `children: ReactNode`. I wonder if there's any useful distinction, or if we should adopt one consistently. Something for outside of this PR.

SiAdcock: Added a [Trello card](https://trello.com/c/BdgSmucL/416-standardise-children-types-for-react-components)

SiAdcock (approved):

src/core/components/select/Select.tsx
jamie-lynch: Interestingly, if I comment out the children prop from the interface everything is still happy. It seems to be picking it up as `React.ReactNode` via the extension of `SelectHTMLAttributes<HTMLSelectElement>`. Maybe we should drop overriding it except for places where there's a particular need to do so?

SiAdcock: Yeah that’s a great idea!

jamie-lynch: I'll update the card you've created as I suspect it's easy to do it as a wholesale change rather than trying to remember as we move the docs across.

PR #900, opened 2021-08-02
Fix "how to use" readme text

jamie-lynch:

This PR fixes some text in the Accordion README in the `How to Use` section.

### DIFF
--- a/src/core/components/accordion/README.md
+++ b/src/core/components/accordion/README.md
@@ -14,4 +14,4 @@ See [storybook](https://guardian.github.io/source/?path=/docs/source-src-accordi
 
 ### How to use
 
-For context and visual guides relating text input usage see the [Source Design System website](https://www.theguardian.design/2a1e5182b/p/38c5aa-accordion).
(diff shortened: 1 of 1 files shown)


src/core/components/accordion/README.md
```diff
@@ -14,4 +14,4 @@ See [storybook](https://guardian.github.io/source/?path=/docs/source-src-accordi
 
 ### How to use
 
-For context and visual guides relating text input usage see the [Source Design System website](https://www.theguardian.design/2a1e5182b/p/38c5aa-accordion).
+For context and visual guides relating to usage see the [Source Design System website](https://www.theguardian.design/2a1e5182b/p/38c5aa-accordion).
```
SiAdcock, line 17:
```suggestion
For context and visual guides relating to accordion usage see the [Source Design System website](https://www.theguardian.design/2a1e5182b/p/38c5aa-accordion).
```

jamie-lynch: I changed this to be generic everywhere (i.e. `relating to usage` vs `relating to ${component} usage`) to reduce the risk of copy-and-paste errors, figuring that the sentence still made sense without it. Happy to add the component names back everywhere if you think that's better though.

Issue #838, opened 2021-06-24
Give radio fieldset a role of radiogroup


SiAdcock:
**Is your feature request related to a problem? Please describe.**

By default, `<fieldset>` elements have a [role of `group`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#technical_summary). Updating the `role` to `radiogroup` adds extra information that screen readers can convey to users.

SiAdcock:
As mention in #902 there is an issue with eslint-plugin-jsx-a11y preventing us from making this change. It should have been fixed in jsx-eslint/eslint-plugin-jsx-a11y#746 but it doesn't appear to be live, or working?

PR #902, opened 2021-08-03
Add API docs for `src-radio`

jamie-lynch:

Include the API docs for the radio package in Storybook to make them more visible to users

## What does this change?

- Split the `Radio` and `RadioGroup` components into their own files

## Screenshots

image

image

## TODO

- [] Update Zeroheight to point to storybook


src/core/components/radio/Radio.tsx
```diff
@@ -0,0 +1,131 @@
...
+	);
+};
+
+export interface RadioProps
+	extends InputHTMLAttributes<HTMLInputElement>,
+		Props {
+	/**
+	 * TODO: Add a description or remove as value is inherited from `InputHTMLAttributes<HTMLInputElement>` (as `string | number | readonly string[] | undefined`)
+	 */
+	value: string;
```
jamie-lynch, line 46: Any thoughts on whether we should add a description or remove this prop and accept the inherited type instead?

src/core/components/radio/Radio.tsx
```diff
@@ -0,0 +1,131 @@
...
+	/**
+	 * Appears to the right of the radio button. If a visible label is undesirable (e.g. for layout reasons) use `aria-label` instead.
+	 *
+	 * If label is omitted, supporting text will not appear either.
+	 */
+	label?: ReactNode; // TODO: Should this be `String | ReactNode`?
+	/**
+	 * Additional text or a component that appears below the label
+	 */
+	supporting?: ReactNode; // TODO: Should this be `String | ReactNode`?
```
jamie-lynch, line 66: Having clicked through the ReactNode type, `string` is a sub type so maybe this is better than saying `string | ReactNode` but equally maybe it's clearer to users that you can just pass a string?

src/core/components/radio/Radio.tsx
```diff
@@ -0,0 +1,131 @@
...
+	/**
+	 * Whether radio button is checked. This is necessary when using the [controlled approach](https://reactjs.org/docs/forms.html#controlled-components) to form state management.
+	 *
+	 * _Note: if you pass the `checked` prop, you MUST also pass an `onChange` handler, or the field will be rendered as read-only._
+	 */
+	checked?: boolean;
+	/**
+	 * TODO: Add a description
+	 */
+	defaultChecked?: boolean;
```
jamie-lynch, line 56: Is "Whether radio button is checked initially" a suitable description? Is it best practice to only use this with the uncontrolled approach?

src/core/components/radio/Radio.tsx
SiAdcock: I'd say remove it. It's not adding any specific value – unless we want to document _all_ props that are inherited from  `InputHTMLAttributes<HTMLInputElement>` 😅

src/core/components/radio/Radio.tsx
SiAdcock, line 56:
```tsx
	/**
	 * When using the [uncontrolled approach](https://reactjs.org/docs/uncontrolled-components.html), use defaultChecked to indicate the initially checked button.
	 */
	defaultChecked?: boolean;
```

src/core/components/radio/Radio.tsx
SiAdcock: I think it's non-obvious that `ReactNode` includes `string`. I've gone around the houses on this many times. I find `string | ReactNode` more descriptive and useful. It might not be technically adding new information, but from most people's perspective, exposing hidden information is effectively the same as adding new information.

src/core/components/radio/Radio.tsx
jamie-lynch: Done in https://github.com/guardian/source/pull/902/commits/7dd7f80c4260edd6a3228a1250940a61742a0b65

src/core/components/radio/Radio.tsx
jamie-lynch: Done in https://github.com/guardian/source/pull/902/commits/99a237c097d4d5c0bd791f0dcb927396afd4f558

src/core/components/radio/RadioGroup.tsx
```diff
@@ -0,0 +1,109 @@
...
+	) : (
+		''
+	);
+	const message = error && (
+		<InlineError id={descriptionId(groupId)}>{error}</InlineError>
+	);
+
+	return (
+		<fieldset
+			aria-invalid={!!error}
```
SiAdcock, line 81: Any chance we can add a role="radiogroup"` while we're here and fix #838

jamie-lynch: I get an eslint error when I do that 😢 

> Non-interactive elements should not be assigned interactive
> [roles.eslintjsx-a11y/no-noninteractive-element-to-interactive-role](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-noninteractive-element-to-interactive-role.md)

SiAdcock: This ought to have been fixed in jsx-eslint/eslint-plugin-jsx-a11y#746 so I'm not sure what's happening there. Let's leave it for now, I'll update #838

src/core/components/radio/radio.stories.tsx
```diff
@@ -1,11 +1,216 @@
-import { Radio } from './index';
```
SiAdcock, line 1: Minor: filename ought to be `Radio.stories.tsx`

jamie-lynch: Ah yes. I thought I'd done this but for it seems that it didn't recognise `radio` -> `Radio` as a change so had to hack it slightly.

src/@types/storybook-emotion-10-fixes.ts
```diff
@@ -13,9 +13,10 @@ export type Parameters = {
 	[key: string]: any;
 };
 
-export type Story = {
-	(arg0: any): JSX.Element;
+export type Story<T = Args> = {
```
sndrs, line 16: is this needed now?

sndrs (approved):
mob reviewed by @guardian/client-side-infra

PR #903, opened 2021-08-03
Add guide for writing stories

jamie-lynch:

This PR adds a document which sets out a guide for writing stories to help ensure consistency and best practice.

## Background

This document is being added during the migration of API docs from [theguardian.design](https://theguardian.design) into Storybook. I have written this up whilst working on #902 so that PR is a good example of the guidelines in action. 

I'm adding it now so that as we continue with the migration and find even more ways to improve the documentation we can record these explicitly for the future, but also so that we can have the discussion around best practice for stories in one place, rather than split across multiple PRs throughout the migration.


docs/10-stories.md
```diff
@@ -0,0 +1,168 @@
+# Stories
+
+Source uses [storybook](https://storybook.js.org/) for its API documentation as well as for visual regression testing.
+
+The Source storybook can be run locally using `yarn storybook`
+
+## Story Configuration
+
+1. Story files should either be names `stories.tsx` or `MyComponent.stories.tsx`
```
jamie-lynch, line 9: Should we just have `MyComponent.stories.tsx` everywhere? Does that also rely on having `MyComponent.tsx` everywhere and `index.ts` files that only contain exports (even in packages with a single component)?

sndrs: i think is the approach i'd favour. it's also where we plan to be in v4 i think? i've taken this approach so far with the branding and accordion ones and i dont think it was too disruptive

docs/10-stories.md
```diff
@@ -0,0 +1,168 @@
...
+1. Story titles should follow the pattern `Source/${package}/${component}` e.g. `Source/src-radio/RadioGroup`
```
sndrs, line 10: or `Editorial/*/*` (for now)

docs/10-stories.md
```diff
@@ -0,0 +1,168 @@
...
+1. Each component should have a Demo story which is designed for documentation purposes and a series of stories detailing specific cases for visual regression testing
+1. Stories should have PascalCase names
+1. The `args` property should be passed to all stories so that the source code is displayed on the docs page
+1. Props should all be commented to autofill the args table
+1. Components should all have doc strings to auto fill the docs page (see [example](#example) below)
```
sndrs, line 15: did you mean as comments?

```suggestion
1. Components should all have JSDoc comments to auto fill the docs page (see [example](#example) below)
```

docs/10-stories.md
```diff
@@ -0,0 +1,168 @@
...
+1. Controls should be disabled for all stories except the Demo story
```
sndrs, line 16: this is an interesting one. you can use the JSDoc `@ignore` comment to hide controls (example in the [accordion](https://github.com/guardian/source/blob/31af12ae2574addae2e75bd349c92f7e51a0f545/src/core/components/accordion/AccordionRow.tsx#L38), which also gives you a place to add comments explaining why. 

i think i prefer this, but is it doing the same thing?

jamie-lynch: Disabling the controls for the non-demo stories removes the whole controls tab from those stories. For those stories at the moment, we're hard coding the args we want, rather than using `Story.args = {...}`. This means that the controls don't actually work for any of the hard coded values. I opted for removing the controls tab as it's confusing to have non-working controls. An alternative would be to use `Story.args = {...}` everywhere so that the controls did work. I didn't go for this as it was more effort and these stories are predominantly for visual regression testing but happy to go with this approach instead.

The following point is disabling an individual control, similar to the example for Accordion. The difference is that it still displays in the arg table, just without a control. I was doing that for props like `cssOverrides` on the basis that it's useful to show that this prop is available but the control won't do anything in storybook so it's better to remove it. For the AccordionRow example, I'd agree that removing it from the args table completely is the best option, as we don't really want to document this as a prop that a user should pass directly.

docs/10-stories.md
jamie-lynch: I'll open a PR to make this consistent across the components that have already been migrated.

docs/10-stories.md
```diff
@@ -0,0 +1,168 @@
...
+    export const Story = ...
+
+    Story.story = {
+        parameters: {
+            controls: {
+                disabled: true
+            }
+        }
+    }
+    ```
```
SiAdcock, line 40: What is the difference between these two examples? Do they do exactly the same thing, or are there special circumstances in which it is necessary to use one over the other?

docs/10-stories.md
```diff
@@ -0,0 +1,168 @@
...
+	},
+	args: {
+		label: 'Label',
+		hideLabel: false,
+		value: 'Value',
+	},
+};
+
+export const NoValue = (args: MyComponentProps) => (
+	<MyComponent {...args} label="Label"/>
```
SiAdcock, line 126: If we're not passing default `args` via the `NoValue.args = {...}` pattern, is it useful to spread the args into the component like this? Is it doing anything?

docs/10-stories.md
jamie-lynch: They do the same thing. I've included it because a number of stories already use either `Story.parameters = {...}` or `Story.story = {...}` so I was hoping to make it clear that you can add it to an existing statement rather than having to add another. On reflection, that isn't particularly clear from this example so maybe let's remove?

docs/10-stories.md
```diff
@@ -0,0 +1,168 @@
...
+### `string | JSX.Element` Controls
+
+For props where the type is either a string or some complex element, it may be desirable to configure the control to be a text type so that users can try it out more easily in Storybook.
+
+```ts
+interface StoryProps {
+    label: string | JSX.Element;
+}
+
+export const Story = ({label}: StoryProps ) => ...
```
SiAdcock, line 157: AdcockLint™️

```suggestion
export const Story = ({ label }: StoryProps) => ...
```

docs/10-stories.md
jamie-lynch: It's not, good point. The only thing that we need is to pass an arg into the function so that Storybook generates the source for the docs page. We could instead do:

```tsx
export const NoValue = (_: MyComponentProps) => (
  <MyComponent label="Label"/>
)
```

docs/10-stories.md
SiAdcock: Oh I see, thanks for explaining. Yes, maybe we should try to standardise on one or the other? It just makes reading, and knowing what approach to take, much easier for future devs

docs/10-stories.md
SiAdcock: Oh wow, that's unusual. So if you don't specifically define an argument, the docs don't get generated?

docs/10-stories.md
SiAdcock: How do we feel about components with lots and lots of stories, like `<Radio>`? (see #902) Currently, stories are split into separate files and grouped vaguely by the prop the stories are demonstrating. Do we want to collapse all these stories into a single file?

docs/10-stories.md
jamie-lynch: Correct - for example:

## With Args

```tsx
export const LegendSupportingTextLight = (args: RadioGroupProps) => (
	<RadioGroup
		{...args}
		name="colours"
		label="Select your preferred colour"
		supporting="You can always change it later"
	>
		{radiosOneSelected.map((radio, index) =>
			React.cloneElement(radio, { key: index }),
		)}
	</RadioGroup>
);
```

or

```tsx
export const LegendSupportingTextLight = (_: RadioGroupProps) => (
	<RadioGroup
		name="colours"
		label="Select your preferred colour"
		supporting="You can always change it later"
	>
		{radiosOneSelected.map((radio, index) =>
			React.cloneElement(radio, { key: index }),
		)}
	</RadioGroup>
);
```

image

## Without Args

```tsx
export const LegendSupportingTextLight = () => (
	<RadioGroup
		name="colours"
		label="Select your preferred colour"
		supporting="You can always change it later"
	>
		{radiosOneSelected.map((radio, index) =>
			React.cloneElement(radio, { key: index }),
		)}
	</RadioGroup>
);
```

image

docs/10-stories.md
jamie-lynch: Yep, makes sense. Let's go for `Story.story = {...}` as it's the higher level object.

docs/10-stories.md
jamie-lynch: I've started #905 to modify the components that we've already migrated API docs for to conform with these recommendations.

On multiple vs one file, I started moving all of the radio stories into one file (well, one each for `Radio` and `RadioGroup`) and then decided against it but I'm leaning back towards it now. I've also added some card around reducing duplication in stories via [decorators](https://storybook.js.org/docs/react/writing-stories/decorators#story-decorators) and [templates](https://storybook.js.org/docs/react/writing-stories/decorators#story-decorators) which would make the single files less unwieldy.

docs/10-stories.md
jamie-lynch: https://github.com/guardian/source/pull/903/commits/1c9d41642d8850dd1d6e252e2a3a6929b47126c5 adds a point about removing args completely using `@ignore` if they shouldn't be shown in docs

docs/10-stories.md
jamie-lynch: https://github.com/guardian/source/pull/902/commits/af04304ebe1f8a41b00247af892afb6483fd8fa5 in #902 consolidates the `Radio` and `RadioGroup` stories into one file per component

sndrs: 👍 from me for one `stories` file per component

docs/10-stories.md
sndrs: 👍

docs/10-stories.md
sndrs: storybook must be able to introspect useful info from you explicitly giving the args type, even if you don't use it?

docs/10-stories.md
```diff
@@ -0,0 +1,171 @@
...
+        },
+        argTypes: {
+            arg2: {
+                control: null
+            }
+        }
+    }
+    ```
+
+1. Any args that should not be shown in docs should be marked with the `@ignore` annotation and a comment explaining why
```
sndrs, line 47:
```suggestion
1. Any props that should not be documented should be marked with the `@ignore` annotation and a comment explaining why
```

docs/10-stories.md
jamie-lynch: Are we happy with this then?

```tsx
export const LegendSupportingTextLight = (_: RadioGroupProps) => (
	<RadioGroup
		name="colours"
		label="Select your preferred colour"
		supporting="You can always change it later"
	>
		{radiosOneSelected.map((radio, index) =>
			React.cloneElement(radio, { key: index }),
		)}
	</RadioGroup>
);
```

sndrs:
how about a rule where we don't use emotion? if you need to style things, you can use the `style` prop, but maybe best is show things unstyled? that's the raw component being demoed then. e.g.

image

sndrs:
is it useful though to style the demo? i feel like that should show us what you get out the box, rather than an example of using it?

jamie-lynch:
> is it useful though to style the demo? i feel like that should show us what you get out the box, rather than an example of using it?

Yeah, maybe for the Demo it should always just be the component on it's own. I'll update the recommendations to include that.

I'd argue that some of the styling is useful for stories that are used for visual regression testing as it can [test them in context](https://guardian.github.io/source/?path=/story/buttonlink--inline-button-link) or [test multiple instances of the component in one story](https://guardian.github.io/source/?path=/story/buttonlink--text-and-icon-button-links). I've posted a couple more ideas for how to improve those code blocks in the `Improve storybook code blocks` card on the backlog. Plus we can add a recommendation to minimise styling in stories.

sndrs:
that's a good point – maybe we just don't want code for the chromatic examples? only the demo with the knobs?

sndrs (approved):
love it!

SiAdcock:
Amazing work, thanks @sndrs and @jamie-lynch 💎

PR #905, opened 2021-08-03
Update migrated stories to conform to recommendations

jamie-lynch:

This PR updates the stories for components where documentation has already been migrated from zeroheight to conform to the recommendations agreed in #903 

## What does this change?

-  Update the stories file in Accordion to `Accordion.stories.tsx`
-  Create a new `TextArea.tsx` file, exports only from `index.tsx` and update stories file to `TextArea.stories.tsx`
-  Create a new `TextInput.tsx` file, exports only from `index.tsx` and update stories file to `TextInput.stories.tsx`


jamie-lynch:
Based on a number of new things that we've learnt, the recommendations are likely to change significantly. I'm closing this PR now so that we can tackle updating everywhere in one go once the recommendations are updated.

PR #909, opened 2021-08-05
Move API docs for `src-brand`

sndrs:

- move the API docs into storybook
- remove some inadvertent exports
- make brand components configurable

## What does this change?

- add API docs to storybook
- rename component files to pascal-case version of their export
- add a story for each component
- make the brand components configurable with `width` and `textColor` props (where appropriate)
- remove `SvgRoundelFilled` and make `SvgRoundel` filled (since they're all filled really)
- **breaking** – remove undocumented export of `SvgRoundel`
  - the only use of it was removed first in https://github.com/guardian/support-dotcom-components/pull/501

## Screenshots

image


src/core/brand/SvgGuardianLiveLogo.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+		width: {
+			control: { type: 'range', min: 10, max: 600 },
+		},
+	},
+};
+
+export const Demo = (args: SvgGuardianLiveLogoProps) => (
+	<SvgGuardianLiveLogo {...args} />
+);
+Demo.storyName = 'SvgGuardianLiveLogo';
```
jamie-lynch, line 19: Would it be better just to name the story `SvgGuardianLiveLogo` rather than overriding?

```suggestion
export const SvgGuardianLiveLogo = (args: SvgGuardianLiveLogoProps) => (
	<SvgGuardianLiveLogo {...args} />
);
```

jamie-lynch: Also, should we make all of the `Demo` stories just be the name of the component?

src/core/brand/SvgRoundelBrand.tsx
```diff
@@ -0,0 +1,17 @@
+import { SvgRoundel } from './SvgRoundel';
+import { brand } from '@guardian/src-foundations/palette';
+
+export type SvgRoundelBrandProps = {
+	/**
+	 * The width the SVG will display at (height is automatically adjusted
+	 * to maintain the correct aspect ratio).
+	 */
+	width?: number;
+};
```
jamie-lynch, line 10: Probably one for future PR but I think we should abstract this into a shared type to save some duplication

src/core/brand/SvgRoundel.tsx
```diff
@@ -0,0 +1,47 @@
...
+	 * [`@guardian/src-foundation`](https://theguardian.design/2a1e5182b/p/2668c8-code).
+	 */
+	textColor?: string;
+	/**
+	 * The roundel's background colour.
+	 *
+	 * This should probably come from
+	 * [`@guardian/src-foundation`](https://theguardian.design/2a1e5182b/p/2668c8-code).
+	 */
+	bgColor?: string;
```
SiAdcock, line 15: Can we go with `backgroundColor` for consistency with how the colour tokens are named?

src/core/brand/SvgRoundelBrand.tsx
sndrs: i did actually do that originally, but wasn't sure if it was better to be explicit per component? easy either way

src/core/brand/SvgGuardianLiveLogo.stories.tsx
sndrs: i did this here because these stories don't have multiple states, so storybook hoists the single story to the top level. wihtout the rename you either have to import `SvgGuardianLiveLogo` `as` something else so you can give the story the right name (`SvgGuardianLiveLogo`) or the nav shows `Demo` as the name, rather than expose the export name

src/core/brand/SvgRoundel.tsx
sndrs: fixed in 8eebde9

jamie-lynch (approved):

PR #910, opened 2021-08-06
Move Link API docs to storybook

jamie-lynch:

Include the API docs for the link package in Storybook to make them more visible to users

## What does this change?

- Split the `Link` and `ButtonLink` components into their own files
  - Move shared code into `types.ts`, `styles.ts` and `shared.ts`
- Don't redefine props from the SharedProps interface
- Update the stories definition to match the agreed styles

I've made a lot of changes to the stories so that they now render a single (Button)Link per story. This makes the stories themselves much simpler and removes the need for any styling.

### List of Stories

image

### Link Playground

image

### Button Link Playground

image

## TODO

- [] Update Zeroheight to point to storybook


src/core/components/link/ButtonLink.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+);
+
+// *****************************************************************************
+
+export const Playground = Template.bind({});
+asPlayground(Playground);
+
+// *****************************************************************************
+
+export const PrimaryButtonLinkLightTheme = (args: ButtonLinkProps) => (
```
sndrs, line 41: we should align on whether `Light` is the same as `Default`. Right now I couldn't say i really know, which i'm guessing means other people might be in the same situation?

src/core/components/link/ButtonLink.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { SharedLinkProps } from './types';
+import { linkStyles } from './styles';
+import { linkContents } from './shared';
+
+export interface ButtonLinkProps
+	extends ButtonHTMLAttributes<HTMLButtonElement>,
+		SharedLinkProps {}
+
+/**
+ * [Storybook](https://guardian.github.io/source/?path=/docs/source-src-link-button-link--demo) •
```
sndrs, line 11:
```suggestion
 * [Storybook](https://guardian.github.io/source/?path=/docs/source-src-link-button-link) •
```

this should force storybook to open the first story whatever it's called (in this not `demo` any more?)

src/core/components/link/Link.stories.tsx
```diff
@@ -0,0 +1,296 @@
...
+	</ThemeProvider>
+);
+RightIconLinkLightTheme.args = {
+	iconSide: 'right',
+};
+asChromaticStory(RightIconLinkLightTheme);
+
+// *****************************************************************************
+
+// export const PriorityLight = (_: LinkProps) => (
```
sndrs, line 172: do we need to keep this stuff?

jamie-lynch: Whoops🤦‍♂️  Nope, I'll remove that

jamie-lynch: Done in https://github.com/guardian/source/pull/910/commits/112d4cef13f3484cbac81f7554c120094cc61cfe

src/core/components/link/ButtonLink.stories.tsx
jamie-lynch: Good point. They do seem to be used interchangeably at the moment which can be a bit confusing. @SiAdcock any thoughts on which we should standardise on?

sndrs: (not a blocker for this though – we'll need to align all of one set the other way)

sndrs (approved):
🔗:shipit:

PR #911, opened 2021-08-06
Emotion 11-ify theming docs

SiAdcock:

The docs still use the theming API for Emotion 10

## What does this change?

-   replace `emotion-theming` with `@emotion/react`
-   update link to Theming docs


jamie-lynch (approved):
📝

PR #913, opened 2021-08-06
Move API docs for `label`/`legend`

sndrs:
## What does this change?

- breaks the components out into individual files
- documents all states of the components with theme examples
- makes `hideLabel` and `optional` default to `false`, which the docs currently say they do but they don't

## Screenshots

image


sndrs:
@SiAdcock i have _zero_ idea what's going on here! https://github.com/guardian/source/pull/913/checks?check_run_id=3263170564#step:4:32

SiAdcock (approved):
Yass! 💎

PR #919, opened 2021-08-11
Add story category helpers

sndrs:

- create notion of docs stories and canvas stories
- add methods that help write them

## What does this change?

- adds `docs` and `canvas` story helper
- `docs` stories are only viewable with the docs tab and have controls
  - as now, they need to be the first export of a `*.stories.tsx` file
- `canvas` stories are only viewable with a `canvas` tab, and have no controls
- uses these in `src-footer` as an example

### demo story (`docs`)

image

### non-demo story (`canvas`)
image


SiAdcock:
This looks great and a huge improvement to how the stories are presented.

Thinking about alternative APIs, I had an idea that was more minimalist and closer to the way Storybook docs look:

```ts
import { docs, canvas } from '../../../lib/story-categories'

export const Demo = TemplateWithChildren.bind({})
Demo.args = {
	showBackToTop: false
}
Demo.parameters = {
	...doc
}

export const DefaultTablet = TemplateWithChildren.bind({})
DefaultTablet.parameters: {
	...canvas
}
```

What do you think?

sndrs:
100% agree.

as discussed in chat, converted to a decorator function that augments the `parameters` property of a standard story instead

SiAdcock (approved):
Amazing work thanks @sndrs! 💎

jamie-lynch:
This looks awesome! 🎉

PR #921, opened 2021-08-12
story defaults spruce

sndrs:
- move accessibility tab to the start (also now open by default)
- close the code preview in docs by default
- move the 🧶 to the start of the playground menu item

image


SiAdcock (approved):
Good calls! 👍

PR #925, opened 2021-08-13
add API docs for `src-layout/Hide`

sndrs:

Add API docs for `src-layout/Hide`. 

In the interests of a small PR, I'll do the other components individually.

## What does this change?

- pulls `Hide` up to top level in `layout`
- deprecates `above` and `below` props in favour of `until` and `from`
  - `above` actually means at-and-above which is not clear
  - aligns with `src-foundations/mq` which uses `until` and `from` (also [sass-mq](https://sass-mq.github.io/sass-mq/#undefined-mixin-mq)) https://github.com/guardian/source/blob/36a6c9f65e693c892c7903e0524bcb4e1840f60b/src/core/foundations/src/mq/index.ts#L147

## Screenshots

image


src/core/components/layout/Hide/Hide.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+	argTypes: {
+		until: {}, // display first
+		from: {}, // display first
+		above: {
+			control: { disable: true },
+		},
+		below: {
+			control: { disable: true },
+		},
+	},
```
jamie-lynch, line 17: Can we disable the `cssOverrides` controls as they don't work in storybook anyway?

```suggestion
	argTypes: {
		until: {}, // display first
		from: {}, // display first
		above: {
			control: { disable: true },
		},
		below: {
			control: { disable: true },
		},
		cssOverrides: {
			control: { disable: true },
		},
	},
```

jamie-lynch: As agreed offline, let's lift this up into the helper function rather than adding everywhere

src/core/components/layout/Hide/Hide.stories.tsx
```diff
@@ -0,0 +1,77 @@
+import { Hide, HideProps } from './Hide';
+import { asChromaticStory, asPlayground } from '../../../../lib/story-intents';
+import { Story } from '../../../../@types/storybook-emotion-10-fixes';
+
+export default {
+	title: 'Source/src-layout/Hide',
+	component: Hide,
+	argTypes: {
+		until: {}, // display first
+		from: {}, // display first
```
jamie-lynch, line 10: Did we decide we're just going to get rid of these for now then?

sndrs: nice one

jamie-lynch (approved):

PR #931, opened 2021-08-16
Add API docs for `src-layout/Container`

sndrs:

Add API docs for `src-layout/Container`. 

## What does this change?

- pulls `Container` up to top level in `layout`
- adds API docs


src/core/components/layout/Container/Container.tsx
```diff
@@ -1,25 +1,35 @@
...
+	/**
+	 * **Deprecated**
+	 *
+	 * Use `sideBorders` instead.
+	 */
+	border?: boolean;
 	sideBorders?: boolean;
 	topBorder?: boolean;
 	backgroundColor?: string;
 	borderColor?: string;
```
jamie-lynch, line 21: These are fairly self-evident but is it worth commenting these anyway?

src/core/components/layout/Container/Container.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+export default {
+	title: 'Source/src-layout/Container',
+	component: Container,
+	argTypes: {
+		border: {
+			control: {
+				disable: true,
+			},
+		},
+	},
```
jamie-lynch, line 14: Can we disable the `cssOverrides` control as it doesn't really work in storybook

```suggestion
	argTypes: {
		border: {
			control: {
				disable: true,
			},
		},
		cssOverrides: {
			control: {
				disable: true,
			},
		},
	},
```

jamie-lynch: As agreed offline, let's lift this up into the helper function rather than adding it everywhere

jamie-lynch (approved):

PR #934, opened 2021-08-17
provide default options for `cssOverrides`

sndrs:

give all `cssOverrides` props in docs view a default set of options (`undefined` or a css object)

I was hoping to fix the default story always being canvas view but it looks like it's being worked on in https://github.com/storybookjs/storybook/issues/13128.

sorting the args table would also be good but [it's coming in 6.4](https://github.com/storybookjs/storybook/commit/3be68b5ed7ddb70154701e783604258efcdb080e#diff-53d1d90b71220b0003266dab7e033c99619f5b22fffdea24d5686f0eb688ce1fR313).

### DIFF
--- a/src/core/helpers/types.ts
+++ b/src/core/helpers/types.ts
@@ -5,7 +5,7 @@ export type ThemeName = 'default' | 'brand' | 'brandAlt';
 export interface Props {
 	className?: string;
 	/**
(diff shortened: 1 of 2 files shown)


jamie-lynch:
Nice! 🔴 

What did we decide about the `className` prop? Should that be ignored completely or just the control disabled?

sndrs:
good point! i've removed it globally from `playground` stories args tables

jamie-lynch (approved):
👍

PR #935, opened 2021-08-17
feat: Added `src-ed-logo`

oliverlloyd:
Adds `Logo`, a new component that returns the three different TheGuardian logos that are currently possible. In addition, this component will adjust the size fo the logo based on breakpoint

## Why?
The main benefit here is in standardising the logo sizing but I think it is also useful to put different logo designs into the editorial Source repo so that they are available elsewhere if needed.

### Mobile
Screenshot 2021-08-17 at 15 41 22

### Tablet
Screenshot 2021-08-17 at 15 41 15

### Desktop
Screenshot 2021-08-17 at 15 41 07


HarryFischer:
@oliverlloyd Let me send you some new SVG code – the yellows aren't quite correct on these

SiAdcock:
Thanks @oliverlloyd for raising this. Do you think [`@guardian/src-brand`](https://github.com/guardian/source/tree/main/src/core/brand) would be a better location for logos?

oliverlloyd:
> Thanks @oliverlloyd for raising this. Do you think [`@guardian/src-brand`](https://github.com/guardian/source/tree/main/src/core/brand) would be a better location for logos?

I considered that but then the idea of different versions of the logo based on the paper's anniversary or awards seemed more 'editorial' - it also just feels easier adding things there rather than the 'core brand' area. But I'm happy either way, maybe it's one for the designers?

oliverlloyd:
> > Thanks @oliverlloyd for raising this. Do you think [`@guardian/src-brand`](https://github.com/guardian/source/tree/main/src/core/brand) would be a better location for logos?
> 
> I considered that but then the idea of different versions of the logo based on the paper's anniversary or awards seemed more 'editorial' - it also just feels easier adding things there rather than the 'core brand' area. But I'm happy either way, maybe it's one for the designers?

Note. This component is taking the existing logo svg, and adding sizing logic to it which specifically relates to placement on a page. But I'm also adding two additional svgs. So maybe we could keep `Editorial/Logo` here, with its sizing, but move the two new svgs over into Brand?

src/editorial/web/components/logo/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+	);
+};
+
+const BestWebsiteLogo = () => {
+	return (
+		<div css={svgSize}>
+			<svg
+				fill="none"
+				xmlns="http://www.w3.org/2000/svg"
+				viewBox="0 0 236 92"
```
SiAdcock, line 47: Is there any way we can get the viewbox size consistent with the standard logo ([300x100](https://github.com/guardian/source/blob/2ffa82a5c366b461b8ae18ec9d6fcbe988501d75/src/core/brand/SvgGuardianLogo.tsx#L29))?

oliverlloyd: Yeah, it would be great if these were consistent. @HarryFischer ?

SiAdcock:
> adding sizing logic to it which specifically relates to placement on a page

I'm just thinking, is it useful to have the sizing applied here? How large or small these logos appear seems to be more of an application-level concern, rather than something dictated by the design system. For instance, as a designer I'd want the flexibility to make these logos to have the right size for a particular design, rather than to be constrained by code in a shared component library. Please let me know if I have missed some context in which the design team have agreed that the pixel values here do represent the  widths that these logos should always take.

oliverlloyd:
> > adding sizing logic to it which specifically relates to placement on a page
> 
> I'm just thinking, is it useful to have the sizing applied here? How large or small these logos appear seems to be more of an application-level concern, rather than something dictated by the design system. For instance, as a designer I'd want the flexibility for these logos to have the right size for a particular design, rather than to be constrained by code in a shared component library. Please let me know if I have missed some context in which the design team have agreed that the pixel values here do represent the widths that these logos should always take 😄

The extra context that's in my head is that this component will (maybe) eventually live inside a shared `Header` component.

In addition, the requirement that led to this PR is for the logo used by Identity to 'have the same sizes' as used on Dotcom. From clicking around it was clear that all logos do (or should) also follow the same pattern, hence the centralisation. I think allowing different applications to decide different sizes actually creates problems

src/editorial/web/components/logo/index.tsx
oliverlloyd: This is now done. It did mean that we now no longer use the Source version of the standard logo though because that has a different viewbox. Do you think it's worth aligning everything to the same values as exist in Source or should that logo be updated to match these?

SiAdcock:
As discussed offline:

1. I'm sold on the usefulness of this component. Let's get it in! 🚀 
1. When we build the development kitchen, we're considering putting this and all other editorial components into the kitchen to begin with.
1. Regarding the actual SVGs, we should use the existing `<SvgGuardianLogo />` that lives in `src-brand` for the standard logo, as you had coded originally
1. For the other two SVGs, it would be nice if they could conform to the 300x100 viewbox of the standard logo. If not, can we try and aim for 300x120?

src/editorial/web/components/logo/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+					fill="#fff"
+				/>
+			</svg>
```
SiAdcock, line 48: Sorry to ask, but what is stopping us using the existing logo SVG `<SvgGuardianLogo>`? Is it because the viewbox is different to the other logos, and hence it causes issues with height and width? 

If so, is it possible to specify only `width` for `svgSize` and allow `height` to be calculated dynamically? Or the other way around? I'm just a bit wary of having two sources of truth for logo SVGs.

HarryFischer: Maybe I misunderstood your request, I thought you wanted them all to have the same size viewbox.
I think just specifying a px value for width and then height:auto would help us keep consistency here(thats how we currently are scaling them in the DCR header). If any future logo variants always have "The Guardian" as the widest element in the SVG we wouldn't run into issues?

SiAdcock: Apologies for the misunderstanding, I wanted to ask if the _Anniversary_ and _Newspaper of the Year_ logos could have the same size viewbox as the _Standard_ logo, and if not, could they have the same size viewbox as each other. I think you're right that `height: auto` should be safe and future-proof, especially if that's what we're already doing in DCR 💪 

@oliverlloyd does this sound okay to you?

oliverlloyd: grilla-yes

SiAdcock (approved):
Looks good to me! 👍

PR #936, opened 2021-08-17
Add API docs for `src-layout/Inline`

sndrs:

Add API docs for `src-layout/Inline`. 

## What does this change?

- pulls `Inline` up to top level in `layout`
- adds API docs


src/core/components/layout/Inline/Inline.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+import { Inline, InlineProps } from './Inline';
+import { asChromaticStory, asPlayground } from '../../../../lib/story-intents';
+
+export default {
+	title: 'Source/src-layout/Inline',
+	component: Inline,
+};
+
+// export * from '../stories/inline/default';
+
```
jamie-lynch, line 12:
```suggestion
```

jamie-lynch (approved):
🚢

PR #939, opened 2021-08-18
Add API docs for `src-layout/Tiles`

sndrs:

Add API docs for `src-layout/Tiles`. 

## What does this change?

- pulls `Tiles` up to top level in `layout`
- adds API docs
- allow `asPlayground` and `asChromaticStory` to accept stories of any prop types
- deprecates `collapseBelow` in favour of `collapseUntil` (pinging @SiAdcock – this ok with you?)
  - brings the wording into alignment with `Hide` and `mq`
- mixes in the `cssOverrides` prop, which was ignored before


src/core/components/layout/Tiles/Tiles.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+Columns5CollapseBelowTabletAtTablet.args = {
+	columns: '5',
+};
+Columns5CollapseBelowTabletAtTablet.parameters = {
+	viewport: { defaultViewport: 'tablet' },
+};
+asChromaticStory(Columns5CollapseBelowTabletAtTablet);
+
+// export * from './stories/tiles/default';
+// export * from './stories/tiles/collapse-below';
```
jamie-lynch, line 109: Can these go now?

```suggestion

// export * from './stories/tiles/default';
// export * from './stories/tiles/collapse-below';
```

src/lib/story-intents.ts
```diff
@@ -10,7 +10,7 @@ import { Story } from '../@types/storybook-emotion-10-fixes';
  * Make sure all props are configurable in storybook's controls table.
  *
  */
-export const asPlayground = (story: Story) => {
+export const asPlayground = (story: Story<any>) => {
```
jamie-lynch, line 13: Out of interest, what does having the `<any>` and the corresponding `Story<TilesProps>` do? Is it so that typescript knows what args to expect? Should we update the recommendations to match this? And is it worth adding an eslint ignore so that we don't get this warning?

sndrs: TS was really not happy with:

```typescript
const Template: Story = (args: TilesProps) => ( ... )
```
saying that `Story<Args>` did not contain `columns` that `TilesProps` requires (which is true). 

`const Template: Story<TilesProps>` resolve that, but then `asChromaticStory`/`asPlayground` complained that the `Story<Args>` they expected were not compatible with `Story<TilesProps>` (again true).

so now they just take a `Story` with any old thing passed into the main function.

this might not be the right approach...

SiAdcock: We could make `asChromaticStory` / `asPlayground` accept a generic type that gets passed to the `Story` type in their signature

e.g.

```ts
// story-intents.ts

export const asPlayground = <T>(story: Story<T>) => {
```

SiAdcock:
```suggestion
export const asPlayground = <Args>(story: Story<Args>) => {
```

src/lib/story-intents.ts
```diff
@@ -49,7 +49,7 @@ export const asPlayground = (story: Story) => {
  *
  * Make sure you have one of these for every possible state of your component.
  */
-export const asChromaticStory = (story: Story) => {
+export const asChromaticStory = (story: Story<any>) => {
```
SiAdcock, line 52:
```suggestion
export const asChromaticStory = <Args>(story: Story<Args>) => {
```

SiAdcock:
@sndrs this looks great, I'm on board with the renaming of `collapseBelow`. Feel free to add stories and update the component to use `collapseUntil` – or I can if you don't get around to it before your holidays!

SiAdcock (approved):

PR #940, opened 2021-08-19
Add API docs for `src-layout/Columns`

sndrs:

Add API docs for `src-layout/Columns`. 

## What does this change?

- [x] pulls `Columns` up to top level in `layout`
- [x] splits out `Column` into its own file
- [x] adds API docs
- [x] deprecates `collapseBelow` in favour of `collapseUntil`
  - brings the wording into alignment with other components and `mq`
- [ ] add stories for `Column` in isolation
  - is it useful to do this? 
  - you would't actually use it isolation
  - is this a good case to adopt a `<Columns.Column />` type approach?


SiAdcock (approved):
This is great. I think you're right not to worry about `<Column>` stories, a `Column` on its own is not very useful 👍

PR #941, opened 2021-08-20
Add `ThemeProviderDecorator`

jamie-lynch:

This change adds a global decorator which wraps the story in a `ThemeProvider` when a `theme` parameter is present. This means that themed stories can use `Template.bind({})` like any other story and pass the theme in as a parameter, saving on LOC.

## What does this change?

-   Add a new `ThemeProviderDecorator` and include in `.storybook/preview.tsx`
-   Refactor themed stories in `Radio.stories.tsx` to pass the theme in as a parameter

## TODO

- [x] Add some documentation
- [x] Apply to all stories


SiAdcock (approved):
This is awesome 💎

PR #943, opened 2021-08-20
fix: Ensure focus halo covers svg children

oliverlloyd:
The `Link` component sets an a tag wrapping any children passed to it but the focus halo on this anchor wasn't covering its children. Setting `display: block;` resolves this.

### Before
Screenshot 2021-08-20 at 15 59 30

### After
Screenshot 2021-08-20 at 15 58 30

### DIFF
--- a/src/editorial/web/components/logo/index.tsx
+++ b/src/editorial/web/components/logo/index.tsx
@@ -118,6 +118,8 @@ export const Logo = ({ logoType = 'standard' }: Props) => (
 		title="The Guardian Homepage"
 		subdued={true}
 		cssOverrides={css`
+			/* Adding block display here so that the focus halo correctly covers the content */
+			display: block;
 			color: ${brandText.primary};
 			:hover {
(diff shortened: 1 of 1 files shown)


shtukas (approved):

SiAdcock (approved):
Thanks @oliverlloyd, happy to merge to get this moving. 👍 

Can you think of a better general solution for this problem in the `<Link>` component?

oliverlloyd:
> Can you think of a better general solution for this problem in the `<Link>` component?

Not really. [This PR](https://github.com/guardian/source/pull/464) introduced `display: inline` for `Link` so that it flows well with text so we can't just bluntly set all links to `block` and adding a prop seems to choose between inline or block seems like we'd be overloading that component.

Maybe there's another way to solve the halo issue generically?

SiAdcock:
Thanks @oliverlloyd, I've added a Trello ticket to pick it up. Let's get this merged

PR #945, opened 2021-08-24
Add playground to icon storybook

SiAdcock:
## What does this change?

-  Add an Icons playground
-  Refactor icons stories to use args
- Add a `size` prop to all icons, which can take the values `'xsmall' | 'small' | 'medium'`, mapping to [values found in `@guardian/src-foundations/size`](https://github.com/guardian/source/blob/827a5a42460e1db17fae28d0a31cde3212f38158/src/core/foundations/src/size/global.ts#L27-L31)
- Note that because of its non-standard shape, `SvgDirectDebitWide` varies its size on the `height` on the `<svg>` rather than the `width`.
- Removed hardcoded widths from brand-coloured logo icons
- Removed hardcoded fill from `SvgPerson`

The `src-icons` stories are inconsistent with the `src-brand` stories because:

- There are so many icons, it seems like overkill to have a playground and Chromatic story for each one. Happy to take a steer on this
- The acceptable sizes of these icons is prescribed by our [size tokens](https://github.com/guardian/source/blob/827a5a42460e1db17fae28d0a31cde3212f38158/src/core/foundations/src/size/global.ts#L27-L31), so a `width` slider would be inappropriate
- The allowed colours of icons is prescribed by our [colourways](https://guardian.zeroheight.com/styleguide/s/15078/p/1377a6-tokens/b/293ddb), so a free-text hex colour editor would be inappropriate

We should revisit to make `src-brand` and `src-icons` documentation more consistent.


jamie-lynch:
The stories look good but it does make the code and props sections of the docs page slightly confusing. Maybe it's possible to override the code section and just ignore those props as (AFAIU) they're not actually props of the Icon, just related to the implementation of the story.

image

SiAdcock:
@jamie-lynch yes, thanks for picking that up. The API of icons is not fully realised and as a result the stories come with their own implementation for demonstration purposes. With `src-brand` we added the API as part of the documentation migration. Happy to give that a go now, but it feels a bit like scope creep?

I'll have a play with the code view to see if I can get something that doesn't look quite so magical!

jamie-lynch (approved):
Looks great 👍

PR #946, opened 2021-08-25
Update Accordion stories to latest recommendations

jamie-lynch:

This change updates the Accordion stories to use the latest [recommendations](https://github.com/guardian/source/blob/main/docs/10-stories.md)


jamie-lynch:
This does change the `WithCTALabels` and `WithoutCTALabels` stories to use the same `AccordionRows` as the playground which are slightly different to before but I don't think it makes any difference.

SiAdcock (approved):
Looks great, thanks! 💎

PR #947, opened 2021-08-25
Update Select stories to latest recommendations

jamie-lynch:

This change updates the Select stories to use the latest [recommendations](https://github.com/guardian/source/blob/main/docs/10-stories.md)


jamie-lynch:
As part of this, I've removed the `constrainedWidth` css from the template so that it didn't appear in the code block. Any thoughts on whether it should have remained for the Playground and/or the Chromatic stories?

SiAdcock:
> Any thoughts on whether it should have remained for the Playground and/or the Chromatic stories?

Let's try and keep custom CSS out of stories. It looks alright how you've implemented it

src/core/components/select/Select.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+				value={state}
+				onChange={(event) => setState(event.target.value)}
+			/>
+			<span css={message}>
+				{state ? `You have selected: ${state}` : ''}
+			</span>
+		</>
+	);
+};
+asChromaticStory(ControlledExample);
```
SiAdcock, line 125: We can probably get rid of this story, it's not really doing anything useful on Chromatic. If we need to demonstrate a code example of a controlled component, we can direct people to the [story in source-shakedown](https://github.com/guardian/source-shakedown/blob/main/stories/select/controlled.tsx)

SiAdcock (approved):
One small suggestion but this looks grand 🎹

PR #950, opened 2021-08-25
Update user feedback stories to latest recommendations

jamie-lynch:

This change updates the TextInput stories to use the latest [recommendations](https://github.com/guardian/source/blob/main/docs/10-stories.md)

### DIFF
--- a/src/core/components/user-feedback/InlineError.stories.tsx
+++ b/src/core/components/user-feedback/InlineError.stories.tsx
@@ -1,4 +1,3 @@
-import { ThemeProvider } from '@emotion/react';
 import {
 	userFeedbackDefault,
 	userFeedbackBrand,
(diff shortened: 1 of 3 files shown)


jamie-lynch:
I'm getting this error locally but I'm not sure why it's only happening for user feedback...

```
File '<path>/source/src/@types/storybook-emotion-10-fixes.ts' is not under 'rootDir' '<path>/source/src/core/components/user-feedback'. 'rootDir' is expected to contain all source files
```

SiAdcock:
> I'm getting this error locally

Is that on build, or in your editor? I'm also seeing that in VS Code in `Icons.stories.tsx`

SiAdcock (approved):

jamie-lynch:
> > I'm getting this error locally
> 
> Is that on build, or in your editor? I'm also seeing that in VS Code in `Icons.stories.tsx`

Both. `yarn validate` works fine but `yarn build:all` throws that same error.

jamie-lynch:
> > > I'm getting this error locally
> > 
> > 
> > Is that on build, or in your editor? I'm also seeing that in VS Code in `Icons.stories.tsx`
> 
> Both. `yarn validate` works fine but `yarn build:all` throws that same error.

Will this block new releases if we're running them locally?

jamie-lynch:
> I'm getting this error locally but I'm not sure why it's only happening for user feedback...
> 
> ```
> File '<path>/source/src/@types/storybook-emotion-10-fixes.ts' is not under 'rootDir' '<path>/source/src/core/components/user-feedback'. 'rootDir' is expected to contain all source files
> ```

Turns out this was down to me forgetting to update the `exclude` value in the `tsconfig` when updating the story file name.

PR #966, opened 2021-09-09
Expand CSS Resets - Add "meyer" resets

OllysCoding:

DCR Currently uses the `reset-css` library, then expands with the `defaults` exported in `resets.ts`. (Note currently DCR does not import from source, but this should be changed soon)

We're looking to standardise the resets used across DCR & AR, as such it makes sense to move them all into source so they can be imported easily.

DCR Currently has to use babel-plugin-preval to import the CSS rest from the `reset-css` library, however it would be preferred if we could import it as a string directly. This would also be useful on AR, which does not use a custom babel config.

## What does this change?

-   export a new `meyer` (named after the creator of the resets) string of resets

I'm interested on peoples thoughts of the naming of the export, or whether it should be included in the `defaults` export (which would make it a breaking change).

### Documentation

-   [ ] Examples in Storybook


jamie-lynch:
Looks like a bug with the `QuoteIcon` stories slipped through to main which is why the UI tests are failing here. #967 should fix that so you can just merge/rebase when that's been merged to main.

OllysCoding:
Cheers for sorting that Jamie :D

sndrs:
i think it's a good idea to add a standardised reset here, and probably over time it should replace the existing ones entirely.

i think i'd go for something like a new export simply called `resetCSS`? something that implies it covers the whole lot and is intended for universal use and is clear what it's doing when you see it used down the page from the import.

i'm not sure about being explicit about `meyer` though, for two reasons:

1. it's already available and adding it here doesn't help a lot
2. it's really an implementation detail. what we're trying to offer is a standardised reset and you don't need to know how it's done if you don't want to.

i do think it's probably worth DCR leading the way here, thoguh, and making this export be whatever the thing that works well for DCR right now. 

capturing that in a reusable form for other sites/applications that run on DCR will be v useful.

OllysCoding:
Since DCR uses both, my preference would be to export a single `resetCSS` string as you suggested, which includes all the resets that DCR does. I'd just want to note that it would then become a breaking change as any platforms that do consume this (I don't know if there are any at the moment?) could face broken CSS as a result.

In this case would we remove the `defaults` that is currently exported, or do something like 

```
const resetCSS = `${defaults} ... <meyers stuff>`
```

SiAdcock:
Thanks for this @OllysCoding.

There's a handful of places using `defaults` so I wouldn't want to spring a whole new reset upon them.

My preference, as you have suggested, is to provide an export called `resetCSS` which mixes in `defaults`. DCR, and anything integrating with DCR, would include this in their styles. In time, this would become the only reset we export from here, and everyone who wants to use Source would include this reset in their styles.

It's worth noting that there are a number of accessibility issues with Meyer, performance questions around the heavy-handed nature of this reset and questions around the inclusion of styles to support browsers that are now considered legacy. We can and should try to address these issues, but we should do so centrally, here in Source, rather than piecemeal across different applications.

OllysCoding:
Thanks for the feedback @SiAdcock 

I've update the PR as per your suggestions. I completely agree, I think it makes sense to update and improve on these resets, and it'll be great to do that in one unified place :)

SiAdcock (approved):
Amazing, thanks @OllysCoding 💎

SiAdcock (commented):
This is great, but with one last caveat. We've moved the source code for resets from `src/core/foundations/src/utils/resets.ts` to [`packages/@guardian/source-foundations/src/utils/resest.ts`](https://github.com/guardian/source/blob/main/packages/@guardian/source-foundations/src/utils/resets.ts). Would you be able to migrate your changes to this location? 🙏

SiAdcock (commented):
~This is great, but with one last caveat. We've moved the source code for resets from `src/core/foundations/src/utils/resets.ts` to [`packages/@guardian/source-foundations/src/utils/resest.ts`](https://github.com/guardian/source/blob/main/packages/@guardian/source-foundations/src/utils/resets.ts). Would you be able to migrate your changes to this location? 🙏 ~

Sorry for the duplicate comment 😓  @OllysCoding I can help with this migration if you're stretched

jamie-lynch (approved):

sndrs (approved):

OllysCoding:
👍 This looks great, thank you, are we good to merge? @SiAdcock

PR #969, opened 2021-09-10
Add vendor prefixes to properties in `@supports` declaration

SiAdcock:

Stylis, the CSS preprocessor used by Emotion under the hood, [doesn't add vendor prefixes to CSS properties in `@supports` declarations](https://github.com/thysultan/stylis.js/issues/34). This is a design decision that they are unlikely to change.

As a result, browsers that don't fully support properties used in these declarations (usually [`appearance`](https://caniuse.com/?search=appearance)) will not render styles declared inside the `@supports` block.

The workaround I've chosen here is to manually add vendor prefixes to the appearance property, when it appears inside a `@supports` declaration.

## What does this change?

Adds `-webkit-` and `-moz-` prefixes to `appearance` in the following components:

-   Radio
-   Checkbox
-   Select

## Screenshots

**Before (Safari 14.1.2)**

Screenshot 2021-09-10 at 14 08 00
Screenshot 2021-09-10 at 14 08 06
Screenshot 2021-09-10 at 14 08 12

**After(Safari 14.1.2)**

Screenshot 2021-09-10 at 14 07 13
Screenshot 2021-09-10 at 14 07 20
Screenshot 2021-09-10 at 14 07 27

### DIFF
--- a/src/core/components/checkbox/styles.ts
+++ b/src/core/components/checkbox/styles.ts
@@ -4,7 +4,7 @@ import { height, width } from '@guardian/src-foundations/size';
 import { textSans } from '@guardian/src-foundations/typography';
 import { focusHalo } from '@guardian/src-foundations/accessibility';
 import { checkboxDefault } from '@guardian/src-foundations/themes';
-import { resets } from '@guardian/src-foundations/utils';
+import { resets, appearance } from '@guardian/src-foundations/utils';
 
 export const fieldset = css`
 	${resets.fieldset};
@@ -52,7 +52,7 @@ export const checkbox = ({ checkbox } = checkboxDefault) => css`
 		${focusHalo};
 	}
 
-	@supports (appearance: none) {
+	@supports (${appearance}) {
 		appearance: none;
 		&:checked {
 			border: 2px solid ${checkbox.borderChecked};
@@ -95,7 +95,10 @@ export const supportingText = ({ checkbox } = checkboxDefault) => css`
 `;
 
 export const tick = ({ checkbox } = checkboxDefault) => css`
-	@supports (appearance: none) {
+	@supports (
+		(appearance: none) or (-webkit-appearance: none) or
+			(-moz-appearance: none)
+	) {
 		/* overall positional properties */
 		position: absolute;
 		width: 6px;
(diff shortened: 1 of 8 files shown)


jamie-lynch (approved):
Nice! 👍

src/core/components/checkbox/styles.ts
```diff
@@ -52,7 +52,10 @@ export const checkbox = ({ checkbox } = checkboxDefault) => css`
 		${focusHalo};
 	}
 
-	@supports (appearance: none) {
+	@supports (
+		(appearance: none) or (-webkit-appearance: none) or
+			(-moz-appearance: none)
+	) {
```
sndrs, line 58: is it worth pulling this out somewhere?

```suggestion
	@supports (${appearance}) {
```

sndrs: nbd tho

SiAdcock: Good shout, I expect this will come up again. The only place that makes sense right now is `@guardian/src-helpers` as it doesn't sound like a `@guardian/src-foundations` thing.

src/core/foundations/src/utils/css.ts
```diff
@@ -0,0 +1,2 @@
+export const appearance =
+	'(appearance: none) or (-webkit-appearance: none) or (-moz-appearance: none)';
```
jamie-lynch, line 2: Following #981, this should probably be in `packages/@guardian/source-foundations/src/utils` with a symlink in `src/core/foundations/src/utils`

sndrs: `packages/@guardian/source-foundations/src/utils/supports-queries.ts`?

src/core/foundations/src/utils/supports-queries.ts
```diff
@@ -0,0 +1,6 @@
+// FYI
+// src/core/foundations/src/utils/supports-queries.ts SYMLINKS TO
+// packages/@guardian/source-foundations/src/utils/supports-queries.ts
+
+export const appearance =
+	'(appearance: none) or (-webkit-appearance: none) or (-moz-appearance: none)';
```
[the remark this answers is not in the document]
SiAdcock: Haha the perils of using `ln path/to/target` instead of `ln -s path/to/target`. Great catch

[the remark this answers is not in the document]
SiAdcock: Thanks @jamie-lynch I have fixed 😅

jamie-lynch (approved):

PR #971, opened 2021-09-14
add `source-foundations` PoC

sndrs:

Add a proof of concept for `source-*` packages creation/migration path. 

This should allow us to create and start publishing the new packages while keeping the old ones maintainable until we're ready to make the switch.

## What does this change?

It creates a new `@guardian/source-foundations` package, discrete from the current repo set up. 

_I didn't want to change the existing stuff, but I'm intending that whenever we stop working on `src-*` packages and only work on `source-*`, the set up in here (eslint config, rollup etc) can be pulled up to the top level to simplify it all a bit._

- add `./packages` to workspace
- add `@guardian/source-foundations` package
- expose the animation exports from `@guardian/src-foundations` (`transitions`)
- build the package with rollup
  - this has benefit of there not being multiple exports with the same name in the final bundle (e.g.  `export { transitions };` in `animation.ts` _and_ `index.ts` where it's 'offcially' exported to users)
- lint with guardian configs


SiAdcock:
This looks great, and not a symlink in sight!

> this has benefit of there not being multiple exports with the same name in the final bundle (e.g. export { transitions }; in animation.ts and index.ts where it's 'officially' exported to users)

Sorry if I'm stating the obvious, but do you mean that by using rollup, it would prevent the consumer from importing transitions from `@guardian/source-foundations/animation`, only `@guardian/source-foundations`, or something like that?

sndrs:
Exactly yeah. If you use it without importing, vscode will only find one possible place to get it from

SiAdcock:
Wow that’s pretty neat. Yay rollup!

jamie-lynch (commented):
This looks good to me. Would the next step be to extend this PR to migrate the rest of foundations or merge this and continue to migrate over the remaining parts of Source in small chunks?

sndrs:
@jamie-lynch just to capture our offline chat for future travellers – exactly yeah. if we try and keep the PRs to one change per PR and slowly build up the new libs, hopefully it will be easier to understand each change as it comes along

PR #976, opened 2021-09-20
Move animations source code to `source-foundations`

sndrs:

Move `animation` to `source-foundation` and share the code `src-foundation` with a symbolic link.

### Why?

- we want to move the source code into `source-` packages 
- we want `source-` packages to be as close to final as possible (not dependent on `src-` implementation details)
- we need to maintain `src-` until `source-` is released
- typescript won't easily compile useful `dist`s for `src-` packages when the source code lives so far away
  - typescript is grumpy playmate if you have a complex platform game of folder locations
  - multiple tsconfigs and builds compounds this

Therefore, this tries to bridge the new world and the old by moving the source to its new home but symlinking to it from the old one.

Symlinking _to_ the new location also means when we delete the `src-` stuff, the symlinks will go, leaving only the finished new world in place 🌄.

## What does this change?

- moves animation stuff to `source-foundations`
- creates symlink back to original file

### DIFF
--- a/packages/@guardian/source-foundations/src/animation.ts
+++ b/packages/@guardian/source-foundations/src/animation.ts
@@ -1,0 +1,11 @@
+// FYI
+// src/core/foundations/src/animation.ts SYMLINKS TO
+// packages/@guardian/source-foundations/src/animation.ts
+
+const transitions = {
+	short: '.2s cubic-bezier(.64, .57, .67, 1.53)',
+	medium: '.3s ease-in-out',
+	long: '.65s ease-in-out',
+};
+
+export { transitions };
--- a/packages/@guardian/source-foundations/src/index.ts
+++ b/packages/@guardian/source-foundations/src/index.ts
@@ -1,1 +1,2 @@
-export { transitions } from '../../../../src/core/foundations/src/animation';
+// animation
+export { transitions } from './animation';
--- a/src/core/foundations/src/animation.ts
+++ b/src/core/foundations/src/animation.ts
@@ -1,9 +1,1 @@
-import { transitions as _transitions } from './theme';
-
-const transitions = {
-	short: _transitions[0],
-	medium: _transitions[1],
-	long: _transitions[2],
-};
-
-export { transitions };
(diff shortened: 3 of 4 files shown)


joecowton1 (approved):

packages/@guardian/source-foundations/src/index.ts
```diff
@@ -1 +1,2 @@
-export { transitions } from '../../../../src/core/foundations/src/animation';
+// animation
+export { transitions } from './animation';
```
jamie-lynch, line 2: Just to jog my memory, did we decide to flatten foundations in v4 (i.e only export from the top level) or are we going to maintain the existing import structure for now?

packages/@guardian/source-foundations/src/animation.ts
```diff
@@ -0,0 +1,11 @@
+// FYI
+// src/core/foundations/src/animation.ts SYMLINKS TO
+// packages/@guardian/source-foundations/src/animation.ts
```
jamie-lynch, line 3: It might be worth making it more explicit that changing things here will also change the currently published packages and to therefore be careful.

jamie-lynch (commented):
I like the approach of symlinking this way round so we can just delete the old files when we want to deprecate the `src-*` packages. Can we add a note either in the files or at the top level to give some context and explain where to make changes for anyone who comes across this in the meantime?

packages/@guardian/source-foundations/src/index.ts
sndrs: discussed offline – flatten and rename export names where they clash, then see how many that is

packages/@guardian/source-foundations/src/animation.ts
sndrs: as discussed offine, boosted perms required in repo before merging 🤞

SiAdcock (approved):
Approved!

joecowton1 (approved):

jamie-lynch (approved):
🥇

PR #981, opened 2021-09-21
Move everything else except typography to source-foundations

jamie-lynch:

Move everything else except typography to `source-foundation` and share the code `src-foundation` with a symbolic link.

### Why?

Building on #977 

## What does this change?

- [x] Move mq stuff to `source-foundations`
- [x] Move size stuff to `source-foundations`
- [x] Move rest of utils stuff to `source-foundations`
- [x] Add a note about themes
- [x] Add symlinks to new files
- [x] Update migration breaking changes record


packages/@guardian/source-foundations/src/utils/focus-style-manager.ts
```diff
@@ -0,0 +1,85 @@
...
+	public start(): void {
+		this.container.addEventListener('mousedown', this.handleMouseDown);
+		this.isRunning = true;
+	}
+
+	/** Disable interaction mode behavior and remove className from container. */
+	public stop(): void {
+		this.reset();
+		this.isRunning = false;
+	}
```
jamie-lynch, line 39: I've modified this to fix some eslint errors but not sure if it's better to leave as-is?

packages/@guardian/source-foundations/src/utils/focus-style-manager.ts
```diff
@@ -0,0 +1,85 @@
...
+		this.container.addEventListener(
+			'keydown',
+			this.handleKeyDown as (evt: Event) => void,
+		);
+	};
+}
+
+const FOCUS_DISABLED = 'src-focus-disabled';
+
+let _focusEngine: InteractionModeEngine | undefined;
```
jamie-lynch, line 69: I've added `| undefined` here to fix an eslint error.

packages/@guardian/source-foundations/src/utils/focus-style-manager.ts
sndrs: i think it's good to have it here – it's our at this point, even if derived from blueprint (none of their changes will show up here unless we make them)

PR #982, opened 2021-09-21
move typography to source-foundations

sndrs:

Move `typography` to `source-foundations` and share the code in `src-foundations` with a symbolic link.

### Why?

Building on #977 

## What does this change?

- moves typography stuff to `source-foundations`
- creates symlink back to original file
- add a migration breaking changes record
- Reinstate `@guardian/src-foundations/theme` to avoid a breaking change


jamie-lynch (approved):
🧱

PR #963, opened 2021-08-31
Create Development Kitchen

jamie-lynch:
## What is the purpose of this change?

This change adds the [Development Kitchen](https://github.com/guardian/source/discussions/833).

## What does this change?

- [x] Rename `@guardian/editorial` package in package.json to `@guardian/source-kitchen-react`
- [x] Reset the version number for `@guardian/source-kitchen-react`
- [x] Rename [editorial folder](https://github.com/guardian/source/tree/main/src/editorial) to kitchen
- [x] Delete Android, iOS and Web directory level
- [x] Update [README.md](https://github.com/guardian/source/tree/main/src/editorial/web) with Dev Kitchen intro and stuff from [discussion](https://github.com/guardian/source/discussions/833)
- [x] Update [scripts](https://github.com/guardian/source/search?q=editorial) to remove editorial paths
- [x] Make kitchen components not independently publishable (i.e. delete package.jsons)
- [x] Add release mechanism for kitchen
- [x] Update [contribution guidelines](https://github.com/guardian/source/blob/main/docs/08-contributing.md#guidelines)
- [x] Add CODEOWNERS for current components

## Before we merge, we must also:

- [x] Release first version of kitchen from this branch
- [x] Deprecate individual editorial (now kitchen) components on NPM
- [x] Deprecate `@guardian/editorial` on NPM
- [x] Write comms to the department (to be sent after we merge)

PR #984, opened 2021-09-22
Remove editorial workspaces from package.json as they no longer exist

jamie-lynch:

This change drops the editorial workspaces from the `package.json` file as, following #963, they no longer exist.

## What does this change?

- Remove `src/editorial/web` and `src/editorial/web/components/*` workspaces from `package.json`


SiAdcock (approved):
✨

PR #985, opened 2021-09-22
Add shakedown package

SiAdcock:

To have a way to verify that the packages produce maintain the expected public API.

The [source-shakedown](https://github.com/guardian/source-shakedown) repo is currently serving this purpose, however it has a few problems:

- it is not very discoverable
- it requires a pre-release of Source to be published
- it's hard to keep up to date

Bringing the shakedown concept into the Source repo should help reduce the impact of these problems, while opening the possibility for better automation of our end-to-end smoke tests. 

## What does this change?

This changes creates a new `source-shakedown` package within the Source repository which contains:

1. A test file which verifies that exports from Source packages have not changed

    This works similar to a snapshot test in that, if anything does change, the test must be updated. This ensures that subsequent releases haven't broken the package structure or unexpectedly removed exports from the public API. It uses jest to run the test.

2. A `.ts.` file which verifies that no types have been removed

    This test works by importing all of the types from all packages into a file and then running `tsc`. It will capture types that have been removed but not fail when new types are added so, for now, the onus is on contributors and reviewers to remember to add new types to this file.
    
It also:

1.  Adds the `shakedown` and `shakedown:shakedown` commands to the root `package.json` which alias the workspace command and run the shakedown tests respectively
2. Runs the `shakedown:shakedown` command as part of the CI checks
3. Updates the `build:all` and `clean:all` commands to include the foundations and kitchen packages
4. Updates the `build` command and adds a `clean` command to the foundations `package.json` to support the previous step


jamie-lynch:
> copy stories for each component, adjusting paths to point to dist directories

Will they be looking at `dist` or will they just be installing the packages from npm?

> hook up CI to run Shakedown Storybook

If it's the packages from npm then maybe it isn't worth having the CI check at the moment. Definitely something that'd be great to have in future though.

SiAdcock:
> Will they be looking at dist or will they just be installing the packages from npm?

I think dist _might_ be better because then all contributors can test the changes they are making locally to ensure they run in a production-like environment before they merge. Saying that, it would _also_ be useful to have the end-to-end coverage that installing from NPM would provide. Maybe if we can do easy, automatic, safe pre-releases on demand, this would become the preferred way?

jamie-lynch:
Do we want to migrate the SSR part of shakedown over too?

SiAdcock:
@jamie-lynch the SSR test is not in a great state in source-shakedown, so we'd probably need to rebuild that from scratch here too. It was very useful in the beginning, catching some of the earlier problems with modules running in CommonJS, but it hasn't caught any issues recently. I see it as nice to have – the 20% basically

SiAdcock:
@sndrs where would you say we are with this? I like the test approach, do you think it's within reach?

.storybook/main.js
```diff
@@ -2,8 +2,8 @@ module.exports = {
 	stories: [
 		'../src/**/*.stories.mdx',
 		'../src/**/*stories.@(js|jsx|ts|tsx)',
-		'../packages/**/*.stories.mdx',
-		'../packages/**/*stories.@(js|jsx|ts|tsx)',
+		'../packages/@guardian/source-react-components-development-kitchen/**/*.stories.mdx',
+		'../packages/@guardian/source-react-components-development-kitchen/**/*stories.@(js|jsx|ts|tsx)',
```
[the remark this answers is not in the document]
SiAdcock: Yeah we should absolutely revert this. It was a quick hack for an earlier experiment and is no longer needed

package.json
```diff
@@ -11,10 +11,10 @@
   ],
   "scripts": {
     "build-storybook": "build-storybook -o dist",
-    "build:all": "ts-node ./scripts/build-all",
+    "build:all": "ts-node ./scripts/build-all && yarn foundations build && yarn kitchen build",
     "ci:build": "NODE_ENV=production build-storybook -o dist",
     "clean": "rm -rf dist",
-    "clean:all": "ts-node ./scripts/clean-all",
+    "clean:all": "ts-node ./scripts/clean-all && yarn foundations clean && yarn kitchen clean",
```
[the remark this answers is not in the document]
SiAdcock: I'm reasonably sure the kitchen already gets built and cleaned correctly by these scripts.

We can run `yarn build` and `yarn clean` from within the `build-all` and `clean-all` scripts for the new foundations package. We'd need to expose a new path in the [`paths`](https://github.com/guardian/source/blob/main/scripts/paths.ts) module and consume it in `build-all` and `clean-all`

packages/@guardian/source-foundations/package.json
```diff
@@ -13,7 +13,8 @@
     "dist"
   ],
   "scripts": {
-    "build": "rm -rf dist && rollup -c",
+    "build": "yarn clean && rollup -c",
+    "clean": "rm -rf dist",
```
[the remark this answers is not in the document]
jamie-lynch: This is now done in https://github.com/guardian/source/pull/996. I'll drop this commit and rebase once that's merged.

package.json
```diff
@@ -95,7 +98,7 @@
     "eslint-plugin-react": "^7.24.0",
     "execa": "^5.0.0",
     "husky": "^7.0.1",
-    "jest": "^26.6.0",
+    "jest": "^27.2.3",
```
jamie-lynch, line 101: This is now done in https://github.com/guardian/source/pull/995. I'll rebase once that's merged.

jamie-lynch:
The things left to do are:

- [ ] Work out why the build is failing
- [ ] Add the types from the various paths within `@guardian/src-foundations`
- [ ] Make sure that all the imports are coming from `dist` (the root level `.tsconfig.json` provides some paths which might not resolve to the `dist` directory)

jamie-lynch:
It looks like the build is failing due to an out of memory error 😢

jamie-lynch:
> Add the types from the various paths within `@guardian/src-foundations`

I'm not sure that we'll be able to do this as it doesn't seem to be possible to import from a specific file using the current method. Given that this is only a problem for `@guardian/src-foundations`, which is hopefully going away soon, does it seem reasonable to go without?

SiAdcock:
Thanks for your continued efforts on this @jamie-lynch 😊

> Given that this is only a problem for @guardian/src-foundations, which is hopefully going away soon, does it seem reasonable to go without?

This sounds reasonable to me. The reason this work was prioritised was to help validate the migration of components into `@guardian/source-react-components`. The Foundations work has already been completed, so validating `@guardian/src-foundations` seems a bit redundant now. 

Is this logic sound?

packages/@guardian/source-package-validation/validate-types.ts
```diff
@@ -0,0 +1,183 @@
+import type {
```
sndrs, line 1: could this just be `export`?
```suggestion
export type {
```

yarn.lock
```diff
@@ -3320,6 +3320,14 @@
...
     pretty-format "^27.0.0"
 
+"@types/jest@^27.0.2":
+  version "27.0.2"
+  resolved "https://registry.yarnpkg.com/@types/jest/-/jest-27.0.2.tgz#ac383c4d4aaddd29bbf2b916d8d105c304a5fcd7"
+  integrity sha512-4dRxkS/AFX0c5XW6IPMNOydLn2tEhNhJV7DnYK+0bjoJZ+QTmfucBlihX7aoEsh/ocYtkLC73UbnBXBXIxsULA==
+  dependencies:
+    jest-diff "^27.0.0"
+    pretty-format "^27.0.0"
+
```
SiAdcock, line 3330: Is this still needed? It doesn't look like any new deps have been installed

package.json
```diff
@@ -22,6 +22,8 @@
     "lint": "yarn lint:js && yarn lint:styles",
     "lint:js": "eslint . --ext .ts,.tsx,.js",
     "lint:styles": "stylelint src/core/**/styles.ts",
+    "package-validation": "yarn workspace @guardian/source-package-validation",
+    "package-validation:validate": "yarn build:all && yarn package-validation validate",
```
sndrs, line 26: having this as an actual packages in the workspace seems to blur the lines between what are maintenance scripts and what are the things we/contributors are building.

what do you think about something like this?

```suggestion
    "validate-dist": "tsc ./scripts/validate-dist/types.ts && jest ./scripts/validate-dist/package-exports.test.ts",
```

this would also need `build` to have run of course, but doing that manually seems ok?

docs/08-contributing.md
```diff
@@ -64,4 +64,4 @@ $ yarn storybook
 ### Raising a pull request
 
 -   Don’t bump the package version as part of a pull request. Maintainers will take care of this as part of the release process.
--   Test that your component can be used in a real project (e.g. [dotcom-rendering](https://github.com/guardian/dotcom-rendering) or the specially-designed [source-shakedown](https://github.com/guardian/source-shakedown)). Use [`yarn link`](https://yarnpkg.com/en/docs/cli/link) to avoid having to perform a real publish, and [`yarn pack`](https://yarnpkg.com/en/docs/cli/pack) to see exactly which files would be published.
+-   Test that your component can be used in a real project (e.g. [dotcom-rendering](https://github.com/guardian/dotcom-rendering)). Use [`yarn link`](https://yarnpkg.com/en/docs/cli/link) to avoid having to perform a real publish, and [`yarn pack`](https://yarnpkg.com/en/docs/cli/pack) to see exactly which files would be published.
```
SiAdcock, line 67:
```suggestion
-   Test that your component can be used in a real project (e.g. [dotcom-rendering](https://github.com/guardian/dotcom-rendering)). Use [`yarn link`](https://yarnpkg.com/en/docs/cli/link) to avoid having to perform a real publish, and [`yarn pack`](https://yarnpkg.com/en/docs/cli/pack) to see exactly which files would be published.
- If your change extends the API of a component or the foundations (e.g. you are adding a new component, a new prop or exporting a new type), please update the [`source-package-validation`](packages/@guardian/source-package-validation/README.md) package.
```

packages/@guardian/source-package-validation/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+It runs two checks:
+
+1. Verify that no exports have changed in the published packages
+
+    This works similar to a snapshot test in that, if anything does change, the test must be updated. This ensures that subsequent releases haven't broken the package structure or unexpectedly removed exports from the public API. It uses jest to run the test.
+
+2. Verify that no types have been removed
+
+    This test works by importing all of the types from all packages into a file and then running `tsc`. It will capture types that have been removed but not fail when new types are added so, for now, the onus is on contributors and reviewers to remember to add new types to this file.
```
sndrs, line 15:
```suggestion
    This test works by importing all of the types from all packages into a file and then running `tsc`. It will capture types that have been removed but not fail when new types are added so, for now, the onus is on contributors and reviewers to **remember to add new types to this file**.
```

yarn.lock
jamie-lynch: Ah. It looks like it was in there twice. Running `yarn` didn't remove it so I've deleted it manually. Good spot 👁️

packages/@guardian/source-package-validation/validate-types.ts
jamie-lynch: Good thinking! Done in https://github.com/guardian/source/pull/985/commits/5122ba3e3f135cdc62063c08e77c8e5de1ac630b

package.json
SiAdcock: Yeah, I think you're right. When we were recreating `source-shakedown` locally, it seemed to make sense as a package in its own right. But now, it's boiled down to just a couple of scripts and dependencies that come from the workspace root.

jamie-lynch: I agree that this should all live in scripts rather than packages to distinguish what's internal and what is a published package. I'll move that now.

For developing locally, I think having the ease of running `yarn package-validation test:types` and `yarn package-validation test:components` is fairly important. Given that these require custom ts and jest configs, I don't think it's a great experience to have to remember and type `yarn jest ./scripts/validate-dist/package-exports.test.ts --config=./scripts/validate-dist/jest.config.js`. 

Having them in their own package removes the need to explicitly point at the config files and the workspace simplifies the running of the individual commands from the root level without adding multiple commands to the root `package.json` so I'd prefer we either keep it or find a nicer way of achieving the same thing / something similar.

sndrs: > Having them in their own package removes the need to explicitly point at the config files and the workspace simplifies the running of the individual commands from the root level without adding multiple commands to the root package.json so I'd prefer we either keep it or find a nicer way of achieving the same thing / something similar.

this makes a lot of sense 👍 

do you think we still need to set up the yarn alias in this case? maybe something like this?

```suggestion
    "validate-dist": "yarn workspace validate-dist validate",
```

jamie-lynch: Yeah, that looks good. I'll update it to be `validate-dist` everywhere.

scripts/validate-dist/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@guardian/source-validate-dist",
+  "version": "1.0.0",
```
sndrs, line 3: since we won't publish this we should make sure it can't be by accident 😉 

we can probably also do away with the scope/context etc?

```suggestion
  "name": "validate-dist",
  "version": "1.0.0",
  "private": true,
```

package.json
```diff
@@ -35,6 +36,7 @@
     "test:coverage": "jest --coverage",
     "tsc": "tsc",
     "validate": "yarn tsc && yarn lint && yarn test",
+    "validate-dist": "yarn build:all && yarn workspace @guardian/source-validate-dist validate",
```
sndrs:
```suggestion
    "validate-dist": "yarn build:all && yarn workspace validate-dist validate",
```

(ignore this if you don't want to go with the suggestion on the package's own `package.json` below)

jamie-lynch: 👍 This should also be done in https://github.com/guardian/source/pull/985/commits/b307fe3df664048fe31f4e707a5cf5089a08bad6

sndrs (approved):
didn't expect this to become such a bruiser!

think it's ended up in a great place though, nice one for keeping at it

jamie-lynch:
> didn't expect this to become such a bruiser!
> 
> think it's ended up in a great place though, nice one for keeping at it

Thanks for all of the feedback. Definitely feels like it's in a better place now.

sndrs:
image

PR #987, opened 2021-09-27
Add coveralls coverage checker in CI

jamie-lynch:

This change adds coverage checking to the CI checks to ensure that future changes don't negatively effect test coverage.

## What does this change?

- Add `test:coverage` and `validate:ci` scripts to `package.json`
- Add [coverallsapp/github-action](https://github.com/coverallsapp/github-action) action to _ci_ workflow
- Update _ci_ workflow to run `validate:ci`
- Add `coverage/` to _.gitignore_


SiAdcock (approved):

package.json
```diff
@@ -36,8 +36,10 @@
     "storybook": "start-storybook -p 6006",
     "tag:beta:all": "ts-node ./scripts/tag-beta-all",
     "test": "jest",
+    "test:coverage": "jest --coverage",
     "tsc": "tsc",
     "validate": "yarn tsc && yarn lint && yarn test",
+    "validate:ci": "yarn tsc && yarn lint && yarn test:coverage --ci",
```
SiAdcock, line 42:
```suggestion
    "validate:ci": "yarn validate && yarn test:coverage --ci",
```

.github/workflows/ci.yml
```diff
@@ -16,5 +16,9 @@ jobs:
       - name: build
         run: |
           yarn install --frozen-lockfile
-          yarn validate
+          yarn validate:ci
           yarn ci:build
```
SiAdcock, line 20: Aside: we should probably rename this script to `build:ci`

SiAdcock (approved):
Blessings 🙏

.github/workflows/ci.yml
jamie-lynch: Yeah, that'd be good. I'll fire another PR over with that in.

package.json
jamie-lynch: This would mean running the tests twice which I'm not such a fan of, but then less duplication is good so 🤷‍♂️

PR #988, opened 2021-09-27
Cache dependencies in CI

jamie-lynch:

This change adds caching to the dependency installation steps in CI workflows.

## What does this change?

-   Add configuration parameter to [guardian/actions-setup-node](https://github.com/guardian/actions-setup-node) enable caching
-   Change install step to add `--frozen-lockfile` to ensure repeatable runs

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -9,9 +9,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2.3.4
       - uses: guardian/actions-setup-node@main
+        with:
+          cache: 'yarn'
       - name: build
         run: |
-          yarn
+          yarn install --frozen-lockfile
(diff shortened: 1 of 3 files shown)


.github/workflows/cd.yml
```diff
@@ -9,9 +9,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2.3.4
       - uses: guardian/actions-setup-node@main
```
sndrs, line 11: should we target `v2.4` we're touching all instances?

jamie-lynch: As in `guardian/actions-setup-node@v2.4` instead of `@main`? I don't have any strong feeling on which is best so yeah, happy to change these.

SiAdcock (approved):

PR #990, opened 2021-09-28
Mark the development kitchen as not having side effects

jamie-lynch:

This change adds the `sideEffects` option to the development kitchen to (hopefully) ensure that the package is tree shaken when used in applications. This is particularly important now that we have combined multiple components into a single package.

## What does this change?

-   Add `"sideEffects": false` to development kitchen `package.json`


sndrs (approved):
Have been able to test it? Fingers crossed!

SiAdcock (approved):
Down with side effects 💯

jamie-lynch:
> Have been able to test it? Fingers crossed!

Sorry, forgot to comment here. Yeah, I've tested it on DCR and the bundle size is down so I think we're good.

https://github.com/guardian/dotcom-rendering/pull/3434

PR #995, opened 2021-09-30
Bump jest version

jamie-lynch:

This change updates the version of jest installed in Source, partly because being up-to-date is a good thing but also so that we can install the latest version of `ts-node` in [this PR](https://github.com/guardian/source/pull/985) and have the two play nicely together. 

## What does this change?

-   Update the desired version of jest in the package.json
-   Run `yarn` to update the yarn.lock


sndrs (approved):
ship it

PR #996, opened 2021-09-30
Include source foundations in build:all and clean:all

jamie-lynch:

At the moment the `build:all` and `clean:all` commands don't include the `@guardian/source-foundations` package. This is confusing.

## What does this change?

-   Add a `sourceFoundations` export to the `paths` script
-   Add `sourceFoundations` to the `clean-all` script
-   Add `sourceFoundations` to the `build-all` script
- Split the build command in the foundations package.json into separate clean and build steps


SiAdcock (approved):
Make change easy ✅

PR #997, opened 2021-09-30
Split workflow into separate steps

jamie-lynch:

This PR splits some of the calls in the ci workflow into separate steps so that they can run in parallel. This is done ahead of https://github.com/guardian/source/pull/985 which adds another long running step.

## What does this change?

-   Split the `build` step into separate `build` and `validate` steps


jamie-lynch:
As part of this, we should update the required checks to include the new `CI / validate` check.

SiAdcock (approved):
This looks great!

PR #998, opened 2021-09-30
Spacing in input fields

coldlink:

Screenshot 2021-09-28 at 10 37 29 (1)

To update the spacing in input field components (`TextInput`/`TextArea`) between the input element and the labels to match closer to design requirements.

## What does this change?

-   Adding a 2px margin-top to the supporting text in the label component, to space it from the main label text
-   Updates the TextInput component to apply the spacings between the label component, inline messages, and input field.
-   Updates the TextArea component to apply the spacings between the label component, inline messages, and input field.
- A margin wasn't added when an inline error/success was visible, as there is already quite a bit of space when it's visible.

## Screenshots

chrome_aTPt7lXXLp

chrome_FQTfyF876v

chrome_JpXUhHxcCH

### Accessibility

### Responsiveness

### Documentation

-   [x] Examples in Storybook

### DIFF
--- a/src/core/components/text-area/styles.ts
+++ b/src/core/components/text-area/styles.ts
@@ -8,6 +8,7 @@ import { resets } from '@guardian/src-foundations/utils';
 export const errorInput = css`
 	border: 4px solid ${border.error};
 	color: ${text.error};
+	margin-top: 0;
 `;
 
 export const textArea = css`
@@ -39,6 +40,18 @@ export const textArea = css`
 			${errorInput}
 		}
 	}
+`;
+
+export const labelMargin = css`
+	margin-top: ${space[1]}px;
+`;
+
+export const supportingTextMargin = css`
+	margin-top: 6px;
+`;
+
+export const inlineMessageMargin = css`
+	margin-top: 2px;
 `;
 
 export const widthFluid = css`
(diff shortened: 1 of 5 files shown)


src/core/components/text-area/styles.ts
```diff
@@ -8,6 +8,7 @@ import { resets } from '@guardian/src-foundations/utils';
 export const errorInput = css`
 	border: 4px solid ${border.error};
 	color: ${text.error};
+	margin-top: 0;
```
SiAdcock, line 11: I'm assuming this is necessary because of the order in which [styles are applied](https://github.com/guardian/source/blob/64bfef4f3b41c4e2177179a6d50373f5ee598058/src/core/components/text-area/TextArea.tsx#L109-L110) to the `textarea` element. If we move the `errorInput` style above the `supportingTextMargin` / `labelMargin` style, does this negate the need for the `margin-top` here?

coldlink: Well, it's actually because there's already a lot of whitespace in the `errorInput` and `successInput`, so we discussed that for now a margin should not be applied if the error or success is visible as it already looks like there is a margin there.

chrome_4pcYtnaWTG

Which means that we have to explicitly negate the margins that we applied using the `supportingTextMargin` / `labelMargin`.

SiAdcock: I see! You could say it's a CSS override to avoid a nested ternary in the JSX... caught between a rock and a hard place 😅  I think your way is best 👍

SiAdcock (approved):
Thanks so much for all your hard work @coldlink! 💎

PR #999, opened 2021-09-30
Export all component prop types

jamie-lynch:

This change updates all the index files to export the `ComponentProps` types for each component, rather than just the subset that we were previously exporting.

## What does this change?

-   Export `ComponentProps` types (as types) for all components
  -  The only exception is `StackProps` which, as it was the only one in that file that wasn't exported already, I've exported not as a type

### DIFF
--- a/src/core/components/checkbox/index.tsx
+++ b/src/core/components/checkbox/index.tsx
@@ -4,4 +4,6 @@ export {
 } from '@guardian/src-foundations/themes';
 
 export { CheckboxGroup } from './CheckboxGroup';
+export type { CheckboxGroupProps } from './CheckboxGroup';
 export { Checkbox } from './Checkbox';
+export type { CheckboxProps } from './Checkbox';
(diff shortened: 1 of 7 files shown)


sndrs:
this was a feature [added in TypeScript 3.8](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export) – do we need to update packages to spec that as a peer dep?

that feels weird since you don't _need_ typescript to use this...

jamie-lynch:
> this was a feature [added in TypeScript 3.8](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export) – do we need to update packages to spec that as a peer dep?
> 
> that feels weird since you don't _need_ typescript to use this...

Hmm... good point. It does feel weird that you'd get a warning on a js project to say that you need ts 😕  But also not great that you wouldn't get a warning if you were using ts < 3.8 and trying to import the types. 

We are already exporting several things as types so maybe if there's no obvious solution we can add a card to resolve this issue but process with this PR for now?

sndrs:
ah ok - yeah if it's already happening then let's not worry about it

sndrs (approved):

PR #1000, opened 2021-10-01
Add ErrorSummary and SuccessSummary components to the kitchen

coldlink:
## What does this change?

Adds the Summary components (`ErrorSummary` and `SuccessSummary`) to the kitchen. Some information about the Summary component and how it should be used is already on the [source design system website](http://theguardian.design/2a1e5182b/p/108ed3-user-feedback/b/3803b4/t/08c895).

These components can display a message with optional context information.

## Screenshots

**ErrorSummary**
chrome_U6tqTxuDdj

**SuccessSummary**
chrome_jYgwPRKwWZ

### DIFF
--- a/packages/@guardian/source-react-components-development-kitchen/components/summary/ErrorSummary.stories.tsx
+++ b/packages/@guardian/source-react-components-development-kitchen/components/summary/ErrorSummary.stories.tsx
@@ -1,0 +1,40 @@
+import { ErrorSummary, ErrorSummaryProps } from './ErrorSummary';
+import type { Story } from '../../../../../lib/@types/storybook-emotion-10-fixes';
+import {
+	asPlayground,
+	asChromaticStory,
+} from '../../../../../lib/story-intents';
+
+export default {
+	title: 'Kitchen/source-react-components-development-kitchen/Error Summary',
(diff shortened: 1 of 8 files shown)


packages/@guardian/source-react-components-development-kitchen/components/summary/ErrorSummary.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+import {
+	asPlayground,
+	asChromaticStory,
+} from '../../../../../lib/story-intents';
+
+export default {
+	title: 'Kitchen/source-react-components-development-kitchen/Error Summary',
+	component: ErrorSummary,
+	args: {
+		error: 'There has been a problem',
```
SiAdcock, line 12: Minor thing, but if you include the `context` prop here with an empty string:

```suggestion
		error: 'There has been a problem',
		context: '',
```

you get text input field for this prop in the Docs page, rather than a button that reveals a text input field

**Before**

Screenshot 2021-10-01 at 13 45 39

**After**

Screenshot 2021-10-01 at 13 45 24

SiAdcock (approved):
Great start, thanks @coldlink ❤️

PR #1001, opened 2021-10-01
Bump @guardian/source-react-components-development-kitchen to v0.0.3

SiAdcock:

Since #1000 was merged, we should publish a new version to allow teams to consume it

## What does this change?

- Bumps @guardian/source-react-components-development-kitchen to v0.0.3


jamie-lynch (approved):
🚢

PR #1003, opened 2021-10-01
Export types as types

jamie-lynch:

This change updates all type exports to make sure they follow the pattern  `export type {Type} from './file'`.

This change has been prompted by the analysis script which finds component usage as it means that we can easily distinguish between components and types.

## What does this change?

-   Ensure all types are actually exported as types


SiAdcock (approved):
Looks great, thanks @jamie-lynch

PR #1005, opened 2021-10-01
Run CI workflow on `pull_request` event

jamie-lynch:

This change updates the trigger for the CI event in order to get `coveralls` to comment on the PR with the coverage change.

## What does this change

-   Change the trigger in the `ci` workflow to `pull_request`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: CI
 on:
-  push:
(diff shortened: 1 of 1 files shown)


jamie-lynch:
I think we might have to merge this for it to take effect as when you change actions with the `pull_request` event it only runs what's in main.

sndrs (approved):

PR #1006, opened 2021-10-04
Add `fit` and `none` as size and spaceAbove options respectively.

ob6160:

In Gateway, we needed a way to have more granular control over the spacing and width of the Divider component.

## What does this change?
This PR proposes two new options added to the `size` and `spaceAbove` props.

`size='fit'` allows the divider to fit the container exactly,
without overlap on the sides.

`spaceAbove='none'` allows the divider to have no margin above so that
the end user can have complete control over its spacing.

## Screenshots

**Before**
Divider projecting past the container.

Screenshot 2021-10-04 at 12 12 58

**After**
Divider fitting within the container.

Screenshot 2021-10-04 at 12 12 24

### Responsiveness

### Documentation

-   [x] Examples in Storybook

### DIFF
--- a/packages/@guardian/source-react-components-development-kitchen/components/divider/Divider.tsx
+++ b/packages/@guardian/source-react-components-development-kitchen/components/divider/Divider.tsx
@@ -1,18 +1,19 @@
 import { css, SerializedStyles } from '@emotion/react';
-
 import { border, text } from '@guardian/src-foundations/palette';
 import { space } from '@guardian/src-foundations';
 import { textSans } from '@guardian/src-foundations/typography';
-
 import { partialStyles, fullStyles } from './styles';
+
+type Sizes = 'full' | 'partial' | 'fit';
+type Spaces = 'tight' | 'loose' | 'none';
 export type DividerProps = {
-	size?: 'full' | 'partial';
-	spaceAbove?: 'tight' | 'loose';
+	size?: Sizes;
+	spaceAbove?: Spaces;
 	displayText?: string;
 };
 
 const decideSpace = (
-	spaceAbove: 'tight' | 'loose',
+	spaceAbove: Spaces,
 	displayText?: string,
 ): SerializedStyles => {
 	switch (spaceAbove) {
@@ -32,6 +33,51 @@ const decideSpace = (
 				: css`
 						margin-top: ${space[12]}px;
 				  `;
+		case 'none':
+			return css`
+				margin-top: 0;
+			`;
+	}
+};
+
+const decideSize = (size: Sizes, displayText?: string) => {
+	switch (size) {
+		case 'fit':
+			return displayText
+				? css`
+						:before {
+							margin-left: 0;
+						}
+
+						:after {
+							margin-right: 0;
+						}
+				  `
+				: null;
+		case 'full':
+			return displayText
+				? css`
+						:before {
+							margin-left: -10px;
+						}
+
+						:after {
+							margin-right: -10px;
+						}
+				  `
+				: fullStyles;
+		case 'partial':
+			return displayText
+				? css`
+						:before {
+							margin-left: 30%;
+						}
+
+						:after {
+							margin-right: 30%;
+						}
+				  `
+				: partialStyles;
 	}
 };
 
@@ -60,18 +106,13 @@ export const Divider = ({
 							margin: auto;
 						}
 						:before {
-							margin-left: ${size === 'partial'
-								? '30%'
-								: '-10px'};
 							margin-right: 10px;
 						}
 						:after {
-							margin-right: ${size === 'partial'
-								? '30%'
-								: '-10px'};
 							margin-left: 10px;
 						}
 					`,
+					decideSize(size, displayText),
 					decideSpace(spaceAbove, displayText),
 				]}
 			>
@@ -89,7 +130,7 @@ export const Divider = ({
 					margin-bottom: ${space[1]}px;
 					background-color: ${border.secondary};
 				`,
-				size === 'partial' ? partialStyles : fullStyles,
+				decideSize(size),
 				decideSpace(spaceAbove),
 			]}
 		/>
--- a/packages/@guardian/source-react-components-development-kitchen/components/divider/Divider.stories.tsx
+++ b/packages/@guardian/source-react-components-development-kitchen/components/divider/Divider.stories.tsx
@@ -51,6 +51,14 @@ asChromaticStory(TightDivider);
 
 // *****************************************************************************
 
+export const NoneDivider = Template.bind({});
+NoneDivider.args = {
+	spaceAbove: 'none',
+};
+asChromaticStory(NoneDivider);
+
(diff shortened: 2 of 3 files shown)


jamie-lynch:
Sorry, I've managed to break the CI checks which is why they haven't run🤦  Will get this fixed ASAP and rebase this PR

packages/@guardian/source-react-components-development-kitchen/components/divider/Divider.tsx
```diff
@@ -1,18 +1,19 @@
 import { css, SerializedStyles } from '@emotion/react';
-
 import { border, text } from '@guardian/src-foundations/palette';
 import { space } from '@guardian/src-foundations';
 import { textSans } from '@guardian/src-foundations/typography';
-
 import { partialStyles, fullStyles } from './styles';
+
+type Sizes = 'full' | 'partial' | 'fit';
+type Spaces = 'tight' | 'loose' | 'none';
```
oliverlloyd, line 8: is the convention to not use plural here? I feel like just `Size` and `Space` is more consistent with other places?

packages/@guardian/source-react-components-development-kitchen/components/divider/Divider.tsx
```diff
@@ -32,6 +33,51 @@ const decideSpace = (
...
 						margin-top: ${space[12]}px;
 				  `;
+		case 'none':
+			return css`
+				margin-top: 0;
+			`;
+	}
+};
+
+const decideSize = (size: Sizes, displayText?: string) => {
```
oliverlloyd, line 43: Is this exposing a weakness with the original api? Passing in `displayText` and using that to decide size seems.... wrong? Not really sure what the right api would be to be honest and I appreciate this is out of scope but might be worth some thought

packages/@guardian/source-react-components-development-kitchen/components/divider/Divider.tsx
```diff
@@ -1,18 +1,19 @@
 import { css, SerializedStyles } from '@emotion/react';
-
 import { border, text } from '@guardian/src-foundations/palette';
 import { space } from '@guardian/src-foundations';
 import { textSans } from '@guardian/src-foundations/typography';
-
 import { partialStyles, fullStyles } from './styles';
+
+type Sizes = 'full' | 'partial' | 'fit';
```
oliverlloyd, line 7: `full` means stretch to the sides in an article (it touches the edges)
`partial` means stretch left but limit the length
`fit` means... don't stretch to touch the sides...

This api has grown over time, which is nice in a way - we're growing! - but it also really _feels_ like it did

Is there a case for making a some breaking changes (while we're still at such an early stage of adoption) to build a more valid api covering all cases?

packages/@guardian/source-react-components-development-kitchen/components/divider/Divider.tsx
oliverlloyd: If we need to pass in `spaceAbove: 'none';` does that not mean that the values we're using - _somewhere_ - are wrong?

Ideally, the goal should be consistent spacing above the divider. If there's a unique case where, for a specific reason, we need a different value we could use `cssOverrides` to break out of the common values?

oliverlloyd (commented):
+1 for breaking out the logic to `decideThing` 👏

ob6160:
@oliverlloyd, I'm interested in removing this line https://github.com/guardian/source/pull/1006/files#diff-473a662da5a7e21da60bc443cbc7e4808b92879691d65180d4ecd8df3c0f669fR98, because it means that we have to do some extra margin fixes when using this component to make it align with the Source spacing guidelines.

It'd be great to know what the knock-on effects of removing this would be and the intent for it originally.

I was also thinking - I like the idea of keeping the margin spacing of components "pure". By that I mean that the component shouldn't decide how it is spaced relative to other components on the page, but instead the person implementing it should have control. I like this because each time a component is used a subtly different set of margins is usually needed. 

Interested to hear your thoughts on that :)

packages/@guardian/source-react-components-development-kitchen/components/divider/Divider.tsx
ob6160: yeah agreed! I'd be interested in refactoring this component, as alluded to in the comment below (https://github.com/guardian/source/pull/1006#issuecomment-933388759).

What do you think about me putting through a followup PR to do that (and your other changes), as it's a bit beyond the scope of this one

packages/@guardian/source-react-components-development-kitchen/components/divider/Divider.tsx
ob6160: Agreed re: plural naming

Having the ability to turn off all margin around the divider to leave spacing up to the end user strikes me as a good potential solution, as there is a negative margin-bottom too which we'd need to add a `spaceBelow` prop for, which seems messy

packages/@guardian/source-react-components-development-kitchen/components/divider/Divider.tsx
ob6160: I'd be interested in exploring the possibility of shipping the Divider with no margin at all.

Instead we could provide some exported css blocks like `dividerSpacingTight` (completely contrived name) which the end user could then apply at will to customise their Divider

packages/@guardian/source-react-components-development-kitchen/components/divider/Divider.tsx
oliverlloyd: Can we do it the other way around? Do the refactor first and _then_ merge this PR? Like this we avoid the state where we avoid the need for a pinky promise to do things later (which may or may not happen regardless of all our best intentions)

packages/@guardian/source-react-components-development-kitchen/components/divider/Divider.tsx
```diff
@@ -1,18 +1,19 @@
 import { css, SerializedStyles } from '@emotion/react';
-
```
oliverlloyd, line 2: I think this spacing between imports is a convention?

packages/@guardian/source-react-components-development-kitchen/components/divider/Divider.tsx
oliverlloyd: I think we discussed this when first creating this component. It's a tricky one. On one hand, the division of space is intrinsic to what a 'divider' is so the size of the margin is part of its style. But in other cases we don't set these vertical margins, expecting the platforms to manage this themselves.

My feeling here is that platform specific margins are different to this particular margin. On Dotcom there is a uniform 12px space above each element on an article which is managed by a wrapper component `Figure`. But a divider needs more space and it would feel wrong to try and build that logic into `Figure` to do this.

I wonder if the real question is: Why do you need a different amount of space here? Why isn't 'tight' or 'loose' right for your use case? Could we adjust the designs for where this component will be used to align with the common standard?

PR #1010, opened 2021-10-04
Add `build:all` step to CI

jamie-lynch:

This PR adds the `build:all` step to CI. This is done to support the work done in https://github.com/guardian/source/pull/985 as the packages must be built before being validated.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,3 +28,12 @@ jobs:
           cache: 'yarn'
       - run: yarn install --frozen-lockfile
       - run: yarn build:ci
+  build-packages:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.4
+      - uses: guardian/actions-setup-node@v2.4.1
+        with:
+          cache: 'yarn'
+      - run: yarn install --frozen-lockfile
+      - run: yarn build:all


sndrs:
how much slower is it? could this maybe just be how we build the old stuff?

i'm nervous about having so many script targets. i know we plan to remove a lot of them, but while they're still there they make life harder while we're working on other quite complex refactoring...

jamie-lynch:
> how much slower is it?

About 20s (~110s vs ~130s). Maybe a small enough difference that we just have the one build script that runs in series, rather than maintaining both? This would be my preference.

> i'm nervous about having so many script targets. i know we plan to remove a lot of them, but while they're still there they make life harder while we're working on other quite complex refactoring...

Agreed that adding more script files and package.json commands is making life more complicated when there's already a lot going on however in this case the extra files relate only to the series vs parallel argument.

>  could this maybe just be how we build the old stuff?

In general, I also agree that it would be good to be clear about how the old and new ways are built. Two obvious choices are:

1) Use the `build-all.ts` script to build both the old and new ways of doing things

    This means adding any new packages to the `path.ts` script and then adding them to the `prioritisedPackages` in `build-all.ts`. Once we've removed the old way of doing things, we can then revisit entirely how we build things.

2) Use `build-all.ts` script to build the old way of doing things and add a different way to build the new things. Then have commands in the package.json that combine the two.

    This would probably look something like:

    ```json
    "build:all": "yarn build:old && yarn build:new",
    "build:old": "ts-node ./scripts/build-all",
    "build:new": "yarn kitchen build && yarn foundations build"
    ```

I don't have a strong opinion on which of those two options we go with but, given that option 1 is already in place, I'd prefer to keep the scope of this PR to getting the existing build process to run in CI so that we can unblock getting the shakedown tests running in CI. Happy to open, pair on or review a separate PR that moves from option 1 to option 2.

sndrs:
yeah i like just making the main script slower but series so CI is happy, 👍

sndrs (approved):

scripts/build-all.ts
```diff
@@ -3,7 +3,8 @@
 import execa, { ExecaReturnValue } from 'execa';
 import { paths, getComponentPaths } from './paths';
 
-const build = (dir: string) => {
+const build = (dir?: string) => {
+	if (!dir) return;
 	console.log(`\nBuilding ${require(`${dir}/package.json`).name}`);
 
 	return execa('yarn', ['--cwd', dir, 'run', 'build'], {
```
sndrs, line 10: [`execa.sync`](https://github.com/sindresorhus/execa#execasyncfile-arguments-options) might help here?

then we should be able to do somtehing like `[...prioritisedPackages, ...packages].forEach(build)` i think?

jamie-lynch: Ooh yeah, that'd be nicer. 

I added the `if (!dir) return;` because I was getting a type error about dir being `String | undefined` when I ran it but not in the editor and this fixed it. Not ideal but hopefully this script will be going soon anyway.

sndrs (approved):

scripts/build-all.ts
sndrs: 👍

PR #1012, opened 2021-10-04
Allow passing of `cssOverrides` prop to the Divider

ob6160:

In Gateway, we needed a way to have more granular control over the spacing and width of the Divider component. After an initial PR (#1006) we decided to allow `cssOverrides` to be passed in, instead of further extending this component to fit more use-cases.

## What does this change?

-   Adds a `cssOverrides` prop to the container element of the Divider.
-   Adds an example story demonstrating its use to fit the Divider to its parent container.
-   Refactors the component a little, moving space and size into their own types.

### Documentation

-   [x] Examples in Storybook


packages/@guardian/source-react-components-development-kitchen/components/divider/Divider.tsx
```diff
@@ -35,10 +37,40 @@ const decideSpace = (
 	}
 };
 
+const decideSize = (size: Size, displayText?: string) => {
```
oliverlloyd, line 40: To be honest, I think the previous method was simpler to follow. By adding this function you're coupling `decideText` in with `size` in a way that wasn't there before.

Can we remove this change from this PR for now and focus in on the `cssOverrides` change - that way we can merge it through quickly and then have some breathing space to take stock around the larger api discussion.

packages/@guardian/source-react-components-development-kitchen/components/divider/Divider.stories.tsx
```diff
@@ -64,3 +65,13 @@ TextDivider.args = {
...
+
+// *****************************************************************************
+
+export const CssOverridesDivider = Template.bind({});
+CssOverridesDivider.args = {
+	cssOverrides: css`
+		background: red;
+	`,
+};
+asChromaticStory(CssOverridesDivider);
```
SiAdcock, line 77: Up to this point, we use Chromatic stories to demonstrate supported variants of each component, rather than a TDD-for-UI. As such, I don't think we need a specific Chromatic story for `cssOverrides`.

packages/@guardian/source-react-components-development-kitchen/components/divider/README.md
```diff
@@ -54,3 +54,9 @@ How much space is given above the divider
 **`string`**
 
 This text is centered inside the dividing line
+
+### `cssOverrides`
+
+**`SerializedStyles | SerializedStyles[] | undefined`**
+
+An escape hatch, used to override default styles if needed.
```
SiAdcock, line 62: To avoid duplicating API docs, we've been following the pattern of linking to Storybook:

```md
## Use

### API

See [storybook](https://guardian.github.io/source/?path=/docs/kitchen-source-react-components-development-kitchen-divider--playground)

### How to use

For context and visual guides relating to usage see the [Source Design System website](https://theguardian.design).
```

Could we remove all props documentation from this README?

SiAdcock (dismissed):
A couple of repo admin points but nothing blocking. All hail `cssOverrides`! 🙈

packages/@guardian/source-react-components-development-kitchen/components/divider/README.md
ob6160: sounds good - will do

packages/@guardian/source-react-components-development-kitchen/components/divider/Divider.stories.tsx
ob6160: good point, removal incoming!

SiAdcock (approved):
Lovely stuff, thanks @ob6160 💎

PR #1014, opened 2021-10-05
Use `np` to release `development-kitchen`

sndrs:

we know the current way of working with the packages is complex. 

this is part 1 of a plan to leave the old world in tact (till it's time for it to go) but not carry the same complexity over the new one.

## What does this change?

-   add `yarn release` script to kitchen (that uses `np`)
-   remove the old `release:*:kitchen` and `publish:release:kitchen` scripts from root `package.json`

You can then release `@guardian/source-react-components-development-kitchen` by running `yarn kitchen release` from root (or `yarn release` from the package root).


packages/@guardian/source-react-components-development-kitchen/package.json
```diff
@@ -33,6 +27,7 @@
   "devDependencies": {
     "@guardian/libs": "^3.2.1",
     "mini-svg-data-uri": "^1.3.3",
+    "np": "^7.5.0",
```
jamie-lynch, line 30: Assuming that we're going to do something similar with the other packages in time, would it be better to move this dependency to the root `package.json` or shall we do that if/when we add it to the next package?

jamie-lynch (approved):

packages/@guardian/source-react-components-development-kitchen/package.json
sndrs: i think because `np` _is_ a dev dep of this specific package we should include it here. it may be that we end up with lots of versions across the packages, but this means we dont have to worry about the version that's installed being in sync, deleting it if we move other packages to a different process etc.

jamie-lynch: Fair enough. 👍

PR #1015, opened 2021-10-05
bump yarn version

sndrs:
bumps yarn to latest v1


.yarnrc
```diff
@@ -0,0 +1,8 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+email a@snd.rs
+lastUpdateCheck 1633433443971
+username sndrs
```
SiAdcock, line 7: Heh. Does this file need your email and GitHub username?

sndrs: oh wow how wierd! i just ran `yarn policies set-version` and committed. i don't think that's important info no :)

sndrs: but then... `# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.`

will it just keep appearing with the last editor's info?

sndrs: that email is wrong too :(

sndrs: i had some yarn config setup for some reason, removed and reran

jamie-lynch (approved):

PR #1016, opened 2021-10-05
Update `Logo` component to use `a` tag, and `cssOverrides`

coldlink:
## What does this change?
- Use an `a` tag for the link
    - This is because the `Link` component has additional stylings which can cause the logo to shift, e.g. `line-height`
- Add `cssOverrides`
    - Allows for customisation of the logo by overriding or adding extra css from the client application


packages/@guardian/source-react-components-development-kitchen/components/logo/Logo.tsx
```diff
@@ -117,25 +117,26 @@ const DecideLogo = ({ logoType }: { logoType: LogoType }) => {
 	}
 };
 
+const logoStyles = css`
+	/* Adding block display here so that the focus halo correctly covers the content */
+	display: block;
+	color: ${brandText.primary};
+	:hover {
+		color: ${brandText.primary};
+	}
```
mxdvl, line 126: If there’s no text (it’s `visuallyHidden`), is it worth setting a colour at all?

coldlink: Good spot! I just copied the styling that was already there. But makes sense to remove it.

SiAdcock (approved):
Looooooove it ❤️

PR #1018, opened 2021-10-05
Remove context from `get-usage` script package name and make private

jamie-lynch:

This PR updates the name of the `get-usage` package to remove the `@guardian/source-` prefix as it is not needed for internal packages. It also adds the `"private": true` flag to the `package.json` to prevent accidental publishing.


SiAdcock (approved):
👍

sndrs (approved):
good spot

PR #1020, opened 2021-10-05
Set up nextgen scripts

sndrs:

delineate the current npm-scripts from first pass of nextgen ones (continuation of #1014)

## What does this change?

-   standardises names
	- `"build-storybook"` → `"build:storybook"`
- makes names more explicit
	- `"build:ci"`  → `"build:storybook:ci"`
-   renames the `*all` npm-scripts and the script files themselves to `*src`
	-  e.g. `"build:all"` → `"build:src"`
	- e.g. `scripts/tag-beta-all.ts` → `scripts/tag-beta-src.ts`
- removes the nextgen packages from the current scripts
- removes the nextgen alias npm-scripts e.g. `"foundations"`
	- they are ingenious, but then I accidentally ran `yarn kitchen` and unexpected things started happening 
		- `src-*` packages started building 🤷
		- then i felt that the clearer it is what scripts will do, they less they interrupt your flow
	- removing them forces us to explicitly determine the dev API of the repo
		- this feels like it has the same benefits as only exporting components from the top level
		- there's no magic, and you don't need to know how packages are set up to run their tasks from root (if you're meant to)
- adds some universal unsuffixed script names (`build`, `clean` etc) that do these things for all possible moving parts (though in fairness not storybook, only packages really i guess)

what you reckon? does this make sense?


package.json
```diff
@@ -11,46 +11,50 @@
...
+    "release:prerelease-bump:src": "yarn validate && yarn verbump:prerelease:src && yarn publish:release",
+    "release:prerelease:src": "yarn validate && yarn publish:release",
     "storybook": "start-storybook -p 6006",
-    "tag:beta:all": "ts-node ./scripts/tag-beta-all",
+    "tag:beta:src": "ts-node ./scripts/tag-beta-src",
     "test": "jest",
     "test:coverage": "jest --coverage",
     "tsc": "tsc",
     "validate": "yarn tsc && yarn lint && yarn test",
     "validate-dist": "yarn build:all && yarn workspace validate-dist validate",
```
jamie-lynch, line 43: This should just be `yarn build` now, right?

```suggestion
    "validate-dist": "yarn build && yarn workspace validate-dist validate",
```

jamie-lynch:
Looks good to me. I just left one comment about the validate-dist command which I think will need updating.

package.json
sndrs: Oh yes good spot

jamie-lynch (approved):

PR #1028, opened 2021-10-06
Refactor label stories

jamie-lynch:

Bring the `Label` and `Legend` stories in line with other stories. This will (hopefully) also make the migration work to `@guardian/source-react-components` easier.

## What does this change?

-   Update `Label` and `Legend` stories to follow current recommendations
-   Remove `ts-dedent` as it's no longer needed


jamie-lynch:
The Label stories don't render an input for me in the FireFox snapshots in Chromatic 😕 
This link looks okay in FF: https://guardian.github.io/source/?path=/story/source-src-label-label--demo

jamie-lynch:
image

Not sure why this has happened...

PR #1037, opened 2021-10-07
Add jsx support to `validate-dist`

sndrs:

i'm not sure why it hasn't come up before, but #1033 is breaking because `validate-dist` wasn't set up to handle jsx. this adds support, adopts ES2020 and sorts the config file


sndrs:
this is the wrong fix for the wrong problems ❌

PR #1044, opened 2021-10-08
Create a `@guardian/eslint-config-source` package

jamie-lynch:

This PR introduces a new `@guardian/eslint-plugin-source` package which contains rule(s) that are specific to Source. The main motivation for introducing this package is to aid the migration from v3 -> v4 which includes updating the location which most Source exports are imported from.

## What does this change?

- Add the skeleton for a custom eslint plugin
- Create the `validImportPaths` rule
- Create tests for the `validImportPaths` rule
- Add an integration test package

### ✅  Should it be `@guardian/eslint-plugin-source` (which lines up with [configs](https://github.com/guardian/configs)) or `@guardian/source-eslint-plugin` (which lines up with the other Source packages)?

It turns out that it has to be `@guardian/eslint-plugin-source` as that's a convention of eslint required to get the configuration to work.

### ✅  What should we do about different export types?

1) ✅  `import themes from '@guardian/src-foundations/themes`
    
    Nothing has a default export so this shouldn't exist anywhere

 1) ✅ `import * as themes from '@guardian/src-foundations/theme'`

    This is currently being updated to `import * as themes from '@guardian/source-react-components'` which is going to mean exporting significantly more items that previously.

    This is something that we want to recommend against doing therefore we wont autofix any imports that do this. In addition, a subsequent PR will add a further rule to mark these import styles as errors in their own right.

1) ✅  `export * from '@guardian/src-button`

    This will currently be updated to `export * from '@guardian/source-react-components`.

    This is something that we want to recommend against doing therefore we wont autofix any exports that do this.  In addition, a subsequent PR will add a further rule to mark these export styles as errors in their own right.

## How it works

We're using ESLint to check that imports come from the right place, and to fix them if they don't. This is implemented by creating a custom plugin which includes a [custom rule](https://eslint.org/docs/developer-guide/working-with-rules). 

For most imports, that means swapping the source to either `@guardian/source-react-components` or `@guardian/source-foundations` as required. 

Creating a plugin allows us to define custom rules, rather than a configuration which just let's us template a base configuration for others to extends. We also provide a "recommended" configuration which can be extended so that including this plugin is as simple as `"extends": ["plugin:@guardian/source/recommended"]`.

#### Collapsing multiple imports

In cases where multiple imports now have the same source, we want to collapse this into one statement. This is done by including the `import/no-duplicates` rule from [eslint-plugin-import](https://www.npmjs.com/package/eslint-plugin-import) package in the recommended configuration.

#### Imports that have changes name

Some values have changed name (e.g. `headline` to `headlineObjectStyles`). These values are updated inline.

#### Imports that no longer exist

For imports that no longer exist, no fix can be applied. The standard message is modified to explain that the imports are no longer available. Where an import statement contains both imports that have moved and imports that no longer exist, the imports that have moved are added in a new import statement and removed from the original one. 

### Integration Test

The integration test package installs the plugin locally and runs in on a test file to ensure that everything works as expected. Currently this process is manual (run `yarn workspace eslint-integration-test lint` and view the output). In future, it would be good to automate a number of test cases to test this works as expected.

## What's left?

- [x] Make the initial case (importing from `@guardian/src-label`) fixable
- [x] Extend the rule to cover all imports (e.g. `@guardian/src-foundations`)
- [x] Update error message
- [x] Setup build process
- [x] Ensure multiple imports from the same location get flattened
- [x] Account for `@guardian/src-helpers` 
- [x] Make the rule account for named exports
- [x] Make the rule account for `export * from '@guardian/src-label'` statements
- [x] Make the rule account for * imports
- [x] Add an integration test
- [x] Setup release process
- [x] Test in DCR


SiAdcock:
A thought that occurred. Since this plugin is specifically geared to the v3 - v4 migration, should this fact be reflected in the naming, or perhaps the versioning? It’s likely we’ll need further code modifications in the future and we probably need to think of a way of distinguishing them

PR #1062, opened 2021-10-13
Add a new rule to disallow `*` imports and exports

jamie-lynch:

Following https://github.com/guardian/source/pull/1044, this PR introduces a new rule which marks import and exporting everything from a source package as an error.

For example:

- `import * as themes from '@guardian/src-foundations/themes'`
- `export * from '@guardian/src-foundations/themes'`

## Why?

This has come to light as part of the migration to fewer, larger packages as we cannot autofix import/export statements that reference everything in a package. In general we don't recommend this approach as:

* It's harder to manage as the package changes
* It's harder to see what a file actually uses from an import
* It's harder for bundlers to deal with
* In the new world, importing everything will be a lot of things

### DIFF
--- a/packages/@guardian/eslint-plugin-source/src/index.ts
+++ b/packages/@guardian/eslint-plugin-source/src/index.ts
@@ -1,7 +1,9 @@
+import { noStarImportsOrExports } from './rules/no-star-imports-or-exports';
 import { validImportPaths } from './rules/valid-import-path';
 
 export const rules = {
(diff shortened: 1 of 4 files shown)


jamie-lynch:
> should this be a draft for now? looks like a lot of stuff from other PRs in there?

Hmm yeah... I thought it would be okay because it's PRed against `source-eslint` but maybe not. I'll mark it as draft for now anyway until the `source-eslint` PR is merged.

PR #1088, opened 2021-10-20
remove unneeded `rollup-plugin-dts`

sndrs:
missed in #1087


jamie-lynch:
It looks like we need this for source-react-components (https://github.com/guardian/source/pull/1160) so maybe we should keep this everywhere for now?

jamie-lynch:
> It looks like we need this for source-react-components (#1160) so maybe we should keep this everywhere for now?

I'm going to close this for now based on #1160. Let's revisit it once we've moved the code across to v4 and there's less magic in the build process.

PR #1089, opened 2021-10-21
Split the ESLint plugin into two packages

jamie-lynch:

This PR updates the eslint plugin provided to be two separate packages that refer to the `@guardian/source-foundations` and `@guardian/source-react-components` packages. This allows us to keep the plugin versions in sync with the packages that they refer to, and to have those packages as peer dependencies so that users will be warned if they have incompatible versions installed.

An alternative approach to this problem is detailed in: #1090

## What does this change?

-   Move the eslint rules into `lib`
-   Add separate `valid-foundations-import-name` and `valid-components-import-name` rules to the `valid-import-name` file
- Split the `@guardian/eslint-plugin-source` package into `@guardian/eslint-plugin-source-foundations` and `@guardian/eslint-plugin-source-react-components`
- Update the recommended configs accordingly
- Build both packages with rollup
- Update the integration tests
- Add `@guardian/source-foundations` and `@guardian/source-react-components` as peer dependencies of the `@guardian/eslint-plugin-source-foundations` and `@guardian/eslint-plugin-source-react-components` respectively

## Possible Changes

* `@guardian/eslint-plugin-source-react-components` could extend `@guardian/eslint-plugin-source-foundations` so that you weren't required to extend from both

## Downsides

If both `@guardian/source-foundations` and `@guardian/source-react-components` are installed the an extra rule is present that runs on every import statement, compared to having a single package.

### DIFF
--- a/lib/eslint/rules/valid-import-path.test.ts
+++ b/lib/eslint/rules/valid-import-path.test.ts
@@ -1,5 +1,8 @@
 import { RuleTester } from 'eslint';
-import { validImportPaths } from './valid-import-path';
+import {
+	validFoundationsImportPath,
+	validComponentsImportPath,
+} from './valid-import-path';
 
 const ruleTester = new RuleTester({
 	parser: require.resolve('@typescript-eslint/parser'),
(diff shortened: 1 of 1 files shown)


jamie-lynch:
#1091 should solve the test failures

PR #1092, opened 2021-10-22
Split ESLint plugin into foundations and components (Attempt 2)

jamie-lynch:

This PR extends the work done in https://github.com/guardian/source/pull/1089 so that consumers can simply use `@guardian/eslint-plugin-source-react-components` to get linting for both react components and foundations.

> These sections are copied from #1089 

> This PR updates the eslint plugin provided to be two separate packages that refer to the @guardian/source-foundations and @guardian/source-react-components packages. This allows us to keep the plugin versions in sync with the packages that they refer to, and to have those packages as peer dependencies so that users will be warned if they have incompatible versions installed.

## What does this change?

 > -   Move the eslint rules into `lib`
> -   Add separate `valid-foundations-import-name` and `valid-components-import-name` rules to the `valid-import-name` file
> - Split the `@guardian/eslint-plugin-source` package into `@guardian/eslint-plugin-source-foundations` and `@guardian/eslint-plugin-source-react-components`
> - Update the recommended configs accordingly
> - Build both packages with rollup
> - Update the integration tests
> - Add `@guardian/source-foundations` and `@guardian/source-react-components` as peer dependencies of the `@guardian/eslint-plugin-source-foundations` and `@guardian/eslint-plugin-source-react-components` respectively

* Add `@guardian/eslint-plugin-source-foundations` as a dependency of `@guardian/eslint-plugin-source-react-components`
* Update the recommended config exported from react components plugin to contain all rules
* Combine the foundations and components import rules so that import statements aren't linted twice
* Update the integration tests to test when linting foundations only too

## What's left to do?

- [x] Combine the foundations and components import rules so that import statements aren't linted twice
- [x] Update the integration tests to test when linting foundations only - https://github.com/guardian/source/pull/1093


SiAdcock (approved):
This is super awesome, thanks so much @jamie-lynch 💎 👌 ❤️

PR #1106, opened 2021-10-25
Forward ref to the underlying HTMLInputElement of the TextInput

ob6160:
Allow for a ref to be passed to the TextInput component so that a reference to the input element can be obtained.

## Reason for the change
A planned usage of the TextInput component needs to listen on the `blur` and `invalid` events for a specific field so we can run validation code based on the built in HTML input validation.

## Example of how it helps in our use case
Here we register two handlers, one to validate on the `blur` event, the other to act if an `invalid` event is fired on that specific field once validated.

carbon (3)


ob6160:
Aha! I hadn't heard of "Synthetic Events" as the name for the pattern `on<EventName>` before. TIL! I can totally use those for my use case. I'll close this PR and we can continue the discussion in guardian/csnx#261  as you suggested

PR #1118, opened 2021-10-29
Get Icon SVGs form Figma api

SiAdcock:

This change introduces a new script which queries the Figma API in order to get the SVG source of the icons provided by Source. This is the first part of a piece of work to automate the synchronisation of icons between Figma and the react component library.

## What does this change?

-   Add a new script to get icons from the Figma API and write to disk
-   Add the 24x24 icons to the repo


scripts/get-icons/index.ts
```diff
@@ -0,0 +1,103 @@
+import axios from 'axios';
+import { existsSync, mkdirSync, writeFileSync } from 'fs';
+
+const ICON_FILE = 'Ai7AELHC6KCz38qKZkvuHo';
+
+const ICON_FRAME = 'UI icons 24(w)x24(w)';
+
+const OUTPUT_DIR = './svgs';
```
SiAdcock, line 8:
```suggestion
const OUTPUT_DIR = '../../packages/@guardian/src-icons/svgs';
```

scripts/get-icons/index.ts
```diff
@@ -0,0 +1,103 @@
...
+	name: string;
+	node_id: string;
+}
+
+interface NodeWithUrl extends Node {
+	url: string;
+}
+
+const getUrlsForNodes = (nodes: Node[]): Promise<NodeWithUrl[]> => {
+	return new Promise((resolve) => {
```
SiAdcock, line 38: Just to help me understand, why do we need to wrap the `axios` response in a Promise? My brain is thinking it's already returning a `Promise` 🤔  😄

scripts/get-icons/index.ts
```diff
@@ -0,0 +1,103 @@
...
+	});
+};
+
+const getAndWriteSVGForNode = (node: NodeWithUrl) => {
+	return new Promise((resolve) => {
+		return axios
+			.get(node.url)
+			.then((res) => {
+				return resolve(
+					writeFileSync(`svgs/${node.name}.svg`, res.data),
```
SiAdcock, line 63:
```suggestion
					writeFileSync(`OUTPUT_DIR/${node.name}.svg`, res.data),
```

SiAdcock:
This is looking pretty awesome, @jamie-lynch! 🥇 

Shall we keep a list of what else we think this needs so we can pick up the last little bits later?

- [ ] Remove `height`, `width` and `fill` attributes from `<svg>`s
- [ ] Pass through [SVGOMG](https://jakearchibald.github.io/svgomg/)
- [ ] Add README
- [ ] Generate React component modules from `.svg` files

Anything else?

scripts/get-icons/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "get-icons",
+  "version": "1.0.0",
+  "private": true,
+  "license": "MIT",
```
SiAdcock, line 5:
```suggestion
  "license": "Apache-2.0",
```

scripts/get-icons/index.ts
jamie-lynch: We don't 👍 Good spot. I've removed it now and it all still works as expected.

jamie-lynch:
> Shall we keep a list of what else we think this needs so we can pick up the last little bits later?

Great shout. I've created a new card on the backlog to track this. 🎟️

scripts/get-icons/index.ts
```diff
@@ -0,0 +1,95 @@
+import axios from 'axios';
+import { existsSync, mkdirSync, writeFileSync } from 'fs';
+
+const ICON_FILE = 'Ai7AELHC6KCz38qKZkvuHo';
+
+const ICON_FRAME = 'UI icons 24(w)x24(w)';
+
+const OUTPUT_DIR = '../../packages/@guardian/src-icons/svg';
```
SiAdcock, line 8: Sorry, on thinking about it, would it be better to pluralise this directory name? We tend to pluralise things elsewhere in this repo (`components`, `foundations`, `icons`, `'libs`)

```suggestion
const OUTPUT_DIR = '../../packages/@guardian/src-icons/svgs';
```

SiAdcock:
I can't approve this PR because I created it! But feel free to self-approve as I think we're both very happy with it 😄 

👍 👍 👍 👍 👍 👍 👍

PR #1124, opened 2021-11-01
Add release notes directory

SiAdcock:

Release notes are ephemeral and shouldn't exist at the same level as more fundamental foundations docs. We should move them into a dedicated `Release notes` folder.

We should also add the v3 release notes as some teams have not yet upgraded to v3

## What does this change?

-   Move v3 - v4 migration notes to release notes directory
-   Add v3 release notes


jamie-lynch (approved):
📚

PR #1128, opened 2021-11-01
Split scanning and stats into their own files

jamie-lynch:

Refactor the get-usage script to make future changes easier

## What does this change?

-   Split scanning and calculating metrics logic into separate files
-   Stop writing csv file (as we don't use it)


SiAdcock (approved):
This is great! 💎

PR #1129, opened 2021-11-01
Create stories for Source v4

jamie-lynch:
## What is the purpose of this change?

This change adds stories for the v4 components so that the documentation is clear for users of the new `source-*` packages.

## What does this change?

- Update the title of kitchen stories to be `Source v4/source-react-components-development-kitchen/*`  
- Create story files for v4 that re-export everything from their v3 equivalents but with the correct v4 title
- Add a script to compare and sync the v3 and v4 stories
- Add a CI check to run the story comparison

PR #1131, opened 2021-11-02
Move foundations docs under "Source v4" category

SiAdcock:

#1129 introduces a "Source v4" category for the Storybook navigation. Arguably, the `source-foundations` docs  belong there too

## What does this change?

-   Move foundations docs under "Source v4" category
-   Group foundations docs under the `source-foundations` package name


jamie-lynch (approved):
🧱

PR #1133, opened 2021-11-02
Update usage script for new `source-*` packages.

jamie-lynch:

We have a script which scans a list of configured repositories to extract the usage of Source components. As we move to v4 and components come from new packages, we need to update the script to take this into account. This PR makes a number of changes to account for this, updating existing lists and metrics, and adding new ones are required.

## What does this change?

-   Include `source-*` packages when getting all components
-   Add function to group new and old components
- Update metrics and lists to account for grouped components


scripts/get-usage/get-all-components.ts
```diff
@@ -72,27 +73,38 @@ export const getAllComponentsAndPackages = async () => {
 	const packages: Record<string, string | string[]> = {
 		[packageNames.brand]: paths.brand,
 		[packageNames.icons]: paths.icons,
-		[packageNames.kitchen]: await getKitchenComponentPaths(),
 		...(await getComponentPackageNamesWithPaths()),
+		'@guardian/source-react-components': join(
```
SiAdcock, line 77: Should we add `@guardian/source-react-components` to `packageNames`?

scripts/get-usage/deduplicate-components.ts
```diff
@@ -0,0 +1,42 @@
+const componentNameChanges: Record<string, string | undefined> = {
+	'@guardian/src-ed-button/Button':
+		'@guardian/source-react-components-development-kitchen/EditorialButton',
+	'@guardian/src-ed-button/LinkButton':
+		'@guardian/source-react-components-development-kitchen/EditorialLinkButton',
+};
+
+export const deduplicateComponents = (components: string[]): string[] => {
```
SiAdcock, line 8: I wonder if we could come up with a more precise name for this function? Something like `groupSrcAndSourceEquivalentComponents`

scripts/get-usage/deduplicate-components.ts
```diff
@@ -0,0 +1,42 @@
...
+	const oldComponents: string[] = [];
+	const newComponents: string[] = [];
```
SiAdcock, line 10: I'm wary that the `old` and `new` prefixes here are a bit ephemeral. I wish we had given the "new world" an actual name instead of `new` or `v4`. I think what we mean here is `src` vs `source`, so maybe these should be the prefixes?

scripts/get-usage/deduplicate-components.ts
```diff
@@ -0,0 +1,42 @@
...
+		if (newComponents.includes(newComponent)) {
+			dedupedComponents.push(`${oldComponent} | ${newComponent}`);
+			newComponents.splice(newComponents.indexOf(newComponent), 1);
+		} else {
+			dedupedComponents.push(oldComponent);
+		}
+	});
+
+	return [...dedupedComponents, ...newComponents];
+};
```
SiAdcock, line 42: I'm a little worried that the complexity of this function could become unsustainable as components are moved around between the Kitchen and stable Source. For instance if we choose to demote Footer to the Kitchen, would we add lines to this function to try and group usage of this component in stable Source and in the Kitchen?

If this functionality is useful now, we should probably put a shelf life on it.

scripts/get-usage/get-enriched-results.ts
```diff
@@ -72,41 +86,87 @@ export const getEnrichedResults = (
...
+				percentageOfComponentsNotUsedAnywhere:
+					'This metric is the percentage of components, with the new and old equivalents grouped, that are not used anywhere',
+				percentageOfComponentsUsedInAtLeastTwoCodebases:
+					'This metric is the percentage of components, with the new and old equivalents grouped, that are used in at least two codebases. It ignores components from @guardian/src-ed-* packages and from the kitchen.',
+				percentageOfComponentsUsedInAtLeastTwoCodebasesIgnoringIconsAndBrand:
+					'This metric is the percentage of components, with the new and old equivalents grouped, that are used in at least two codebases. It ignores components from @guardian/src-ed-*, @guardian/src-brand and @guardian/src-icons packages as well as the kitchen and anything from @guardian/srouce-react-components that starts with Svg.',
+				percentageOfComponentsFromSourcePackages:
+					'This metric is the percentage of the components which are used that come from an @guardian/source-* package.',
+			},
+		},
```
SiAdcock, line 132: I love these headline metrics, they make reporting easy and help us understand what we ought to be focussing on 🎉

scripts/get-usage/get-all-components.ts
jamie-lynch: Yes, we totally should. This was me being lazy!

scripts/get-usage/deduplicate-components.ts
jamie-lynch: To summarise the discussion we had offline:

That's a really good point and something that I was slightly concerned about, even without having fully comprehended the complexity it would introduce. If we did make that change and felt we needed to introduce some further grouping, I do think it would be possible, but maybe a preferable scenario is that we encourage speedy migration and then drop monitoring for `src-*` components so that we don't have to shoulder both bits of complexity at the same time.

SiAdcock (approved):
This is awesome, thanks @jamie-lynch!

PR #1134, opened 2021-11-03
Improve toggle styling

marjisound:
This PR is using flex in css to keep the label and toggle switch aligned. Before when the label was long and the container width was smaller than the text, text would jump underneath the toggle. 

## What does this change?

-   add css styles to improve the toggle and label alignment
-   set margin left to 0 to give control to the consumer on how much space they need to have 

## Screenshots

**Before**
image

**After**
image

### Accessibility

### Responsiveness

### Documentation

-   [ ] Examples in Storybook


packages/@guardian/source-react-components-development-kitchen/src/components/toggle-switch/styles.ts
```diff
@@ -90,4 +93,5 @@ export const labelStyles = css`
 	display: inline-block;
 	transform: translateY(1px);
 	color: ${neutral[7]};
+	flex-shrink: 100;
```
oliverlloyd, line 96: 100 seems very large. What was the goal here with this property? There might be a different approach

packages/@guardian/source-react-components-development-kitchen/src/components/toggle-switch/styles.ts
```diff
@@ -3,9 +3,12 @@ import { neutral, success } from '@guardian/src-foundations';
 import { textSans } from '@guardian/src-foundations/typography';
 
 export const toggleSwitchStyles = css`
+	display: flex;
+	align-items: center;
 	button {
```
oliverlloyd, line 8: This is outside the scope of this PR but in general I'd avoid using CSS selectors like this. I find it's easier to scale css when you apply just the css for each element to itself, and don't use selectors to target beyond that. Eg. Have a `buttonStyles` piece of css.

The exception to this is when it's hard to target something because it's buried inside an external component (which might be the case here!)

packages/@guardian/source-react-components-development-kitchen/src/components/toggle-switch/styles.ts
marjisound: I wasn't sure what value to choose here. If I remove the flex-shrink, the button width would do down. And when flex-shrink is smaller than 50 it would still affect the button width.  I still wasn't sure if 100 is the best value. What other approach do you suggest? 

With no flex-shrink 
image

With flex-shrink 30
image

With flex-shrink 50 
image

packages/@guardian/source-react-components-development-kitchen/src/components/toggle-switch/styles.ts
marjisound: One reason I did this was that I wanted to have different style for `slim` & `medium` but I thought maybe it's better not to do this kind of check multiple times `isSlim ? slimStyles : mediumStyles,` once for the container div and once for the button. 

Also I wanted to add `aria-checked` style to button but wasn't sure how to do it in emotion. 
```
button[aria-checked='false'] {
    background-color: rgba(153, 153, 153, 0.5);
}
```
or 
```
button[aria-checked='true']:after {
  left: 20px;
  background: ${neutral[100]};
}
```

packages/@guardian/source-react-components-development-kitchen/src/components/toggle-switch/styles.ts
SiAdcock: You could set `flex: none` on the button, so it maintains the width you set.

packages/@guardian/source-react-components-development-kitchen/src/components/toggle-switch/styles.ts
SiAdcock: The Emotion way of selecting based on attributes:

```tsx
const buttonStyles = css`
 /* applied directly to the button */
  width: 3.188rem;
  height: 1.938rem;
  border-radius: 15.5px;

 /* applied to the button when not checked */
  &[aria-checked='false'] {
    background-color: rgba(153, 153, 153, 0.5);
  }

 /* applied to the :after pseudo-element when the button is checked */
  &[aria-checked='true']:after {
    left: 20px;
    background: ${neutral[100]};
  }
`
```

packages/@guardian/source-react-components-development-kitchen/src/components/toggle-switch/styles.ts
marjisound: Thanks @SiAdcock it worked well with `flex: none` 🙂

packages/@guardian/source-react-components-development-kitchen/src/components/toggle-switch/styles.ts
marjisound: Thank you both for giving me the idea. it's all great learning for me here :) I moved all the button styles to the button element

packages/@guardian/source-react-components-development-kitchen/src/components/toggle-switch/ToggleSwitch.stories.tsx
```diff
@@ -64,3 +65,18 @@ MediumWithLabel.args = {
...
 asChromaticStory(MediumWithLabel);
+
+// *****************************************************************************
+
+const toggleCssOverrides = css`
+	width: 150px;
+	background-color: lightblue;
+`;
+
+export const limitedWidth = Template.bind({});
```
oliverlloyd, line 76: I don't think we need this story (or at least not here, in Source). This seems like a use case for Dotcom and so it should probably live there

packages/@guardian/source-react-components-development-kitchen/src/components/toggle-switch/ToggleSwitch.tsx
```diff
@@ -74,15 +74,9 @@ export const ToggleSwitch = ({
...
-			css={[
-				toggleSwitchStyles,
-				isSlim ? slimStyles : mediumStyles,
-				cssOverrides,
-			]}
-			{...props}
-		>
+		<div css={[toggleSwitchStyles, cssOverrides]} {...props}>
 			<button
+				css={[isSlim ? slimStyles : mediumStyles]}
```
oliverlloyd, line 79: 👍

packages/@guardian/source-react-components-development-kitchen/src/components/toggle-switch/styles.ts
```diff
@@ -25,13 +30,12 @@ export const toggleSwitchStyles = css`
 `;
 
 export const mediumStyles = css`
-	button {
-		width: 3.188rem;
-		height: 1.938rem;
-		border-radius: 15.5px;
-	}
+	${buttonStyles}
```
oliverlloyd, line 33: You could also just include `buttonStyles` in the main element, as in: `css={[buttonStyles, isSlim ? slimStyles : mediumStyles]}`

This is perhaps a bit easier to reason about because then adding slimStyles is _only_ adding that css

oliverlloyd (commented):
This PR is really shaping up! I think we're almost there but just a couple more points. And thanks for responding so patiently to all these changes by the way

packages/@guardian/source-react-components-development-kitchen/src/components/toggle-switch/styles.ts
marjisound: Definitely looks better this way, thanks 👍

packages/@guardian/source-react-components-development-kitchen/src/components/toggle-switch/ToggleSwitch.stories.tsx
marjisound: Yeaa I was thinking so too, removed the story

SiAdcock (approved):
Got my vote! Great work @marjisound 💎

oliverlloyd (approved):
👍 from me too!

PR #1135, opened 2021-11-04
Fix errors with hash links on docs pages

jamie-lynch:

Currently when clicking an anchor link, the url shown in the browser is not updated accordingly. When navigating to a URL that includes a hash link, this is not respected. This PR fixes those problems. 

## What does this change?

-   Don't try and click the Docs button if the URL contains a hash
-   Set the `window.location.hash` value on page load


SiAdcock (approved):
Thanks @jamie-lynch 💎  And all of this can go away when storybookjs/storybook#13128 is fixed 🙏

PR #1136, opened 2021-11-04
feat: change text colour of input with error to be neutral.7

ob6160:
After checking internally with design & ux I'm making this change to the input error text so that it's `neutral.7` instead of `error.400`

## Before
Screenshot 2021-11-04 at 12 03 05

## After
Screenshot 2021-11-04 at 12 01 19

### DIFF
--- a/packages/@guardian/source-foundations/src/palette.ts
+++ b/packages/@guardian/source-foundations/src/palette.ts
@@ -289,6 +289,7 @@ export const text = {
 	anchorSecondary: neutral[7],
 	userInput: neutral[7],
 	inputLabel: neutral[7],
+	inputError: neutral[7],
 	inputLabelSupporting: neutral[46],
 	inputChecked: brand[400], //choice card only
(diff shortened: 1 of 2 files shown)


ob6160:
[Design details to check with Akemi](https://trello.com/c/nMbLPIdG/3534-design-details-to-check-with-akemi)

SiAdcock (approved):
Thanks so much @ob6160, this looks great 💎

SiAdcock:
Quick Q: do we want to also apply this to select and textarea themes?

ob6160:
> Quick Q: do we want to also apply this to [select](https://github.com/guardian/source/blob/main/packages/%40guardian/src-foundations/src/themes/select.ts) and [textarea](https://github.com/guardian/source/blob/c5103cdbe398fcaf41baf939f0740c9ad0baffd5/packages/%40guardian/src-text-area/styles.ts#L10) themes?

good Q! I'll ask Dom and get back to you

PR #1137, opened 2021-11-04
Allow element type to be chosen for Container

jamie-lynch:

This change allows the tag implemented by the `Container` component to be configured via a prop. This allows users to create more semantic markup when using this component. 

For example:

```tsx
<Container element="article"><p>Some text</p></Container>
```

now results in

```html
<article class="...">
  <p>Some text</p>
</article>

```

## What does this change?

-   Add `element` prop to Container and use 
-   Add a new story to cover the functionality

## Questions

**1. What elements should be available as options?**

   Currently, I've allowed the following to be passed to the `element` prop:

   * `div`
   * `article`
   * `aside`
   * `footer`
   * `header`
   * `nav`
   * `section`

**2. What's the best prop name?**

`element`? `el`? `as`?


SiAdcock (approved):
Brilliant, thanks @jamie-lynch! 🎉

PR #1132, opened 2021-11-02
Add docstrings for typography

SiAdcock:
## What is the purpose of this change?

It would be useful to have inline documentation for our Foundations' API. This will allow developers to see API documentation within their editor while they are using the API (see screenshots below)

## What does this change?

-   Add docstrings for headline, body, textSans and titlepiece
-   Link typography docs to our fonts repo
-   Add usage guidance and font families to typography API documentation

## Screenshots

**Titlepiece**

Screenshot 2021-11-02 at 14 23 15

**Headline**

Screenshot 2021-11-02 at 14 40 37

PR #1138, opened 2021-11-04
Foundations docstrings

SiAdcock:

Docstrings are useful for inline documentation (see also #1132)

## What does this change?

Adds docstrings to all Foundations exports


jamie-lynch (commented):
Just one link that I think needs updating but otherwise this looks great!
📖 🧶

jamie-lynch (approved):
LGTM 🚢

PR #1139, opened 2021-11-05
Move svgs to svgs directory

jamie-lynch:

https://github.com/guardian/source/pull/1118 added a script which pulled in all of the icon svg files from Figma. The final decision on the output directory was `packages/@guardian/src-icons/svgs` but, when the script was run, the output directory was set to `packages/@guardian/src-icons/svg` so the icons location in the repo didn't match the expectation. 

This PR moves them to where they should be.


SiAdcock (approved):
Good catch, I had not clocked that! 💎

PR #1140, opened 2021-11-05
Add payment icons to `get-icons` script

jamie-lynch:

#1118 adds a script to pull in icons from Figma. This PR updates that slightly to also pull in payment related icons which are held in a different frame.


SiAdcock (approved):
💵

PR #1143, opened 2021-11-05
Remove <svg> attributes

jamie-lynch:

#1118 introduced a script that pulls in the icon svg files from the Figma API in a bid to ultimately replace the manual process we currently go through when icons are added or updated. 

One of the current steps in that process is deleting a number of attributes from the  elements that make up the svg. This PR tries to implement those manual steps as part of the script that pulls in the svgs.

## What does this change?

Add a new `stripAttributes` function which is run on the svg content to:

*   Remove the `width`, `height` and `fill` attributes from the `<svg>` tag
* Remove the `fill` attribute from any child elements

### Before

```html
<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
  <path fill-rule="evenodd" clip-rule="evenodd" d="M17.45 6.12501L12.35 1.00003H11.65L6.55005 6.12501L7.40005 6.97501L11 4.22502V10H13V4.22502L16.6 6.97501L17.45 6.12501ZM7.40005 17L6.55005 17.9L11.65 23H12.35L17.45 17.9L16.6 17L13 19.75V14H11V19.75L7.40005 17Z" fill="#052962"/>
</svg>
```

### After

```html
<svg viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
  <path fill-rule="evenodd" clip-rule="evenodd" d="M17.45 6.12501L12.35 1.00003H11.65L6.55005 6.12501L7.40005 6.97501L11 4.22502V10H13V4.22502L16.6 6.97501L17.45 6.12501ZM7.40005 17L6.55005 17.9L11.65 23H12.35L17.45 17.9L16.6 17L13 19.75V14H11V19.75L7.40005 17Z"/>
</svg>
```


SiAdcock:
Great stuff, thanks @jamie-lynch! Quick question: it looks like this PR deletes the markup from the logo icons. Is this a side effect of the script? Ideally the script would not touch these SVGs, I think

jamie-lynch:
> it looks like this PR deletes the markup from the logo icons

As in the four files that have been deleted completely? I did this manually as, for now, the svg directory is just things that have been pulled in automagically. Happy to revert that if we want those svgs to still be there.

SiAdcock:
>  the svg directory is just things that have been pulled in automagically.

I think this directory should contain all SVGs regardless of where they come from. My thinking is that we should expose this entire directory to the public so that people can import raw SVGs if they need them. I don't believe there are any downsides to this, but let me know if you think of any 🙂

Perhaps we can explain which icons don't get populated automagically in the [`get-icons` README](https://github.com/guardian/source/tree/main/scripts/get-icons) ?

jamie-lynch:
> > the svg directory is just things that have been pulled in automagically.
> 
> I think this directory should contain all SVGs regardless of where they come from. My thinking is that we should expose this entire directory to the public so that people can import raw SVGs if they need them. I don't believe there are any downsides to this, but let me know if you think of any 🙂
> 
> Perhaps we can explain which icons don't get populated automagically in the [`get-icons` README](https://github.com/guardian/source/tree/main/scripts/get-icons) ?

Yeah, that makes sense. I'll revert those changes here and add some docs. 👍

SiAdcock (approved):
A beautiful optimisation, thanks @jamie-lynch  👍

PR #1144, opened 2021-11-05
fix: Use `as const` hint to help IDE autocomplete

mxdvl:

Improve IDE experience when using Source’s core values: `breakpoints` sizes ~and `palette` colours~.

Build on top of @ob6160’s #1101. Requires less maintenance than #1138. Also mentioned in https://github.com/guardian/dotcom-rendering/pull/3490#discussion_r737389629

## What does this change?

Uses the `as const` TS assertion in
- breakpoints
- ~palette~ (It doesn’t play nice with `Theme`)

## Screenshots

| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/76776/140520110-ff824c2d-ca4a-4267-b4a1-abf4df1a867b.png
[after]: https://user-images.githubusercontent.com/76776/140520210-e28454a4-1337-4d73-aa36-22d33e44230f.png

### Accessibility

### Responsiveness

-   [X] Tested at all breakpoints 😛 (not really)

### Known issues

~Colours don’t get a visual treatment, so you can only read the string.~

### DIFF
--- a/packages/@guardian/source-foundations/src/breakpoints.ts
+++ b/packages/@guardian/source-foundations/src/breakpoints.ts
@@ -42,4 +42,4 @@ export const breakpoints = {
 	desktop: 980,
 	leftCol: 1140,
 	wide: 1300,
-};
+} as const;
--- a/packages/@guardian/source-foundations/src/breakpoints.ts
+++ b/packages/@guardian/source-foundations/src/breakpoints.ts
@@ -15,23 +15,6 @@ export type Breakpoint =
 /**
  * [Storybook](https://guardian.github.io/source/?path=/docs/source-v4-source-foundations-media-queries--page#breakpoints) •
  * [Design System](https://theguardian.design/2a1e5182b/p/41be19-grids)
- *
- *	`breakpoints.mobile` -> 320px
- *
- *	`breakpoints.mobileMedium` -> 375px
- *
- *	`breakpoints.mobileLandscape` -> 480px
- *
- *	`breakpoints.phablet` -> 660px
- *
- *	`breakpoints.tablet` -> 740px
- *
- *	`breakpoints.desktop` -> 980px
- *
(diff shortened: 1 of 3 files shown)


ob6160 (approved):
aha, brilliant - glad it can be applied to more places. seems like a good improvement from my perspective.

Edit: looks like the `Theme` interface might need updating to get the type checker to pass? (just had a cursory glance)

mxdvl:
@ob6160 I had a look at updating `Theme`, and it is a pretty daunting task. Due to the minimal benefits of having colours as strings–who can read `#FBA681` anyways–I’ve decided to keep it only for breakpoints for the time being.

ob6160:
> @ob6160 I had a look at updating `Theme`, and it is a pretty daunting task. Due to the minimal benefits of having colours as strings–who can read `#FBA681` anyways–I’ve decided to keep it only for breakpoints for the time being.

that makes sense 🙂.  I'm guessing using the example in that file: `const [sport300, sport400, sport500, sport600, sport800] = colors.blue` and the current object structure was a conscious design decision.

If there aren't strong reasons for it to be the way it is I'm wondering if a future improvement could be to make that a mapping instead so you could have a human readable way to access colours: `colors.reds.news.100` for example

ob6160 (dismissed):

ob6160:
> The `colors` object was an attempt to align with [System UI's theme specification](https://system-ui.com/theme/), an early contender for defining design tokens. I don't think there's a strong case to stick to it anymore. This object is only ever used internally, so we can change the structure if necessary.


re: this, I'd be happy to spend an hour or so at some point and open a proposal PR based on that suggestion I made above, any thoughts on whether it'd be useful?

SiAdcock (approved):

SiAdcock:
@ob6160 I'll set up a chat with you 😊

PR #1154, opened 2021-11-08
Revert "Expose more typography primitives"

SiAdcock:
Reverts guardian/source#1153

Argh, these are already exported in the [source-foundations index](https://github.com/guardian/source/blob/9cc996250971354f78e99a11760bf037d4293dd0/packages/%40guardian/source-foundations/src/index.ts#L79-L84). No need to do it again here


jamie-lynch (approved):

jamie-lynch (commented):
Actually, as we discussed online, maybe we should leave this in place so that it works for src-foundations

SiAdcock:
Let's keep these as they will now be available in `src-foundations` 🎉

PR #1156, opened 2021-11-08
Retain error/success border size and colour on `:active` state

ob6160:
We would like the border transition on `:active` to not happen when these components are in an **error** or **success** state.

This gif demonstrates the current behaviour (clicking on the input to trigger `:active`):
out

## What does this change?
This changes the behaviour so that when the **select**, **text area** and **input**`:active` state is activated and they are in error or success states the border remains the same.

## Accessibility
We had some discussions around the impact of this change upon accessibility. Namely, showing a flash of activation for keyboard/mouse users. The standard tab focus outer border will still work with this change. It would be interesting to hear peoples' opinion on this.

Perhaps, when the `:active` state is activated we could show a different shade of the error colour. Something like an `errorBorderActiveColour`?


SiAdcock (approved):
Thanks @ob6160, sorry for the delay in reviewing. This looks awesome! I'll raise with Akemi the point about whether we should have a different `:active` border colour for fields in an error/success state

PR #1160, opened 2021-11-09
Use rollup-plugin-dts in source-react-components

SiAdcock:

`rollup-plugin-ts` currently generates `.d.ts` for `source-react-components`.

As per the rollup config, `src-foundations` imports are re-written as `source-foundations` and externalised. 

This works fine with the ESM and CJS outputs, but it doesn't work for `.d.ts` files. In this case, the original `src-foundations` paths are preserved.

`rollup-plugin-dts` correctly respects our paths in the `.d.ts` files

## What does this change?

-   Install and consume `rollup-plugin-dts` in `source-react-components`


jamie-lynch (approved):
🧻 🆙

PR #1161, opened 2021-11-09
Add as const type assertion on palette

ob6160:
@mxdvl's PR #1144 introduced `as const` assertions for improved developer experience when accessing theme tokens such as breakpoints. This was a great improvement 🙂. 

Sadly we weren't able to include an `as const` assertion on the palette to narrow down to the individual hex code colour values. This was because it conflicted with the looser types defined on the Theme interface (`string` vs `#FFFFFF` for example): https://github.com/guardian/source/pull/1161/files#diff-06f78fe3b5cad62762cc5b85c0f25c385f5065842bb33f4b3b44de3066e4b57f

## What does this change?
This PR allows the `as const` assertion to work by typing the Theme interface using the default theme objects that we already have defined for every component.

This has the added benefit of meaning that we don't need to update the Theme interface every time that we add a theme token to a component, only its default in `src-foundations/themes`.

Using this technique, we are able to type the Theme interface like this:

```typescript
import {
	accordionDefault,
	// ... other defaults ... (shortened for brevity)
} from '@guardian/src-foundations/themes';

declare module '@emotion/react' {
	export interface Theme {
		/**
		 * As an example, definition for the accordionDefault theme will be:
		 *	accordion: {
		 *		textPrimary: "#121212";
		 *		borderPrimary: "#999999";
		 *	}
		 **/
		accordion: typeof accordionDefault.accordion;
		button: typeof buttonDefault.button;
		checkbox: typeof checkboxDefault.checkbox;
		choiceCard: typeof choiceCardDefault.choiceCard;
		footer: typeof footerBrand.footer;
		label: typeof labelDefault.label;
		layout: {};
		link: typeof linkDefault.link;
		radio: typeof radioDefault.radio;
		select: typeof selectDefault.select;
		textArea: {};
		textInput: typeof textInputDefault.textInput;
		userFeedback: typeof userFeedbackDefault.userFeedback;
	}
}
```

## Screenshots
| Before      | After      |
|-------------|------------|
| ![before][] | ![after][] |

[before]: https://user-images.githubusercontent.com/1771189/143948885-4a664a14-2f70-4743-b2ca-9e0680e872b0.png

[after]: https://user-images.githubusercontent.com/1771189/143948898-d14ff4c2-55b9-4776-aed0-68f0082b2cbe.png


mxdvl (approved):
Oui, j’aime ça!

ob6160:
> Oui, j’aime ça!

Merci beaucoup!

packages/@guardian/source-foundations/src/palette.ts
```diff
@@ -310,7 +310,7 @@ export const background = {
 	ctaTertiaryHover: '#E5E5E5',
 	input: neutral[100],
 	inputChecked: brand[500],
-};
+} as const;
```
ob6160, line 313: @SiAdcock  quick thought – I added `as const` here, because some of these colours don't pull from the `colors` object defined above. Is that alright, or would we ideally want them all to live in the `colors` object?

SiAdcock: The hardcoded colours here are snowflakes that don't live on the palette. Ideally these would be defined using some darkening algorithm e.g.  `ctaSecondaryHover: darken(ctaSecondary, '10%')` 

Using `as const` for now is, I guess, part of the cost of maintaining snowflakes like this.

SiAdcock (approved):
This is an amazing improvement! Thanks so much @ob6160 💎 👌 🥇

packages/@guardian/source-foundations/src/palette.ts
ob6160: interesting.. lots of ways we could define colour variants like this.. will have a think 💭

PR #1162, opened 2021-11-10
Handle theme name changes when coming from component package

jamie-lynch:

This change fixes a bug in the plugin where theme exports are not correctly renamed if they are imported from a component package.

### Context

In Source v4, numerous changes have been made to the way that the packages are structured. One such change involves the theme exports for each component now coming from `@guardian/source-react-components`. Previously they were available both from `@guardian/src-foundations/themes` and the relevant `@guardian/src-<component>` package. As well as changing location, they have also been renamed to add the word `Theme`. 

To help with this change, we have created two eslint plugins that will show errors relating to specific imports and, in many cases, autofix those problems. 

### What wasn't working?

The plugins would rename the theme variables if they were imported from `@guardian/source-foundations/themes` but not if they came from a `@guardian/src-<component>` package. This PR fixes that.

## What does this change?

-   Update logic to not check for the package location when checking if the import name is one of the renamed values
-   Add a new test case to cover this

## Next steps

One consideration with this renaming functionality is that, while it renames the imports, it will not rename the variables where they are used throughout the code. This may lead to a confusing situation where all the imports are now valid but the code is not, and it is not necessarily clear to the user why that is the case. Maybe we shouldn't be renaming them, just adding some information into the error message that explains what has happened?

### DIFF
--- a/lib/eslint/rules/valid-import-path.test.ts
+++ b/lib/eslint/rules/valid-import-path.test.ts
@@ -146,7 +146,7 @@ import {  size as s } from '@guardian/source-foundations';`,
 		output: `import { headlineObjectStyles as hl, bodySizes as bs } from '@guardian/source-foundations';`,
 	},
 	{
-		// Themes that now come from react-components
+		// Themes that now come from react-components and have changed names (from foundations)
 		code: `import { labelDefault } from '@guardian/src-foundations/themes';`,
 		errors: [
 			{
@@ -403,6 +403,17 @@ ruleTester.run('valid-import-path', validImportPath, {
 				},
 			],
 			output: `import { palette } from '@guardian/src-foundations';`,
+		},
+		{
+			// Themes that now come from react-components and have changed names (from component)
+			code: `import { buttonReaderRevenue } from '@guardian/src-button';`,
+			errors: [
+				{
+					message:
+						"@guardian/src-* packages are deprecated. Import from '@guardian/source-react-components' instead.",
+				},
+			],
+			output: `import { buttonThemeReaderRevenue } from '@guardian/source-react-components';`,
 		},
 	],
 });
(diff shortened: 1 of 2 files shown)


SiAdcock:
Sorry for the delay in reviewing this. I think you're right that it would be confusing for the user if imports are updated, but usages within the file are not. 

The possible alternatives:

- Could we rename the export as we import it? i.e `import { buttonThemeDefault as buttonDefault } from '@guardian/source-react-components'`
- Instead of renaming, add an error message that explains the new naming and asks user to fix it themselves
- We look for all occurrences of theme name imports (`accordionDefault`, `buttonReaderRevenue`, etc) and replace them wholesale. This might not be dangerous right now, but the longer the rule hangs around, the higher the likelihood of erroneous replacements occurring. e.g. `guardian/facia-tool` has a component called [`ButtonDefault`](https://github.com/guardian/facia-tool/blob/main/fronts-client/src/components/inputs/ButtonDefault.ts), which is close to (although not an exact match of) the `buttonDefault` theme.

jamie-lynch:
> * Instead of renaming, add an error message that explains the new naming and asks user to fix it themselves

I think my preference would be for this option. Whilst adding an alias makes the migration easier, it's probably cleaner to just update everywhere. Given how easy it is to view the errors that can't be autofixed and (with some clear errors messages), replace all instances manually, I'm not sure it's worth trying to solve the complexity of replacing everywhere in code. 

We already have a similar path in the code when exports have been removed completely so hopefully it won't be too hard to do something similar. I'll give it a go.

jamie-lynch:
I've updated this to just add a message, rather than renaming imports/exports that have changed names.

SiAdcock (approved):
Lovely stuff 👍

PR #1164, opened 2021-11-11
Add CheckboxGroup docstring

jamie-lynch:

This PR adds the standard component docstring to the CheckboxGroup component, which didn't previously have one.


SiAdcock (approved):
Good catch! 💎

PR #1167, opened 2021-11-11
Update kitchen to use Source v4

jamie-lynch:

Update the kitchen to use Source v4 so that projects that are updating can drop all dependencies on `src-*` packages.

## What does this change?

-   Update the dependencies
-   Update the `tsconfig.json`
- Update all imports

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -39,6 +39,14 @@ module.exports = ({ config, mode }) => {
 
 	config.resolve.extensions.push('.ts', '.tsx');
 	config.resolve.alias = {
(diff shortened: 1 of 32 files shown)


jamie-lynch:
I've created a new `kitchen-v1` branch off of main which this PR is now against so that we can keep working on v1 whilst allowing contributions and releases of the 0.X version.

jamie-lynch:
I've published version `1.0.0-rc.0` to npm with the `next` tag so that `yarn add @guardian/source-react-components-development-kitchen` will still give you the latest `0.X` release.

SiAdcock (approved):
Superb! 💎

PR #1177, opened 2021-11-17
Make running chromatic on dependabot a manual process

jamie-lynch:

Currently chromatic runs on dependabot PRs via a workaround where a workflow runs on all dependabot PRs and, when it completes, that triggers the chromatic workflow. We are seeing lots of runs of the chromatic workflow, partly as dependabot rebases PRs on each merge to main, and maybe also because of an issue with this approach.

This PR makes it a manual process to run chromatic on dependabot PRs. As chromatic is a required check, it will have to be run before the PR is merged but this means it can be run once just before merging.

## What does this change?

-   Remove the `Dependabot PR Check` workflow
-   Add `workflow_dispatch` as a trigger to the chromatic workflow

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,9 +1,6 @@
 name: Chromatic
 on:
-  workflow_run:
-    workflows: ['Dependabot PR Check']
-    types:
-      - completed
+  workflow_dispatch:
   push:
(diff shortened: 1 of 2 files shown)


SiAdcock (approved):
Great call! 💎  Let's see how this feels for now 👍

PR #1186, opened 2021-11-22
Update eslint-integration-test

jamie-lynch:

This PR updates the `eslint-integration-test` playground and README to account for the fact that there are now two separate eslint plugins. 

This work follows on from https://github.com/guardian/source/pull/1092 which split the plugin into two and https://github.com/guardian/source/pull/1093/files which allowed custom eslint config to be passed into each test.

## What does this change?

-   Upgrade the playground to provide different files and configs for each plugin
-   Update the README to explain how the tests and playground work with the two separate plugins


scripts/eslint-integration-test/react-components-playground.ts
```diff
@@ -0,0 +1,9 @@
+// This file acts as a playground for the `@guardian/eslint-plugin-source` plugin
```
SiAdcock, line 1:
```suggestion
// This file acts as a playground for the `@guardian/eslint-plugin-source-react-components` plugin
```

scripts/eslint-integration-test/react-components-playground.ts
```diff
@@ -0,0 +1,9 @@
...
+// You can modify it's contents and see what happens.
```
SiAdcock, line 2:
```suggestion
// You can modify its contents and see what happens.
```

SiAdcock (approved):
This is great, thanks @jamie-lynch

PR #1188, opened 2021-11-26
Process brand icons automagically

jamie-lynch:

The `get-icons` scripts pull in the svg source for icons from the Figma API and undertakes a small amount of processing to align the source with the expected format. Previously, this process ignored brand icons as they did not follow the standard pattern. This PR updates the process to account for those differences. 

The key difference is that the `fill` attribute is not removed from child components of the `<svg>` tag. There were some other inconsistencies which this script does not respect. It is hoped that those changes were the results of inconsistent practice and not deliberate differences and, therefore, the icons output by this script will remain visually consistent, despite the subtle changes to the source.

## What does this change?

-   Don't filter out brand icons when getting relevant components from Figma
-   Don't remove attributes from child tags for brand icons


SiAdcock (approved):
This is awesome, thanks so much @jamie-lynch 💎 

It's probably a good idea to move icons with brand colours onto their own frame in Figma so we don't need to keep a list of those icons in here. I'll check with Akemi to see if she's happy with this. Let me know if you can think of other ways we can improve the organisation in Figma to make this script more efficient

PR #1189, opened 2021-11-26
Remove coveralls

jamie-lynch:

This PR removes the coveralls step from the CI workflow as we never got it to work as expected so, until we're ready to investigate further, there's not much point sending the coverage reports.


SiAdcock (approved):
👋  🧥

PR #1194, opened 2021-11-29
Replace fill-Rule with fillRule to prevent react dom warning

ob6160:
Fixes a warning that was thrown by react-dom in Gateway when the `google-brand` component was used: **"Warning: Invalid DOM property `fill-rule`. Did you mean \`fillRule\`?"**

Screenshot 2021-11-29 at 19 31 47


jamie-lynch (approved):
📏

PR #1198, opened 2021-12-01
Get icons react components

jamie-lynch:

This PR adds some new functionality to the `get-icons` script that writes out React components for each icon. 

## How?

- [x] Process the svg for use within a React component
    - [x] Fix the indentation
    - [x] [Replace `style` attributes with the correct equivalent](#style)
    - [x] Update case of `fill-rule` and `clip-rule` 
- [x] Generate the source for a React component by wrapping the processed SVG in some boilerplate
- [x] Write this to a file within `packages/@guardian/src-icons/components/icons`
- [x] Write `index.ts` file with all exports

### Structure

Currently the `src-icons` package has a flat structure. This PR adds a new `components` directory (alongside the `svgs` directory the script already writes too). Within the components directory is:

- `index.ts` - Exports
- `types.ts` - Type definitions relating to icons
- `icons/` - A directory containing a file per icon component

### Component Name

Currently our icon components are named `Svg${IconName}`. These new components are named `${IconName}Icon` instead, as this better reflects the component whilst abstracting the specific technical details from the user. 

### Style

Some icons contain elements with the style attribute (e.g. `style="mask-type:alpha"`). Part of this script transforms these tags to split the parts of style into their own attributes (e.g. `mask-type="alpha"`).


SiAdcock:
Wow, this is amazing and much needed! Thanks so much @jamie-lynch for taking it on. 

I have a lazy question about the API. If a consumer wanted to use these new icons, what would the import statement look like? From what I can tell, it would be something like this, but I might be wrong:

```ts
import { AlertPhoneIcon } from '@guardian/src-icons/components';
```

jamie-lynch:
> Wow, this is amazing and much needed! Thanks so much @jamie-lynch for taking it on.
> 
> I have a lazy question about the API, forgive me 😅. If a consumer wanted to use these new icons, what would the import statement look like? From what I can tell, it would be something like this, but I might be wrong:
> 
> ```ts
> import { AlertPhoneIcon } from '@guardian/src-icons/components';
> ```

In theory yes, however, currently neither the `components` or `svgs` directories are included in the published package. 

I should have expanded on this more in the PR description but I haven't thought much about exposing these components from the package yet, nor how the migration story would be for consumers and for the internal components. Maybe that's something that could be done in a subsequent PR?

SiAdcock:
> I haven't thought much about exposing these components from the package yet, nor how the migration story would be for consumers and for the internal components. Maybe that's something that could be done in a subsequent PR?

Good shout. I'll add this to the Big Big Task List

SiAdcock (approved):
This is a lovely addition, thanks @jamie-lynch 💎

PR #1199, opened 2021-12-01
Handle lack of FIGMA_TOKEN more gracefully

jamie-lynch:

The `get-icons` scripts relies on a `FIGMA_TOKEN` env var being present. Currently if this is not set, the resultant error is unhandled and it is not immediately obvious what the problem is. This PR adds a check to ensure that it has been set at the beginning of the process and throws an appropriate error if not. 

## What does this change?

-  Add a `Log.error` method which prints red text
-   Check if `FIGMA_TOKEN` is present and error if not

### DIFF
--- a/scripts/get-icons/index.ts
+++ b/scripts/get-icons/index.ts
@@ -3,6 +3,7 @@ import { existsSync, mkdirSync, writeFileSync } from 'fs';
 import { formatSVG } from './format';
 import { FIGMA_OPTIONS, ICON_FILE, ICON_FRAMES, OUTPUT_DIR } from './config';
 import { stripAttributes } from './process';
+import { Log } from './log';
 
 interface FigmaComponentsResponse {
 	meta: {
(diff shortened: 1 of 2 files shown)


SiAdcock (approved):
Great! We should do better error logging everywhere. This is a useful pattern to build on 👍

scripts/get-icons/index.ts
```diff
@@ -59,6 +60,10 @@ if (!existsSync(OUTPUT_DIR)) {
 	mkdirSync(OUTPUT_DIR);
 }
 
+if (!process.env.FIGMA_TOKEN) {
+	Log.error('\x1b[31mFIGMA_TOKEN env var not set\x1b[0m');
```
jamie-lynch, line 64:
```suggestion
	Log.error('FIGMA_TOKEN env var not set');
```

SiAdcock (approved):
Great stuff 👍

PR #1201, opened 2021-12-06
Add breaking change note about fromEntries polyfill removal

SiAdcock:

In #982, we removed an inline polyfill for `Object.fromEntries`. We now expect consumers to define the environment in which Source is executed. This is in line with our [NPM package recommendations](https://github.com/guardian/recommendations/blob/main/npm-packages.md#using-guardian-npm-packages)

Since this has caused a high volume of Sentry errors in dotcom, we should make consumers aware of this change so they can test accordingly.

## What does this change?

- Add a note to the v4 release notes advising consumers of this breaking change.


sndrs (approved):

sndrs:
i found it a little hard to pull out actual things i'd need to do from that doc, it's hard to scan and i'd end up creating my own checklist that we could probably provide for users? 

just a thought from first reading it though – 👍 for this mention

PR #1202, opened 2021-12-06
chore(types): Typography constants

mxdvl:

Make all the typography data `readonly` types via the “`as const` assertions”, which enables better IDE information and detailed numeric values instead of `number`. TypeScript now ensures all keys are explicitly required. For example:

```ts
// instead of
const obj: Record<string, () => string>;

// you get
const obj: Record<'small' | 'medium' | 'large', () => string>;
```

It is possible that some of the narrower template string types are a step too far and it might be a good decision to step some of it back.

`Object.fromEntries` is not supported by all browsers, and freeings ourselves from its dependencies is helpful.

## What does this change?

- Use the `as const` assertions in typography.
- Narrow `TypographyStyles` and `Fs` types
- Remove the `TypographySizes` type
- Remove `Object.fromEntries` in favour of explicit keys and values.

### DIFF
--- a/.changeset/happy-donkeys-drum.md
+++ b/.changeset/happy-donkeys-drum.md
@@ -1,6 +1,5 @@
 ---
 '@guardian/source-foundations': minor
-'validate-dist': patch
 ---
 
 Typography constants with IDE hints
--- a/.changeset/happy-donkeys-drum.md
+++ b/.changeset/happy-donkeys-drum.md
@@ -1,0 +1,6 @@
+---
+'@guardian/source-foundations': minor
+'validate-dist': patch
+---
+
+Typography constants with IDE hints
--- a/packages/@guardian/source-foundations/src/index.ts
+++ b/packages/@guardian/source-foundations/src/index.ts
@@ -98,6 +98,7 @@ export type {
 	FontWeightDefinition,
 	Option,
 	TypographyStyles,
+	TypographySizes,
 	TitlepieceSizes,
 	HeadlineSizes,
 	BodySizes,
--- a/packages/@guardian/source-foundations/src/index.ts
+++ b/packages/@guardian/source-foundations/src/index.ts
@@ -98,7 +98,6 @@ export type {
 	FontWeightDefinition,
 	Option,
 	TypographyStyles,
-	TypographySizes,
 	TitlepieceSizes,
 	HeadlineSizes,
 	BodySizes,
--- a/packages/@guardian/source-foundations/src/typography/fs.ts
+++ b/packages/@guardian/source-foundations/src/typography/fs.ts
@@ -27,17 +27,18 @@ export const fs: Fs =
 	(category) =>
 	(level, { lineHeight, fontWeight, fontStyle, unit }) => {
 		const fontFamilyValue = fontMapping[category];
-		const fontSizeValue =
+		const fontSizeValue: `${number}rem` | number =
 			unit === 'px'
-				? fontSizeMapping[category][level]
-				: `${remFontSizeMapping[category][level]}rem`;
-		const lineHeightValue =
+				? Number(fontSizeMapping[category][level])
+				: // @ts-expect-error -- the types actually overlap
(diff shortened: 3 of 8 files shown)


packages/@guardian/source-foundations/src/typography/fs.ts
```diff
@@ -27,17 +27,18 @@ export const fs: Fs =
...
 	(level, { lineHeight, fontWeight, fontStyle, unit }) => {
 		const fontFamilyValue = fontMapping[category];
-		const fontSizeValue =
+		const fontSizeValue: `${number}rem` | number =
 			unit === 'px'
-				? fontSizeMapping[category][level]
-				: `${remFontSizeMapping[category][level]}rem`;
-		const lineHeightValue =
+				? Number(fontSizeMapping[category][level])
+				: // @ts-expect-error -- the types actually overlap, see https://gist.github.com/mxdvl/5e31fd5b13670b6a41ddac6c65efeee4
```
SiAdcock, line 33: I really cannot get my head around why this is necessary 😅

packages/@guardian/source-foundations/src/typography/index.ts
```diff
@@ -82,14 +82,22 @@ const titlepiece = Object.fromEntries(
  *
  * `headline.xlarge()` -> 50px
```
SiAdcock, line 83: Given that this docstring (and the docstrings for other `TypographyApi` instances) is here purely for IntelliSense, I think they're no longer necessary. What do you think?

Screenshot 2021-12-08 at 07 55 00

mxdvl: Removed in bbeb4f69c07586cbcade3cd2587e81b8f5e8627e

packages/@guardian/source-foundations/src/typography/fs.ts
mxdvl: The conversion and division by 16 is not known by the TS compiler. The pixel dimensions as a single integer gets converted to a `number`.

The way around this would be to statically declare pixels to rem for each of the sizes, making this less DRY.

mxdvl: I think this gets quite ugly:

Screenshot 2022-01-17 at 11 18 00

packages/@guardian/source-foundations/src/index.ts
```diff
@@ -98,7 +98,6 @@ export type {
 	FontWeightDefinition,
 	Option,
 	TypographyStyles,
-	TypographySizes,
```
SiAdcock, line 101: This would be a breaking change. Is there a way to keep it in for now? I don't think anyone is using it yet, so maybe we can mark it as deprecated and remove it in the next major version

mxdvl: fixed in [5925e0f](https://github.com/guardian/source/pull/1202/commits/5925e0ffc03013e9af5877d4f422b91ffbe24597)

scripts/validate-dist/source-validate-types.ts
```diff
@@ -8,7 +8,6 @@ export type {
 	FontWeightDefinition,
 	Option,
 	TypographyStyles,
-	TypographySizes,
```
SiAdcock, line 11: Since this is no longer being removed, can we reinstate this type here? Then if we regenerate the changeset, it should remove the bump to `validate-dist`

.changeset/happy-donkeys-drum.md
```diff
@@ -0,0 +1,6 @@
+---
+'@guardian/source-foundations': minor
+'validate-dist': patch
```
SiAdcock, line 3:
```suggestion
```

mxdvl:
Just need a final 🟢 !

SiAdcock (approved):
👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍

PR #1203, opened 2021-12-06
use `tsconfig-paths-webpack-plugin`

sndrs:
## What does this change?

-   `tsconfig-paths-webpack-plugin` means we dont have provide ts aliases anymore
-   disabling controls for chromatic stories triggers [a bug](https://github.com/storybookjs/storybook/issues/15503) that we dont need to, because they're canvases anyway. it seems to clear up the issue with selecting the wrong story type as you navigate around

## Why?

I tried to remove babel but no chance :(

so:

- i tried to update to latest to adopt the [tentative support for emotion 11](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#emotion11-quasi-compatibility) but got nowhere
  - they're enabling feature flags in 6 because many of the changes are breaking but they dont want to do a breaking release
  - it seemed to add complexity we didn't need to manage to me
- I tried to remove our extra webpack config and just set babel preset directly to use the `automatic` runtime/`importSource`, but the mdx webpack loader doesn't support the `automatic` runtime
- i tried to set the webpack `tsx` rule babel config only, but then it turns out `mdx` rules inherits from the `tsx` one

so gave up on all that.

_but_ i did find the things in this PR out along the way, so thought we could use them

it's Better Than Nothing™ 😜:pray:

### DIFF
--- a/.storybook/manager-head.html
+++ b/.storybook/manager-head.html
@@ -1,0 +1,25 @@
+<script>
+	// Hacky way of making sure that we're on the docs page for playground stories.
+	// Setting viewMode: 'docs' in lib/story-intents.ts should do this, but there's a bug.
+	// https://github.com/storybookjs/storybook/issues/13128#issuecomment-798927176
+	// It's very closely bound to storybook implementation details so liable to break after updating.
+	window.onload = () => {
+		const observer = new MutationObserver((mutationsList) => {
+			for (const mutation of mutationsList) {
+				mutation.addedNodes.forEach((addedNode) => {
+					if (addedNode.matches('[href$="--playground"]')) {
+						addedNode.href = addedNode.href.replace(
+							'path=/story/',
+							'path=/docs/',
+						);
+						if (addedNode.dataset.selected) {
+							addedNode.click();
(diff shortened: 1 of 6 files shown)


SiAdcock:
Thanks @sndrs 

> it's Better Than Nothing™ 😜🙏

😄 

> `tsconfig-paths-webpack-plugin` means we dont have provide ts aliases anymore

🎉 

> it seems to clear up the issue with selecting the wrong story type as you navigate around

Wow! Does that mean we can delete [these hacks](https://github.com/guardian/source/blob/7c21c52b05a5535d938697b4ae1b8f5751b24277/.storybook/preview.tsx#L26-L66)? 🙏

sndrs:
> Wow! Does that mean we can delete these hacks? 🙏

oooh i hadn't seen those, i'll see

sndrs:
> Wow! Does that mean we can delete these hacks? 🙏

it didn't. but looking into it did lead me to an alternative to polling in a004c39

SiAdcock (approved):
Amaze 🤩

PR #1205, opened 2021-12-07
move accordion code to `source-react-components`

sndrs:

PoC moving code from v3 to v4

## What does this change?

-   moves the accordion source code into `source-react-components`


sndrs:
covered by #1208

PR #1207, opened 2021-12-09
Move source code to v4 packages

sndrs:
## What does this change?

-  move (almost) all the source code that was spread across the `src-*` package directories to the v4 `source-*` directories

## It's enormous

yeah it is. we're going to have to rely on our automated QA stuff to validate it.

we also need to figure out what to do about `packages/@guardian/src-helpers/storybook-bg.ts` and the docs (docblocks and mdx) need updating to refer to the correct links and packages

PLUS I think there's some actual TS things we need to look at that the new config has revealed/is making too much of a fuss about


SiAdcock:
Good lord this is a huge one! Great work persevering 💪

Is it worth trying to break this down? For instance if we just tackle the kitchen in the first instance... then just move theme... then just move helpers ... then ... ? I don't think it matters so much if delivery in this branch is incremental, given that we eventually intend to merge the v4.0.0 branch into main in one shot. Unless incremental delivery is literally impossible. Or if we can fix the issues here by simply working through the error messages. Do you have a feeling on that?

sndrs:
its really hard to break down, because so many of the changes affect the other changes. when you move a package e.g. `src-icons` you need to update tsconfigs, storybook configs etc and all the places that use `src-icons` (modules, stories etc), some of which have moved to v4 location, others which haven't. the things still in old place have one set of changes for the new location of `src-icons`, then you move one of them and you need to update the path once more, meaning you have a second set of changes for a new location. VSCocde sometimes manages that for you, but not always. each move is really hard to get a working build/viable PR for, because the tendrils go so far – do you keep building `src` on CI while this is happening for example? what about the icons import script that refs the `src-icons` dirs?

i think it will be easier to merge this and fix the bugs as if they were bugs in an existing codebase personally. will ping you :)

sndrs:
@jamie-lynch `validate-dist` has already saved our bacon! 🥓 🎉

PR #1215, opened 2021-12-13
Align peerdeps

sndrs:

bring the way we use peerdeps up to speed

## What does this change?

- eslint plugins express eslint version they need
- packages express the widest possible range of peerdeps
- workspace installs the lowest possible version of dev deps that our packages also have peerdeps on


packages/@guardian/source-react-components/package.json
```diff
@@ -24,9 +24,9 @@
     "rollup-plugin-ts": "^1.4.7"
   },
   "peerDependencies": {
-    "@emotion/react": "^11.1.2",
-    "@guardian/source-foundations": "^4.0.0-rc.1",
-    "react": "^17.0.1"
+    "@emotion/react": "^11.0.0",
+    "@guardian/source-foundations": "4.0.0-rc.5",
+    "react": "^17.0.0"
```
SiAdcock, line 29: This is cool and specific and I love it 😊 For posterity, is there a reason you couldn't go with `17.x` as the semver range?

SiAdcock (approved):
Yeeees 💯

packages/@guardian/source-react-components/package.json
sndrs: Just for consistency with any minor/patch version minimums, as with prettier in configs. Suddenly you need to go from 4.x to ^4.1.0. Going with the one format seemed easier. Do you prefer the x?

SiAdcock: No, I much prefer the consistency here. Good call 👍

PR #1225, opened 2021-12-31
Fix typo in max length arg name

SiAdcock:

The max length story for TextArea is broken because the arg name has a typo. 

## What does this change?

-   Fix typo in arg name


sndrs (approved):
superb

PR #1248, opened 2022-01-25
feature: Adds paperclip icon

AshCorr:

Adds the Paperclip icon to the source icon set.

This icon is currently used by Frontend to indicate a link in the comments section but does not exist in DCR. The purpose of this PR is to add the icon to Source so that it can be used in DCR for rendering comments. 

See https://github.com/guardian/discussion-rendering/pull/550 for related DCR PR.

## What does this change?

-   Adds SvgPaperclip react component.

## Screenshots

image
image

### Documentation

-   [x] Examples in Storybook

### Known issues

- Icon is styled a bit differently to the other icons that exist in Source. Unsure if theres a requirement for all icons to follow the same style guide.
- Viewbox is 0 0 15 15 instead of the usual 0 0 30 30 used by other icons. Is there a tool that can be used to easily resize SVG's?


AshCorr:
Decided that we would not re-implement this feature in DCR

PR #1249, opened 2022-01-25
Make label clickable

joecowton1:

@alinaboghiu pointed out in [this PR](https://github.com/guardian/dotcom-rendering/pull/3786) that the existing Source toggle was missing a clickable label and also contained a bug where the tick would appear too early. 

Swapping `rem` for `px` on the pseudo elements within the toggle also resolves an issue where the tick and inner circle would move across breakpoints.

## What does this change?

-  removes outer div
-  moves label to wrap button
- fixes bug with tick

## Screenshots

**Before**

https://user-images.githubusercontent.com/77005274/151030283-e6f2ec4c-2e09-4cf2-8f08-9cffc6d71b3f.mov

**After**

https://user-images.githubusercontent.com/77005274/151030333-164a6414-7bf4-45c5-b5e5-0c44332de500.mov

### Accessibility

### Responsiveness

### Documentation

-   [x] Examples in Storybook


.changeset/three-carrots-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-react-components-development-kitchen': minor
```
sndrs, line 2:
```suggestion
'@guardian/source-react-components-development-kitchen': patch
```
this should be a patch, but only because we havn't worked out how to release this properly yet (in proper semver terms this is a `minor` change, but this package can't really follow semver since any change could be breaking at any point). we should resolve this somehow

SiAdcock (approved):
Looks great, thanks @joecowton1 💎

vlbee:
@joecowton1 @SiAdcock 

I appreciate this has gone in already, but what's the rationale for a clickable label on a toggle and is it always desirable?

In our use case [here](https://www.figma.com/file/z74mPkMGzcATz3ZxrHetrs/Onboarding-UX?node-id=3449%3A51332) and [here](https://github.com/guardian/manage-frontend/pull/784), the labels are very long and there is a bigger gap between label/switch so not sure it makes sense to make the label itself clickable provide there is a chunky clickable area around the switch.  

I couldn't find any definitive guidance on this.

FWIW,  the label/row on the CMP banner toggle switch opens a drop down. 

image

PR #1251, opened 2022-01-27
Remove lerna

sndrs:

go all in on changesets

## What does this change?

- removes lerna
- updates the docs to reflect changesets


SiAdcock (approved):
Thanks for everything `lerna` 👋

PR #1254, opened 2022-01-31
Publish packages as discrete modules

sndrs:

stop publishing packages as monolithic files, to make them tree-shakeable

I was 100% under the impression that tree-shaking ES modules marked as side effect free combined with minification/dead-code elimination at the application end would get rid of anything consumers didn't use from the single files we're currently publishing. this is not true!

Here is a hello-world preact app that renders a button from react-components:

### Before

image (1)

### After
image (2)


SiAdcock (approved):
Amazing catch, great fix. Thanks @sndrs  👏

PR #1255, opened 2022-02-02
Remove toggle margin and make label display: flex

joecowton1:

Removes an unnecessary margin from the ToggleSwitch component so we don't need to subsequently override it in DCR. 

## What does this change?

-   Removes margin
-   Switches label `display: inline-flex` to `display: flex` so we add full-width borders above it

### Accessibility

### Responsiveness

### Documentation

-   [x] Examples in Storybook


joecowton1:
@SiAdcock @sndrs little nudge if you've time to look at this one please

SiAdcock (approved):

PR #1256, opened 2022-02-03
Downgrade storybook to 6.3 

SiAdcock:

Something in Storybook 6.4.14 is causing canvases rendered in iframes to not receive args for complex `argTypes` with [`mapping`](https://storybook.js.org/docs/react/writing-stories/args#mapping-to-complex-arg-values). For now, we should downgrade to a version of Storybook that isn't broken. 

## What does this change?

-   Downgrade Storybook packages to 6.3.13

## Screenshots

**Before**

Screenshot 2022-02-03 at 14 54 49

**After**

Screenshot 2022-02-03 at 14 55 30


SiAdcock:
On second thoughts this is a silly idea. We've created [a new repo](https://github.com/guardian/source-dac) just for a11y testing

PR #1263, opened 2022-02-08
Add optional labelId prop on ToggleSwitch

vlbee:

The Design Kitchen's Toggle Switch has a default label id of a hardcoded string "notify" 
This change introduced an optional prop to replace this, keeping "notify" as default.

This means when testing a React component with multiple switches, React Testing Library cannot match the right component purely by label text using the `.getByLabelText` query and you have to use `.getAllByLabelText` or a test id which is not as nice. See https://testing-library.com/docs/queries/about#priority. 

This optional prop allows for easier test writing. 

## What does this change?

No UI changes. 
Also fixed some spelling. 

## Screenshots

**Before**

**After**

### Accessibility

### Responsiveness

### Documentation

-   [ ] Examples in Storybook


joecowton1:
Wow there was a lot of typos 😳  If you could add a change set for this one too please that would be great!

packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/ToggleSwitch.tsx
```diff
@@ -83,12 +87,16 @@ export const ToggleSwitch = ({
 	};
 
 	return (
-		<label css={[labelStyles, cssOverrides]} {...props} id="notify">
+		<label
+			css={[labelStyles, cssOverrides]}
+			{...props}
+			id={labelId ?? 'notify'}
```
SiAdcock, line 93: Maybe we can use `generateSourceId` to generate a unique ID for this component if one is not passed.

```tsx
import { generateSourceId } from '@guardian/source-foundations'`

...

<label
			css={[labelStyles, cssOverrides]}
			{...props}
			id={labelId ?? generateSourceId()}
```

SiAdcock: Or what might be better is if the consumer of this component didn't have to care about the fact that there's a label.

```tsx
import { generateSourceId, descriptionId } from '@guardian/source-foundations'`

const ToggleSwitch = ({ id }) => {
  const buttonId = id || generateSourceId();
  const labelId = descriptionId(buttonId);

  return (
    <label id={labelId}>
      <button id={buttonId} aria-labelledby={labelId} />
   </label>
  )
```

.changeset/calm-tools-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-react-components-development-kitchen': patch
+---
+
+Add optional label Id prop on ToggleSwitch
```
SiAdcock, line 5:
```suggestion
Add optional id prop on ToggleSwitch
```

SiAdcock (approved):
I made a tiny suggestion, otherwise this is good to go! 👍

SiAdcock (approved):

.changeset/calm-tools-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-react-components-development-kitchen': patch
```
sndrs, line 2: This is new functionality so should probably be a minor?

PR #1282, opened 2022-02-24
Remove fragment from ChoiceCardGroup story template

SiAdcock:

There is a redundant fragment in the ChoiceCardGroup story template. Not only is this cluttering the code, it is also causing a bug which means it's possible to select multiple choice cards even in single-select mode.

## What does this change?

-   Remove redundant fragment


kenoir (approved):
🧹

PR #1285, opened 2022-02-25
Accessible icons

SiAdcock:

Icons are not currently announced by screen readers in an accessible way.

Icons receive keyboard focus in Internet Explorer which is usually undesirable.

## What does this change?

- Prevent icons receiving focus in IE using `focusable="false"`
- Hide SVG element from screen readers using `aria-hidden="true"`
- Add a visually hidden [icon label](https://www.figma.com/file/Ai7AELHC6KCz38qKZkvuHo/%E2%97%90-Icons?node-id=2810%3A6938) to every icon component
-   Add a prop to icon components, `isAnnouncedByScreenReader: boolean = false`, which determines whether icon's label is announced by screen readers
-   Respect `isAnnouncedByScreenReader` in icon stories

## Screenshots

**Before**

Screenshot 2022-02-25 at 15 08 09

**After**

Screenshot 2022-02-25 at 15 13 44 1

### Accessibility


packages/@guardian/source-react-components/src/icons/SvgCross.tsx
```diff
@@ -1,17 +1,36 @@
...
+			<span
+				css={css`
+					${visuallyHidden}
+				`}
+			>
+				Close
+			</span>
+		) : (
+			''
+		)}
```
SiAdcock, line 34: Do you have any thoughts on perhaps abstracting this?

```tsx
<>
  <svg>...</svg>
  { isAnnouncedByScreenReader ? <AccessibleLabel>Close</AccessibleLabel> : '' }
</>
```

packages/@guardian/source-react-components/src/icons/SvgCross.tsx
```diff
@@ -1,17 +1,36 @@
...
+	isAnnouncedByScreenReader = false,
+}: IconProps): EmotionJSX.Element => (
+	<>
+		<svg
+			viewBox="0 0 30 30"
+			xmlns="http://www.w3.org/2000/svg"
+			width={size ? iconSize[size] : undefined}
+			aria-hidden={true}
+			focusable={false}
+		>
```
SiAdcock, line 17: I wonder if it's possible / helpful to abstract away this repetition in every `<svg>` element?

```tsx
<>
  <IconSvg size={size}>
    <path> ... </path>
  <IconSvg>
  { 
    // label code ...
  }
</>
```

SiAdcock: @mxdvl do you have any thoughts on this from an automation perspective? My thinking is:

- `<path>` elements are the unique part of an SVG. The `<svg>` element is likely to be identical for every icon
- if we optimise each SVG via SVGO or similar, the most important part of the transformation is likely to be in the `<path>` elements. If SVGO optimises the `<svg>` element, we are likely going to apply that to all SVGs, not just one or two

tjsilver:
I have tested the icons in situ with a screen reader in `Button`, `ButtonLink`, `Link` and `LinkButton` and the SVG was not announced by the screen reader, as expected.

tjsilver (approved):

PR #1300, opened 2022-03-11
Fix get-icons scripts

SiAdcock:

`get-icons` uses the root level `tsconfig.json`, which specifies `esnext` as the output module format. The ESModule format is not recognised by Node, and so `ts-node` fails.

Some of the output paths are erreneous, and perhaps add too much hierarchy. We should look to correct and flatten them

## What does this change?

-   add a tsconfig.json to the get-icons package that specified as output module format of `commonjs`
-   fix some erroneous paths
-  flatten the `icons/components` directory to remove the `icons` subdirectory

## Out of scope

Here, we do not check in the output of running this script. We'll address this in a subsequent PR.


tjsilver (approved):
Paired on this and changes look sensible.

PR #1307, opened 2022-03-14
get-icons #4 - Populate existing icon components from Figma

SiAdcock:

In this change, existing icon components have been re-created using the [`get-icons` script](https://github.com/guardian/source/tree/main/scripts/get-icons).

New icons will be added in a subsequent PR

## What does this change?

- create a `get-icons` script in the root `package.json`
- ensure `get-icons` script creates icons with the correct name
- add accessible label to icons using `get-icons`
- adapt SVG viewboxes to ensure components appear 30x30, even though they are populated from 24x24 frames in Figma
- add a special viewbox size exception for the Direct Debit Wide icon, which should be 24x48
- re-create existing icons using `get-icons` script
- add accessible labels for each icon to this repo


packages/@guardian/source-react-components/src/icons/components/SvgCross.tsx
```diff
@@ -6,31 +6,33 @@ import type { IconProps } from '../types';
...
-				clipRule="evenodd"
-				d="M15.325 17.025l7.225 6.625 1.075-1.075-6.6-7.25 6.6-7.25L22.55 7l-7.225 6.625-7.25-6.6L7 8.1l6.625 7.225L7 22.55l1.075 1.075 7.25-6.6z"
-			/>
-		</svg>
-		{isAnnouncedByScreenReader ? (
-			<span
-				css={css`
-					${visuallyHidden}
-				`}
+}: IconProps): EmotionJSX.Element => {
```
tjsilver, line 9: The cross appears larger than the height of the letters in the buttons - especially in the default and small sizes - we should probably check with design if they're happy with that: 
https://www.chromatic.com/test?appId=5e933197d94d9c002275a15e&id=62330ead626d2d003a416a3c
image

https://www.chromatic.com/test?appId=5e933197d94d9c002275a15e&id=62330ead626d2d003a416a3d
image
https://www.chromatic.com/test?appId=5e933197d94d9c002275a15e&id=62330ead626d2d003a416a3e

image

scripts/get-icons/labels.ts
```diff
@@ -0,0 +1,55 @@
+export const labels: Record<string, string> = {
```
tjsilver, line 1: 💯

scripts/get-icons/index.ts
```diff
@@ -50,22 +53,27 @@ const getUrlsForNodes = (nodes: Node[]): Promise<NodeWithUrl[]> => {
...
+				formattedSvg.trimEnd(),
+			);
+			const titleCaseName = kebabToTitle(name);
+			const label = labels[name];
+			writeFileSync(`${SVG_OUTPUT_DIR}/${name}.svg`, formattedSvg);
 			writeFileSync(
-				`${REACT_COMPONENT_OUTPUT_DIR}/${titleCaseName}Icon.tsx`,
-				generateReactComponent(titleCaseName, formattedSvg.trimEnd()),
+				`${REACT_COMPONENT_OUTPUT_DIR}/Svg${titleCaseName}.tsx`,
+				generateReactComponent(name, componentSvg, label),
```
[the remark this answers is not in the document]
SiAdcock: Sadly not. The description field is already being used by designers for keyword searching.

packages/@guardian/source-react-components/src/icons/components/SvgCross.tsx
SiAdcock: Good spot. I think this is okay. Akemi deliberately increased the size of this cross (from 16px to 20px) when she migrated the icons into the 24x24 canvas.

This is how the button appears in Figma, in Akemi's comprehensive [Icon Test Page](https://www.figma.com/file/Ai7AELHC6KCz38qKZkvuHo/%E2%97%90-Icons?node-id=1211%3A1096):

Screenshot 2022-03-31 at 11 23 52

sndrs:
if the components are now auto-generated, do we want to be committing them? if i want to make a change to an icon i see the files there in the source code, it's not clear that anything i do in there will be overwritten next time the process runs.

could we run it on build and just pull in the latest each time? i guess that makes builds a bit risky since figma could contain breaking changes and an unrelated change would pull that in...

SiAdcock:
> could we run it on build and just pull in the latest each time? 

💡  Interesting idea! Let's explore...

> i guess that makes builds a bit risky since figma could contain breaking changes and an unrelated change would pull that in...

Yeah, we'd need to carefully manage changes to the icons file in Figma. It would need to be considered _live and dangerous_ by designers. Designers would have to get used to a kind of continuous deployment mindset.

I have some other concerns about this approach:

1. The build process would need to be aware of the current state of icons, and check for diffs after it runs. It would then need to automatically raise a changeset or otherwise correctly bump the version of `@guardian/source-react-components`. There are potentially a few weird edge cases we'd have to be aware of.
2. The build process would become dependant on the Figma API's availability. If the Figma API went down for a time, the build would fail until it's back up, even for non-icon changes.
3. Changes to icons should be infrequent, so there's a bit of inefficiency here. We'd always be adding 30-50s to the build for the sake of running a script that is rarely needed.

On the other hand, we would never need to think about icons again! 😄 

If we keep it like it is currently implemented, with a manual task that we need to run:

- We'll need to remember to run it after icon changes are made in Figma. Raise a pull request, make a changeset. It's faff and we might forget. Could this be tidied up a bit? Maybe we could create a manually-triggered GitHub action that runs the `get-icons` script and creates a pull request and placeholder changeset.
- We need to remember to never directly update the icon components in the source code. We could add a comment to the top of each module discouraging direct edits.

Despite these concerns, my feeling right now is to keep it like it is and see how things go. If we do hit the kind of problems we have identified (or any others!), then we can explore further automation solutions to fix them. But please do let me know what are your thoughts on this 🙏

SiAdcock:
## Todo

- Move SVGs to a vendor folder
- Automate generation of components at build time
- Add a readme to the icons directory saying what's happening

SiAdcock:
Superceded by #1344

PR #1310, opened 2022-03-17
Use `@deprecated` jsdoc tag

sndrs:
- adds the jsdoc `@deprected` flag, so that VS code can let you know if you use a deprecated feature
- deprecates the `Props` export (cc @oliverlloyd)

## Screenshot
image

### DIFF
--- a/packages/@guardian/source-react-components/src/hide/Hide.tsx
+++ b/packages/@guardian/source-react-components/src/hide/Hide.tsx
@@ -18,12 +18,16 @@ export interface HideProps extends HTMLAttributes<HTMLDivElement>, Props {
 	 * **Deprecated**
 	 *
 	 * Use `from` instead.
+	 *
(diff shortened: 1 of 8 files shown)


packages/@guardian/source-react-components/src/hide/Hide.tsx
```diff
@@ -18,12 +18,16 @@ export interface HideProps extends HTMLAttributes<HTMLDivElement>, Props {
 	 * **Deprecated**
 	 *
 	 * Use `from` instead.
+	 *
+	 * @deprecated Use `from` instead.
 	 */
 	above?: Breakpoint;
 	/**
 	 * **Deprecated**
```
sndrs, line 26: these one's are for storybook, which ignores the `@deprecated` jsdoc tag

image

SiAdcock (approved):
Nice! 💯

PR #1320, opened 2022-04-06
feat(lines): pure SVG patterns

### LINKED ISSUE #1319: Blurry `Lines` when zoomed
**Describe the bug**
As reported by @paperboyo:

> Them gradient lines always inconsistent widths on zoom in Firefox
> No idea if there is a way of working around this Firefox bug, but they look for me like this more often than not as they are now (always have):

**Expected behavior**

**Screenshots**
image

**Desktop (please complete the following information):**

-   OS: macOS
-   Browser: Firefox
-   Version: [e.g. 22]

**Additional context**
Browser is zoomed **115%**

mxdvl:

Use SVG elements natively to draw patterns, instead of `div`s with background images and embedded images.

This allows to keep AR’s current Content-Security Policy: https://github.com/guardian/dotcom-rendering/pull/4519

## What does this change?

Drop the background-image repeating css rules in favour of pure SVG patterns that can fill up to our largest breakpoint of 1300px.

There is no visual change.

-   Dotted Lines
-   Squiggly Lines
-   Dashed Lines
-   Straight Lines, Fixes #1319 

## Screenshots

image

image

image

image

### Accessibility

-   [X] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [X] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

### Responsiveness

### Documentation

-   [X] Examples in Storybook

### Known issues

SVGs patterns will “only” stretch to `breakpoint.wide`, or `1300px`.


packages/@guardian/source-react-components-development-kitchen/src/lines/SquigglyLines.tsx
```diff
@@ -0,0 +1,78 @@
...
+export const getHeight = (count: LineCount): number =>
+	thickness + gap * Math.max(count, 1);
+
+const d = [
+	`M 0 ${thickness / 2}`,
+	`q ${squiggliness} 0 ${wavelength / 4} ${amplitude / 2}`,
+	`t ${wavelength / 4} ${amplitude / 2}`,
+	`t ${wavelength / 4} -${amplitude / 2}`,
+	`t ${wavelength / 4} -${amplitude / 2}`,
+	`t 12 0`,
```
JamieB-gu, line 22: This has led me down a really interesting deep dive into how Bézier curves work, thank you 🙂.

I have some questions. None of them are blockers for this PR, more just to have an interesting conversation:

1. I may have misunderstood, but is your amplitude here actually 1.5 and these should all be `amplitude` rather than `amplitude / 2`? The amplitude is the difference between the equilibrium and the maximum value, rather than the difference between the maximum and the minimum. For a cosine wave (which is what I think you're drawing here?):

amplitude

2. What does the final `t 12 0` do? Given where you are in the curve at that point I'd expect that to draw a straight line from `(12, 0)` to `(24, 0)`. I think I'm missing something 🤔

3. Are the subsequent `t`s on lines 20, 21, and 22 needed? Can you get by with the first `t` on line 19 and then just coords after that? Or have you included them more for symmetry rather than necessity?

I'm not that knowledgeable about this so I may have misunderstood one or all of the above points 🙈. Let me know what you think 🙂

packages/@guardian/source-react-components-development-kitchen/src/lines/StraightLines.tsx
```diff
@@ -0,0 +1,52 @@
...
+}): EmotionJSX.Element => {
+	const height = getHeight(count);
+	const viewBox = `0 0 ${maxWidth} ${height}`;
+
+	const lines = [];
+	for (let index = 0; index < count; index++) {
+		lines.push(
+			<line
+				key={index}
+				x1={1}
```
JamieB-gu, line 30: Not `0`? Is this because the thickness of the line would result in the viewbox cropping the edge?

JamieB-gu (dismissed):
Really nice! 🎉 

I've asked some questions largely for my own interest in the methodology, not because I think there are any issues 👍

packages/@guardian/source-react-components-development-kitchen/src/lines/SquigglyLines.tsx
```diff
@@ -0,0 +1,78 @@
+import type { SerializedStyles } from '@emotion/react';
+import type { EmotionJSX } from '@emotion/react/types/jsx-namespace';
+import { breakpoints } from '@guardian/source-foundations';
+import type { LineCount } from './Lines';
+
+const wavelength = 12;
+const amplitude = 3;
+const thickness = 1;
+const gap = 3;
+const squiggliness = wavelength / 8;
```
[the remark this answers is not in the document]
mxdvl: 🦑

packages/@guardian/source-react-components-development-kitchen/src/lines/StraightLines.tsx
mxdvl:
```suggestion
				x1={0}
```

packages/@guardian/source-react-components-development-kitchen/src/lines/SquigglyLines.tsx
mxdvl: 1. Good catch! Will fix. I thought amplitude was from the tip of the mount to the bottom of the valley.
2. Not sure it’s needed, but I think it helped with pattern edged. Probably best to use `stroke-linecap="square"`!
3. There’s only one `q`, `t` are just coords. I don’t think SVGs allow to just string coordinate pairs without the `t` instruction

JamieB-gu: > There’s only one `q`, `t` are just coords. I don’t think SVGs allow to just string coordinate pairs without the t instruction

I think you might be able to do this:

`q` *dx1* *dy1* *dx* *dy* `t` *dx2* *dy2* *dx3* *dy3* *dx4* *dy4*

based on what the docs say about subsequent coordinate pairs for the `t` command:

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/d#quadratic_b%C3%A9zier_curve

JamieB-gu (dismissed):

sndrs (changes requested):
i think these need a bit of `aria-hidden="true"` and `focusable="false"`?

packages/@guardian/source-react-components-development-kitchen/src/lines/Lines.tsx
```diff
@@ -36,18 +35,42 @@ export interface LinesProps extends Props {
 export const Lines = ({
 	effect = 'straight',
```
sndrs, line 36: i think having these configurable as a prop in this way means you'll ship each style even if it's not used. i don't imagine we often need to change them once rendered? would they work as separate components?

mxdvl: Yes, they would work great as separate components. I’ve marked this as deprecated in favour of importing the exact `*Line` component needed.

sndrs: nice one 👍

sndrs (approved):

PR #1329, opened 2022-04-12
divx => devx

georgeblahblah:
## What does this change?

* Fix the email address for `devx`. I checked the Google Groups and couldn't find `divx@theguardian.com`, so I assume this should be `devx` - sorry if that's a wrong assumption!


sndrs:
`divx@` is an alias for `clientsideinfra@` :)

thanks thoguh!

PR #1330, opened 2022-04-13
bump `rollup-plugin-ts` to v2 and remove `rollup-plugin-dts`

sndrs:
updates the build to latest and simpler setup and should help #1113 pass

note that `rollup-plugin-ts@2` builds `.d.ts` files alongside the module files, so there is no `types` dir anymore


tjsilver (approved):
I notice that this doesn't result in any version bumps - is the changeset included as best practice?

sndrs:
it's a way of keeping changesets quiet, but you can also just not bother and ignore their comment, either is fine if you don't want a release

PR #1337, opened 2022-04-21
Use storybook-deployer

SiAdcock:

We are currently using the not-very-frequently-updated [`maxheld83/ghpages`](https://github.com/maxheld83/ghpages) Action to deploy Storybook to Github Pages. Recently this action started [failing](https://github.com/guardian/source/runs/6098708595?check_suite_focus=true) for unknown reasons, so we thought we'd take the opportunity to explore alternatives.

[`storybook-deployer`](https://github.com/storybookjs/storybook-deployer) is [recommended and maintained by Storybook](https://storybook.js.org/docs/react/sharing/publish-storybook#publish-storybook-to-other-services), so is likely a good migration target for this task.

## What alternatives did you consider?

As a drop-in replacement for maxheld83/ghpages, [`@jamesives/github-pages-deploy-action`](https://github.com/JamesIves/github-pages-deploy-action) looks promising. Although given some of the problems we've had when deviating from the Storybook Happy Path, it may be safer to align on a Storybook-specific solution.

Storybook [recommends using Chromatic](https://storybook.js.org/docs/react/sharing/publish-storybook#publish-storybook-with-chromatic) rather than GH Pages to host our Storybook. The main downside to this is that we'd need to consider folks who have bookmarked / linked to the old ghpages URL. 

## What does this change?

-   Install storybook-deployer
-   Create a new workflow to deploy Storybook using storybook-deployer
-   Remove old `cd.yml` workflow, which was previously used to deploy Storybook to GH Pages


.github/workflows/deploy-storybook.yml
```diff
@@ -14,10 +14,8 @@ jobs:
       - name: build
         run: |
           yarn install --frozen-lockfile
-          yarn validate
           yarn build:storybook:ci
       - name: deploy
+        run: yarn deploy:storybook -- --ci
```
sndrs, line 19: i think yarn passes these through automatically

```suggestion
        run: yarn deploy:storybook --ci
```

package.json
```diff
@@ -16,6 +16,7 @@
     "build:react-components": "yarn workspace @guardian/source-react-components build",
     "build:storybook": "build-storybook -o dist",
     "build:storybook:ci": "NODE_ENV=production build-storybook -o dist",
+    "deploy:storybook": "storybook-to-ghpages",
```
sndrs, line 19: is this something we want in the main scripts? does this mean i could run it from my machine too? somehow this feels like something you should only be able to do from github?

SiAdcock: It's hard to make sure that this can only be run from GitHub, but if we keep it out of scripts, it makes it less obvious 😄 👍

sndrs:
Should we remove the gh pages  action too?

sndrs (dismissed):

SiAdcock:
> Should we remove the gh pages action too?

Gads, I thought I'd done that! It must have snuck back in following a rebase 😅 I've now deleted it.

PR #1342, opened 2022-04-21
Add Nx and use it to manage `build` and `test` tasks

sndrs:
Adds [Nx](https://nx.dev/) and uses it to look after `build` and `test` tasks plus a new `e2e` task.

## Scope

This is a very [light-touch addition](https://nx.dev/getting-started/nx-core) of Nx. It doesn't use Nx plugins or [executors](https://nx.dev/executors/using-builders#using-executors--builders). 

My hunch is that we'd get more of Nx's benefits if we _did_ that, but that's more than a first pass's set of changes.

## Benefits

Uses Nx's cloud cache so that task output is cached and subsequent runs only do what they need to.

## Changes

Apart from adding Nx, this also includes some refactoring that was either required or inspired by Nx:

- there's only one `build` task
  - if nothing has changed for a set of packages, Nx will fetch the cached version so there's little benefit in manually separating them now
- jest unit tests (`test`) run per package
  - enables Nx to ignore tests if it can
- ESLint integration tests run as part of `test`
  - they now express direct dependency on the things they test, (which when you think about it, seems correct 😄)
  - this tells Nx it needs to build their target packages before the tests can run 
- the new `e2e` task replaces `validate-dist`
  - the actual tests have been moved to their respective packages (`src/index.test.ts`)
  - `index.test.ts` tests are run against `dist` in `e2e` and `src` in `test`
- CI workflow jobs have the same name as the Nx target/npm-script they run

cc @jamie-lynch @idrise @jamesgorrie @ashishpuliyel – dipping a toe in...


SiAdcock (approved):
Really looking forward to working with this! 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍

sndrs:
i rebased and squashed some commits into more meaningful chunks, but nothing changes code-wise since last review

SiAdcock (approved):

PR #1401, opened 2022-04-28
Only run chromatic tests when review is requested...

sndrs:
...and when somethign happens in `main` (since you can commit directly to `main` so we want to know that hasn't broken)

thanks to @coldlink for the suggestion


SiAdcock (approved):
💰

PR #1409, opened 2022-04-29
move eslint rules out from `lib`

sndrs:
- move the rules to their respective package
- does duplicate some stuff
  - prob worth untangling this a bit at some stage, i'm not sure how intertwined things need to be


SiAdcock (approved):
Nice. Perhaps we can come back and revisit this the next time we update these plugins

PR #1428, opened 2022-05-13
RFC: `<Header />` component API

SiAdcock:

There is a plan to standardise headers across the Guardian's web products.

This proposal outlines an API that tries to capture the areas of standardisation while being as flexible as possible for end users.

> _See [the visual designs](https://www.figma.com/file/V8u2r27TFdjAWNM3p84ypC/%E2%9C%AD-Header?node-id=1989%3A71918) for more context._

## What does this change?

- Add RFC for the Header component API

## Risks and uncertainty

-   The support buttons / thank you message is delivered by Automat
    -   What is the state of Automat? Will it continue to be maintained, or should we take the opportunity to simplify?
    -   Currently in support-dotcom-components. Should the component code live in Source?
-   Client side rendered consumers will have to ship multiple logos, or they would have to lazy load the correct SVG
-   How much can we hardcode?
    -   Logos
    -   Editions switch
    -   Support messaging
    -   Masthead links
        -   Do we hardcode the URLs or should platforms control the actual links?
-   Should Source be aware of / managing Islands?
    -   We need to avoid re-rendering the entire header on the client because Nav is pretty huge, and doesn’t need to be dynamic
    -   Should we be imposing the Island architecture on all platforms?
    -   Could we allow consumers to compose their own header
        -   Via props
        -   `<EditorialHeader masthead={<Island><Masthead /></Island>} />`


docs/rfcs/header-component-api.md
```diff
@@ -0,0 +1,418 @@
...
+        </EditorialHeader.EditionSwitch>
+    </BrandBar>
+    <Navigation />
+    <Subnav />
+</EditorialHeader>
+```
+
+#### Props
+
+-   `isSignedIn` _boolean_
```
SiAdcock, line 339: Should this be `showSupportButtons`? It feels more representative of what will be presented.

```suggestion
-   `showSupportButtons` _boolean_
```

docs/rfcs/header-component-api.md
```diff
@@ -0,0 +1,418 @@
...
+    -   whether to show the support buttons
```
SiAdcock, line 340: This is determined by a call to the identity API made client-side. If we pass it in as a prop, it will trigger a re-render of the whole header, which might not be ideal.

docs/rfcs/header-component-api.md
```diff
@@ -0,0 +1,418 @@
...
+-   `showSubscriberMessage` _boolean_
+    -   whether to show the Thank You message
+-   `edition` _string_ `'uk'` | `'au'` | `'us'` | `'international'`
+
+#### `<EditorialHeader.EditionSwitch />`
```
SiAdcock, line 345: As a first pass, should we hardcode this component into the `EditorialHeader`, rather than exposing it as part of the API? It seems unlikely that consumers would want to customise it or change it once implemented.

sndrs: we'd need to make sure it included all the tracking links etc, which feels like the wrong type of thing to keep in source?

sndrs: i wonder if it's a viable rule to exclude any business logic from our component, and always require state derived from business logic to be passed in?

docs/rfcs/header-component-api.md
```diff
@@ -0,0 +1,458 @@
...
+-   `label` _string_
+    -   display label of the `BrandBar.LinkList`
+-   `icon` _ReactNode_
+    -   icon that appears alongside the label
+-   `from` _Breakpoint_
+    -   the breakpoint at which the `BrandBar.LinkList` becomes visible
+-   `until` _Breakpoint_
+    -   the breakpoint at which the `BrandBar.LinkList` is no longer displayed
+
+### `<Navigation />`
```
oliverlloyd, line 128: I wonder if it is worth using 'pillar' in the naming here? It seems integral but that could just be my perspective

docs/rfcs/header-component-api.md
oliverlloyd: This last separation makes sense. I think it will be easier to use this component if the business logic is held locally (e.g., in DCR) and the thing that matters to Source, and what Source brings to the table, is presentation

docs/rfcs/header-component-api.md
```diff
@@ -0,0 +1,458 @@
...
+-   `onClick` _function_
+    -   logic that is executed when the `BrandBar.Link` is clicked
+-   `label` _string_
+    -   display label of the `BrandBar.Link`
+-   `icon` _ReactNode_
+    -   icon that appears alongside the label
+-   `from` _Breakpoint_
+    -   the breakpoint at which the `BrandBar.Link` becomes visible
+-   `until` _Breakpoint_
+    -   the breakpoint at which the `BrandBar.Link` is no longer displayed
```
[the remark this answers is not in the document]
SiAdcock: Yeah, I suspect we can apply the media query to the link directly rather than a parent div

docs/rfcs/header-component-api.md
```diff
@@ -0,0 +1,469 @@
...
+-   `showSubscriberMessage` _boolean_
+    -   whether to show the Thank You message
+-   `readerRevenueLinks` _{ subscribe: string; support: string; contribute: string }_
+    -   urls to assign to the support buttons
+
+## Risks and uncertainty
+
+-   The support buttons / thank you message is delivered by Automat
+    -   What is the state of Automat? Will it continue to be maintained, or should we take the opportunity to simplify?
+    -   Currently in support-dotcom-components. Should the component code live in Source?
```
TomPretty, line 456: I'd be happy to chat about how these currently work and also some of the changes we are planning on making. As a really quick roundup:

The copy is configured (as A/B tests) through the Reader Revenue Control Panel (by our colleagues in BRR). The components are currently in support-dotcom-components (although we plan to move them to DCR repo in the future). When a user lands on a page, a request is sent to the support-dotcom-components backend which decides which A/B test & variant the user is in and then responds with the required data (mainly just copy). The user then makes a separate request for the component itself (although this is the part we plan to remove, and instead just have the components as a package in the DCR repo). 

I'm not quite sure how or if we'd want to have this baked into the source component? 

Happy to chat more!

SiAdcock: Hi @TomPretty thanks for the explanation. I'm a fan of reducing the complexity here, so thanks for your efforts 😍

Where these components ultimately live is up for grabs, but DCR is a good starting point. We're building the entire header here to begin with, so that aligns us nicely.

The copy – if there's already value in making that configurable, then we should replicate that in Source. I'll make changes to the API to show the intention. I'm guessing the current API can be found in the [header module folder](https://github.com/guardian/support-dotcom-components/tree/main/packages/modules/src/modules/header) in SDC?

Do you have a timeline for moving these components into DCR?

docs/rfcs/header-component-api.md
```diff
@@ -0,0 +1,449 @@
...
+-   `icon` _ReactNode_
+    -   icon that appears alongside the label
+-   `from` _Breakpoint_
+    -   the breakpoint at which the `Masthead.Link` becomes visible
+-   `until` _Breakpoint_
+    -   the breakpoint at which the `Masthead.Link` is no longer displayed
+
+### `<Masthead.Menu />`
+
+A dropdown menu that is displayed in the `Masthead.Links` section of the header.
```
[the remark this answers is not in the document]
SiAdcock: Thanks @tjmw, it's useful to know what's coming down the pipe 😊  I'd like to chat with you more when we start to work on this menu, if that would be okay?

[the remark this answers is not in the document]
tjmw: Hey @SiAdcock yeah definitely happy to chat whenever's convenient! 👍

PR #1434, opened 2022-05-18
Update focus colour

tjsilver:

Updates the colour of the `focusHalo` from `#007CC4` (contrast ratio 4.48:1) to `#0077B6` (contrast ratio 4.86:1).

**Before**

image

**After**

image

### Accessibility

### Responsiveness

### Documentation

-   [ ] Examples in Storybook

### DIFF
--- a/.changeset/good-pillows-do.md
+++ b/.changeset/good-pillows-do.md
@@ -1,6 +1,6 @@
 ---
-'@guardian/source-foundations': minor
-'@guardian/source-react-components': minor
+'@guardian/source-foundations': major
+'@guardian/source-react-components': major
 ---
 
 Update focusHalo colour
--- a/.changeset/good-pillows-do.md
+++ b/.changeset/good-pillows-do.md
@@ -1,6 +1,0 @@
----
-'@guardian/source-foundations': major
-'@guardian/source-react-components': major
----
-
-Update focusHalo colour
--- a/.changeset/good-pillows-do.md
+++ b/.changeset/good-pillows-do.md
@@ -1,0 +1,6 @@
+---
+'@guardian/source-foundations': minor
+'@guardian/source-react-components': minor
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/good-pillows-do.md
```diff
@@ -0,0 +1,6 @@
+---
+'@guardian/source-foundations': major
+'@guardian/source-react-components': major
```
SiAdcock, line 3: Just regarding this, we’re assuming react-components is getting a bump because foundations is a peerDependency

SiAdcock: If this is the case: wouldn’t we also expect react-components-development-kitchen to receive a bump also? 🤔

tjsilver: Yes! In the changesets list in the box above, it shows a major bump to all packages... 
image

SiAdcock: Ohhh cool! I'm probably misunderstand what the `.changeset` markdown file is for, I'll do some more reading 😅

SiAdcock: So maybe we can remove `'@guardian/source-react-components': major` from this file? Assuming changesets is just "aware" that it needs to be bumped, we don't need to specify it in the changeset file?

SiAdcock: My suggestion would be to remove this from here, and then I think we're good to merge. What do you think?

SiAdcock: Aha. Changesets is picking up changes to `@guardian/source-react-components` because this package has changed vs the [baseBranch](https://github.com/guardian/source/blob/858ebd7c52feb4db4a0971ffe52f972ee16a9505/.changeset/config.json#L7)

SiAdcock (approved):
Nice

PR #1438, opened 2022-05-23
Replace deprecated colour tokens with palette

tjsilver:

This PR replaces uses of the deprecated colour tokens with the newer `palette` tokens for consistency.

## What does this change?

There should be no perceptible changes apart from the updated example in the Components README.

### DIFF
--- a/.changeset/clean-berries-cough.md
+++ b/.changeset/clean-berries-cough.md
@@ -1,0 +1,6 @@
+---
+'@guardian/source-foundations': major
+'@guardian/source-react-components': major
+---
+
+Replace deprecated colour tokens with palette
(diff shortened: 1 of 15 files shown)


.changeset/clean-berries-cough.md
```diff
@@ -0,0 +1,6 @@
+---
+'@guardian/source-foundations': major
+'@guardian/source-react-components': major
```
SiAdcock, line 3: Given there's no perceptible changes and this constitutes a refactor, should we consider these `patch` changes?

```suggestion
'@guardian/source-foundations': patch
'@guardian/source-react-components': patch
```

SiAdcock:
Should this go into the `v5` branch? Maybe it's dependant on changes in `ts-link-ally`

Then I'm guessing this requires a rebase in order to avoid the weird diffs in Chromatic

PR #1442, opened 2022-05-30
Increase spacing between choice cards and label

SiAdcock:

The choice card designs have 8px of space between the label and the choice cards. We should update the code implementation to match.

**Note:** There is more complexity here. Label, supporting text and an error message are all optional, but the spacing should be applied if any one of these is present.

## What does this change?

-   apply 8px top margin to choice card container if label, supporting text or error message is shown

## Screenshots

**Before**

Screenshot 2022-05-30 at 16 07 37
Screenshot 2022-05-30 at 16 07 27
Screenshot 2022-05-30 at 16 07 18

**After**

Screenshot 2022-05-30 at 16 06 41
Screenshot 2022-05-30 at 16 06 50
Screenshot 2022-05-30 at 16 07 01

### DIFF
--- a/packages/@guardian/source-react-components/src/choice-card/ChoiceCardGroup.tsx
+++ b/packages/@guardian/source-react-components/src/choice-card/ChoiceCardGroup.tsx
@@ -5,7 +5,13 @@ import { Children, cloneElement } from 'react';
 import type { Props } from '../@types/Props';
 import { Legend } from '../label/Legend';
 import { InlineError } from '../user-feedback/InlineError';
-import { fieldset, flexContainer, gridColumns, gridContainer } from './styles';
+import {
+	containerTopMargin,
+	fieldset,
+	flexContainer,
+	gridColumns,
+	gridContainer,
+} from './styles';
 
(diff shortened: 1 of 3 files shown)


packages/@guardian/source-react-components/src/choice-card/ChoiceCardGroup.tsx
```diff
@@ -68,6 +74,9 @@ export const ChoiceCardGroup = ({
 	...props
 }: ChoiceCardGroupProps): EmotionJSX.Element => {
 	const groupId = id ?? generateSourceId();
+	const applyContainerTopMargin = () => {
```
sndrs, line 77: this reads to me like it have a side affect. also, does it need to be a function?

how do you feel about this?

```js
const topMargin = (label && !hideLabel) || supporting || error ? containerTopMargin : '';
```

SiAdcock: Good shouts! Thanks 🙏

sndrs (approved):

PR #1444, opened 2022-05-30
enable turbosnap

sndrs:

“TurboSnap is an advanced Chromatic feature that speeds up builds for faster UI testing and review using Git and Webpack’s dependency graph. It identifies component files that change, then intelligently builds and snapshots only the stories associated with those components.”

https://www.chromatic.com/docs/turbosnap

## What does this change?

-   enables turbosnap
-   runs chromatic on the `push` github actions event, as per the docs
	- we changed this recently to reduce chromatic runs but it's not made a huge difference (we don't have many pushes to un-PRed branches)

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,9 +1,6 @@
 name: Chromatic
 on:
   push:
-    branches:
-      - main
-  pull_request:
 jobs:
   upload-storybook:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


sndrs:
linking to https://github.com/guardian/dotcom-rendering/pull/4768 for posterity

ashishpuliyel (approved):
This seems, almost suspiciously, easy.

SiAdcock (approved):
Let’s give it a shot! 🏹

PR #1459, opened 2022-06-10
Use nx to run `tsc`

sndrs:

speed up type checks, but using nX cache if possible

- use Nx to run `tsc`


sndrs:
let's look at this in the monorepo

PR #1427, opened 2022-05-13
Add `disabled` prop to `ToggleSwitch`

silvacb:
## What is the purpose of this change?
As part of the ticket on creating a NoJS version of the key events toggle (https://trello.com/c/F710fnsR/187-hide-the-filter-key-events-when-js-is-off), this PR introduces a `disabled` prop which disables the `ToggleSwitch` and implements disabled specific styling. 

## What does this change?
1) A`disabled` prop is added to `ToggleSwitch`.
2) Disabled specific styling is added to `styles.ts`.

## Video

https://user-images.githubusercontent.com/55602675/168302872-4baaa753-ec56-444f-8a4b-18cd9a934deb.mov

### Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [x] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

### Cross browser and device testing

-   [x] Tested with touch screen device

### Responsiveness

-   [x] Tested at all breakpoints
-   [x] Tested with with long text
-   [x] Stretched to fill a wide container

### Documentation

-   [x] Full API surface area is documented in the README
-   [x] Examples in Storybook

PR #1460, opened 2022-06-14
Proposal: NoJS ToggleSwitch

silvacb:
### Summary
This is a proposal for the implementation of `ToggleSwitch` when JS is disabled

### Proposed Solution
If JS is disabled and a user clicks on an existing toggle, a tooltip will appear displaying the message _Please turn on JavaScript to use this feature_ (other suggestions for the message welcome).
The tooltip is displayed on **focus**, which has been selected because the focus selector:
- can be used without JS
- does not require the tooltip to have a close button
- works across all platforms.

Other possibilities have been considered but abandoned:
1. Disable the toggle if no JS: abandoned because of design considerations (originally proposed in #1427)
2. Render a blank tag if no JS: abandoned because of content reflow issues
3. Get around no. 2 by adding a placeholder div instead of the toggle: decided the blank space would look odd

### Implementation 
Check if JS is disabled as suggested in the `AccordionRow` component.
Implement platform-specific styling of a tooltip that appears when a user with disabled JS tries to focus on a toggle. 

## What does this change?
-   adds a check to see if JS is enabled to the `ToggleSwitch` component
-   adds a tooltip element to the `ToggleSwitch` component
-   renders the tooltip if JS is disabled
-   adds `tooltipStyles` to styles.ts to implement platform-specific styling

### Before

https://user-images.githubusercontent.com/55602675/174097314-4051d1bb-bcab-4480-8337-fd0ba2f5b992.mov

### After

https://user-images.githubusercontent.com/55602675/174097612-0baf5a2d-3454-4809-9cad-7474c322e0b2.mov

## Comments?
We would like to know what you think about the proposed solution. Could it be useful or could you think of a better way to implement this? Are there any concerns regarding this proposal?

## Note
For the purposes of testing out the tooltip, we have set the height of the stories in `ToggleSwitch.stories.tsx` to 100px.

### Accessibility

### Responsiveness

### Documentation

-   [x] Examples in Storybook


SiAdcock (approved):
Sorry for slowness, I'm gonna approve. I think it's more important for us to unblock you on this now, and offer suggestions for improvements when we're closer to promoting ToggleSwitch out of the Kitchen and into stable Source. 

I might add further thoughts to this PR after this is merged 😊

SiAdcock (approved):

PR #1463, opened 2022-06-22
RFC: line in the sand disallowing potentially unusable components

sndrs:

Proposes a line in the sand that requires Source components to not display if they are not usable.


docs/rfcs/no-unusable-components-line-in-the-sand.md
```diff
@@ -0,0 +1,75 @@
...
+This would render:
+```html
+<form action="/a11y-settings" method="post">
+	<label for="darkmode-toggle">dark mode</label>
+	<input type="checkbox" id="darkmode-toggle" name="darkmode" checked="false" />
+	
+	<label for="hi_contrast-toggle">high contrast</label>
+	<input type="checkbox" id="hi_contrast-toggle" name="hi_contrast" checked="false" />
+	
+	<button>save settings</button>
```
SiAdcock, line 63:
```suggestion
	<button>{submitMessage}</button>
```

sndrs: this a block of html?

SiAdcock: Ha! Oh I had forgotten about HTML! Disregard 🤦

docs/rfcs/no-unusable-components-line-in-the-sand.md
```diff
@@ -0,0 +1,75 @@
+# No unusable Source components
+
+This proposes a line in the sand regarding how Source components should behave when their ideal version requires technologies that a client may not offer.
+
+## Example problem
+
+A toggle switch is similar to a checkbox, but unlike a checkbox, the state it sets should apply immediately (no submit button).
```
SiAdcock, line 7:
```suggestion
A Toggle Switch is similar to a Checkbox. Unlike a Checkbox, the state it sets should apply immediately i.e. a user does not need to click a Submit Button.
```

docs/rfcs/no-unusable-components-line-in-the-sand.md
```diff
@@ -0,0 +1,75 @@
...
+
+It needs javascript to achieve this. There is no obvious default widget to fall back to – it's not just a visual treatment of standard widget.
+
+If we server render it, users without javascript enabled (or who have not yet downloaded the bundle) will be presented with an unresponsive CTA.
```
SiAdcock, line 11:
```suggestion
If we server render it, [users without JavaScript enabled](https://kryogenix.org/code/browser/everyonehasjs.html) (or who have not yet downloaded the bundle) will be presented with an unresponsive Toggle Switch until the JavaScript arrives.
```

docs/rfcs/no-unusable-components-line-in-the-sand.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Proposal
+
+A new line in the sand for Source: 
+
+> Do not present a user with a component they cannot use.
+
+Given the example above, this could mean either:
+
+- not displaying it to users without javascript
```
SiAdcock, line 21:
```suggestion
- not displaying it to users until the JavaScript loads
```

docs/rfcs/no-unusable-components-line-in-the-sand.md
```diff
@@ -0,0 +1,75 @@
...
+- identifying and displaying a non-JS fallback
+
+Note that this *does not mandate fallbacks*. A team could decide that the component is an enhancement and not critical, and so they decide to just not show it to people without javascript.
```
SiAdcock, line 24: I'm unsure about this line, apologies if I'm missing something. Is the suggestion:

> We (Design System Team) don't need to create fallbacks for every component, because teams might not need to use the fallback. We only create a fallback when the component has a known critical role in a project that consumes Source.

?

docs/rfcs/no-unusable-components-line-in-the-sand.md
```diff
@@ -0,0 +1,75 @@
...
+
+But rendering it in an inoperative state would cross the line in the sand. 
+
+### Anti-patterns
+
+This proposal would not permit a "*please enable javascript to use this*" type message.
```
SiAdcock, line 30:
```suggestion
This proposal would not permit a "*Please enable JavaScript to use this*" type message.
```

docs/rfcs/no-unusable-components-line-in-the-sand.md
```diff
@@ -0,0 +1,75 @@
...
+
+Users should decide (or are unable to change) the environment Source components run in. Our job is to provide the best experience we can given what we encounter. 
+
+We should treat this approach as an analogue of "this site needs Chrome".
+
+> n.b. this is not always an anti-pattern – for example on an internal tool where can broaden the range of requirements
```
SiAdcock, line 36: Does this mean (for example) that we (P&E) are able to tell users of internal tools that they must enable JavaScript?

docs/rfcs/no-unusable-components-line-in-the-sand.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Example solution
+
+In the case of a toggle switch, we could create a component that provides optional props that would enable us to render a working fallback. 
+
+If they're provided we render the fallback and if we can, enhance/decorate it.
```
SiAdcock, line 42:
```suggestion
If they're provided we render the fallback. If JavaScript becomes available, we can enhance it.
```

docs/rfcs/no-unusable-components-line-in-the-sand.md
```diff
@@ -0,0 +1,75 @@
...
+
+If they're not provided, we only show the component once we can make it functional. In reality this probably means either rendered hidden on the server then hydrated and shown, or only rendered in the client.
```
SiAdcock, line 44:
```suggestion
If they're not provided, we only show the component once we can make it functional. In reality this means either:

 - render the component hidden on the server, then hydrate and show it on the client OR
 - only render the component on the client
```

docs/rfcs/no-unusable-components-line-in-the-sand.md
SiAdcock: It feels like we need to define standard vs enhanced and the various levels of support, e.g.:

1. Not shown – a component is not shown if it cannot be used in the environment
2. Standard – a minimally functional but stable component that works everywhere
3. Enhanced – a streamlined and delightful experience for modern browsers and environments

docs/rfcs/no-unusable-components-line-in-the-sand.md
```diff
@@ -0,0 +1,75 @@
...
+	<label for="hi_contrast-toggle">high contrast</label>
+	<input type="checkbox" id="hi_contrast-toggle" name="hi_contrast" checked="false" />
+	
+	<button>save settings</button>
+</form>
+```
+
+> Note that we would still style the elements to match the design.
+
+### Without fallback
```
SiAdcock, line 69:
```suggestion
### Without fallback

In this case, the component is not shown if it cannot be used in the environment.
```

docs/rfcs/no-unusable-components-line-in-the-sand.md
sndrs: i think it's limiting the scope of this doc to source components. i think another team building internal tooling could make a good case for requiring that, for example. but it's beyond the scope of this doc. i watned to make sure that this didn't read like a proposal of a universal doctrine

docs/rfcs/no-unusable-components-line-in-the-sand.md
sndrs: i think i'm suggesting that Source components should provide mechanism for fallbacks where they are needed, and consumer can choose whether they use them?

but i agree it's not clear – i think i've confused what should be in source, and what we could do elsewhere. i'll look at that again

docs/rfcs/no-unusable-components-line-in-the-sand.md
```diff
@@ -0,0 +1,80 @@
...
+
+If they're not provided, we only show the component once we can make it functional. In reality this means either:
+
+ - render the component hidden on the server, then hydrate and show it on the client OR
+ - only render the component on the client
+
+
+### With fallback
+```jsx
+<ToggleSwitchGroup action="/a11y-settings" onchange={applyA11ySettings} submitMessage="save settings">
```
ob6160, line 52: Specifying the presence of a fallback implicitly by the presence of the `action` and `submitMessage` props confused me as i was reading this. That said, I can't think of a better solution right now

docs/rfcs/no-unusable-components-line-in-the-sand.md
```diff
@@ -0,0 +1,80 @@
...
+> Do not present a user with a component they cannot use.
+
+Given the example above, this could mean either:
+
+- not displaying it to users until the JavaScript loads OR
+- identifying and displaying a non-JS fallback
+
+Note that this *does not mandate fallbacks*. A team could decide that the component is an enhancement and not critical, and so they decide to just not show it to people without javascript.
+
+But rendering it in an inoperative state would cross the line in the sand.
```
ob6160, line 26: Would we expect consumers of Source to adhere to this line in the sand for when they're composing components of their own? Or will we just uphold this standard within components that belong to the design system? I appreciate that this is difficult to control and is probably something that we do not want to do. Thought I'd pose the question anyway.

To spread awareness of this line, I'm wondering whether we should also think about how this is documented for the components where this is required (in code, storybook or elsewhere)

docs/rfcs/no-unusable-components-line-in-the-sand.md
ob6160, line 52: Specifying the presence of a fallback implicitly by the presence of the `action` and `submitMessage` props confused me as i was reading this. That said, I can't think of a better solution right now

ob6160 (commented):
seems like a very sensible proposal, i left a few thoughts on the doc — mainly around: how we document and capture when components have fallbacks.. and when they do how we make the fact that the user is enabling/using the fallback as transparent as possible

docs/rfcs/no-unusable-components-line-in-the-sand.md
sndrs: no this is strictly for source

ob6160 (approved):

sndrs:
@SiAdcock i've updated the text to make it expllictly about source (i hope) but it's made me wonder how much this should a line in the sand only for source?

shouldn't we avoid displaying inoperative UI everywhere? i.e. is a source RFC the right place for this?

PR #1464, opened 2022-06-23
Use Source's ESLint plugins in Source

SiAdcock:

There are compelling reasons to dogfood our own ESLint plugins:

- ensure Source is conforming to the standards and rules we place on its consumers
- make it easier to ensure these plugins are working as expected

## What does this change?

- Add `@guardian/eslint-plugin-source-foundations` and `@guardian/eslint-plugin-source-react-components` to ESLint config


tsconfig.json
```diff
@@ -27,6 +27,12 @@
       ],
       "@guardian/source-react-components-development-kitchen": [
         "packages/@guardian/source-react-components-development-kitchen/src/index"
+      ],
+      "@guardian/eslint-plugin-source-foundations": [
+        "packages/@guardian/eslint-plugin-source-foundations/src/index"
+      ],
+      "@guardian/eslint-plugin-source-react-components": [
+        "packages/@guardian/eslint-plugin-source-react-components/src/index"
```
SiAdcock, line 35: This is not strictly necessary for this change, but I thought I'd keep it in anyway as it follows the pattern of defining paths to packages in the root tsconfig. Happy to remove if it's confusing / harmful / useless 🙂

tjsilver (dismissed):
Very sensible! 👍

tsconfig.json
tjsilver: Makes sense to follow an existing pattern, I guess?

.github/workflows/ci.yml
```diff
@@ -14,6 +14,7 @@ jobs:
         with:
           cache: 'yarn'
       - run: yarn install --frozen-lockfile
+      - run: yarn build
```
sndrs, line 17: does this mean we need to build the lint packages before linting?

ob6160: i've just tested with a
`yarn clean`
and a
`yarn lint`
and it seems that doing a build doesn't seem to be necessary for our lint process

ob6160: I have removed it in my latest push

ob6160: ah — nevermind, I needed to build both of the dogfooded eslint plugins. as the others are unnecessary I pushed a change to the ci step that builds just those two:

```yml
 # Build just the two eslint plugins needed.
      - run: yarn nx run-many --target=build --projects='eslint-plugin-source-foundations,eslint-plugin-source-react-components'
```

.github/workflows/ci.yml
```diff
@@ -14,6 +14,8 @@ jobs:
         with:
           cache: 'yarn'
       - run: yarn install --frozen-lockfile
+      # Build just the two eslint plugins needed.
+      - run: yarn nx run-many --target=build --projects='eslint-plugin-source-foundations,eslint-plugin-source-react-components'
```
sndrs, line 18: why does this need to happen in CI but not locally?

ob6160: in that other thread, when i said: 
`"and it seems that doing a build doesn't seem to be necessary for our lint process"`
I was wrong.

It seems that you need to build both of these packages as the reference is to the local versions, not those hosted via npm — meaning that the `lint` operation will only work when they have been built and are usable

joecowton1:
Does anyone have any objections to this being merged as it is? @sndrs @ob6160 @tjsilver

ob6160:
> Does anyone have any objections to this being merged as it is? @sndrs @ob6160 @tjsilver

none from me

ob6160 (approved):

sndrs:
did we resolve the issue of needing to build these eslint configs now before being able to lint?

PR #1465, opened 2022-06-23
Stop `valid-foundations-import-path` rule being logged to console

jamesmockett:

Stops the `valid-foundations-import-path` rule from being logged to the console when running ESLint.

## What does this change?

-   Removes `console.log` and `inspect` import from this rule


sndrs (approved):
thank you! (sorry)

SiAdcock (approved):
Magic 🪄 💥

Thanks so much!

PR #1472, opened 2022-07-06
Prevent layout jumps during page rendering

kaelig:
>If you move from one page of a site without a scrollbar to another with a scrollbar, you’ll see a slight layout shift as things squeeze inward a bit to make room for the scrollbar. -- https://css-tricks.com/elegant-fix-jumping-scrollbar-issue/

There are [other possible fixes](https://aykevl.nl/2014/09/fix-jumping-scrollbar) for this performance issue, but I went for the same as the one used in guardian/frontend:

https://github.com/guardian/frontend/blob/2cb01d2b326603de6354db9d9b6fe856a8b7cb2c/static/src/stylesheets/base/_base.scss#L1-L2

### DIFF
--- a/packages/@guardian/source-foundations/src/utils/resets.ts
+++ b/packages/@guardian/source-foundations/src/utils/resets.ts
@@ -35,6 +35,7 @@ const defaults = `
     html {
         -moz-osx-font-smoothing: grayscale;
         -webkit-font-smoothing: antialiased;
+        overflow-y: scroll;
     }
     html, body {
         text-rendering: optimizeLegibility;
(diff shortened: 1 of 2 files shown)


packages/@guardian/source-foundations/src/utils/resets.ts
```diff
@@ -35,6 +35,7 @@ const defaults = `
     html {
         -moz-osx-font-smoothing: grayscale;
         -webkit-font-smoothing: antialiased;
+        overflow-y: scroll;
```
sndrs, line 38:
```suggestion
        // always show the vertical scroll bar to stop the page
        // jumping about when content is shorter than the viewport
        overflow-y: scroll;
```

PR #1475, opened 2022-07-06
Add TextDecorationThickness to typography

joecowton1:
## What is the purpose of this RFC?

Following discussion with @akemitakagi we're looking into ways to make `Link` component underlines on hover adapt to the font size.

## Adding a CSS property to the typography

This approach is used in the code of this RFC.

This suggestion uses the [text-decoration-thickness](https://caniuse.com/?search=text-decoration-thickness) CSS property to set the underline size. 

In keeping with existing typography style conventions, we've added new `underlineThickness` objects to define the `text-decoration-thickness` property for each font at each specific `font-size`.

```js
const titlepieceUnderlineThickness = {
	small: underlineThickness[4], //6px
	medium: underlineThickness[5], //6px
	large: underlineThickness[5], //7px
} as const;

const headlineUnderlineThickness = {
	xxxsmall: underlineThickness[1], //3px
	xxsmall: underlineThickness[2], //4px
	xsmall: underlineThickness[2], //4px
	small: underlineThickness[2], //4px
	medium: underlineThickness[3], //5px
	large: underlineThickness[4], //6px
	xlarge: underlineThickness[5], //7px
} as const;

const bodyUnderlineThickness = {
	small: underlineThickness[1], //3px
	medium: underlineThickness[1], //3px
} as const;

const textSansUnderlineThickness = {
	xxsmall: underlineThickness[0], //2px
	xsmall: underlineThickness[1], //3px
	small: underlineThickness[1], //3px
	medium: underlineThickness[1], //3px
	large: underlineThickness[2], //4px
	xlarge: underlineThickness[2], //4px
	xxlarge: underlineThickness[2], //4px
	xxxlarge: underlineThickness[3], //5px
} as const;

const underlineThicknessMapping = {
	titlepiece: titlepieceUnderlineThickness,
	headline: headlineUnderlineThickness,
	body: bodyUnderlineThickness,
	textSans: textSansUnderlineThickness,
};

```

### Using `text-decoration-thickness: auto`

This would avoid adding complexity to the code, but would delegate underline thickness to the browser.


packages/@guardian/source-foundations/src/typography/data.ts
```diff
@@ -17,6 +17,8 @@ const lineHeights = [1.15, 1.35, 1.5] as const;
 
 const fontWeights = [300, 400, 500, 700] as const;
 
+const underlineThickness = [2, 3, 4, 5, 6, 7] as const;
```
tjsilver, line 20: Considering the number of comments required after all the implemenations of underlineThickness[x], would it be worth storing them in an object, with the pixel thickness in the key e.g. something like ```{
"thickness2px": 2,
"thickness3px": 3, 
... etc
}```? It might make it easier to tell at a glance the exact thickness that is being applied, and would remove the need for comments. I realise this isn't necessarily consistent with the rest of the codebase though...

joecowton1: I did discuss this with @sndrs and he suggested something similar - we felt that maybe the whole structure of this file could be refactored in such a way that it removed the need for comments. 

Something along the lines of this may be simplest of all..

```js
const headlineUnderlineThickness = {
	xxxsmall: 3px,
	xxsmall: 4px,
	xsmall: 4px,
	small: 4px,
	medium: 5px,
	large: 6px,
	xlarge: 7px,
} as const;
```

For the sake of keeping things consistent while making these changes maybe it makes sense to get this merged, and then refactor the whole thing in this spirit in a separate PR?

tjsilver (approved):
Changes make sense, and a wider refactoring to make the easier to reason about without the need for comments can come separately, as discussed. Great work!

sndrs (dismissed):

joecowton1:
Ticket for refactoring created here: https://trello.com/c/mKvuU54M/991-refactor-typography-datats

Waiting on approval from @akemitakagi before merging.

PR #1477, opened 2022-07-11
Add 14px to Body font

joecowton1:

Adds `xsmall` (14px) to the `body` font.

This font style is used for standfirsts on cards on Dotcom fronts.

## Screenshots
Screenshot 2022-07-11 at 14 44 11


tjsilver (dismissed):
👍 Changes make sense. Nit: it might be good to have a little more explanation of the context/reason for this addition, for future travellers.

tjsilver (approved):
👍

PR #1482, opened 2022-07-14
Show palette colours in Chromatic

joecowton1:

We were hiding the palette colours behind dropdown's in the Storybook story. 

As a result of this (and the format of the `.mdx` story) we weren't diffing colour changes in Chromatic.

## What does this change?

Removes the dropdown so all colours are visible all the time and wraps the palette colours in `<Story />` tag which means they're included in Chromatic now.

We've also refactored the way the colour palette examples are mapped.

## Screenshots

**Before**
Screenshot 2022-07-14 at 15 46 40

**After**
Screenshot 2022-07-14 at 15 49 08


tjsilver (approved):
This is really useful, in light of all our recent palette changes!

.changeset/red-coins-bow.md
```diff
@@ -0,0 +1,6 @@
+---
+'@guardian/source-foundations': patch
+'@guardian/source-react-components': patch
+---
+
+Show palette colour changes in Chromatic diffs
```
sndrs, line 6: do we need to do a release for this?

.storybook/preview-head.html
```diff
@@ -369,7 +369,6 @@
 
 	.tokens__list {
 		flex-grow: 2;
-		margin: 0 40px auto -20px; /* -20px to override storybook default padding */
```
sndrs, line 372: do you know what this was for before? how does removing this change things?

packages/@guardian/source-foundations/src/colour/palette.stories.mdx
```diff
@@ -2,11 +2,11 @@ import { Meta } from '@storybook/addon-docs';
 import { palette } from './palette';
 
 <Meta
-	title="Packages/source-foundations/palette"
-	parameters={{ previewTabs: { canvas: { hidden: true } } }}
+	title="Packages/source-foundations"
+	parameters={{ previewTabs: { canvas: { hidden: true } }}}
```
sndrs, line 6: what was behind these changes?

packages/@guardian/source-foundations/src/colour/palette.stories.mdx
```diff
@@ -2,11 +2,11 @@ import { Meta } from '@storybook/addon-docs';
...
 />
 
-# `palette`
+# Palette
```
sndrs, line 9: since `palette` is a named export (not just a concept), we wanted to convey that by styling it as code. did it break somtehing?

.storybook/preview-head.html
```diff
@@ -399,7 +398,7 @@
 		border: 1px solid #333333;
 	}
 
-	.tokens__list__dropdown {
+	.tokens__list__item {
```
sndrs, line 401: i think this may have broken the styling of the deprecated items? 

image

joecowton1: I think the styling is broken because by default storybook applies it's own styles (font / spacing) to `.mdx` files. We may have to add our own to get it looking like the other stories.

.vscode/extensions.json
```diff
@@ -5,6 +5,7 @@
 	"styled-components.vscode-styled-components",
```
tjsilver, line 5: We should probably add `.vscode/` to `.gitignore`

joecowton1: Good call, added.

packages/@guardian/source-foundations/src/colour/palette.stories.tsx
```diff
@@ -0,0 +1,184 @@
+import type { Story } from '../../../../../lib/@types/storybook-emotion-10-fixes';
```
tjsilver, line 1: This file icon being different to the other docs looks a bit odd in the menu: 
image
Also, we have lost the heading and example documentation from the top of the file: 
image
Should we maybe separate the doc and the palette so that users can still view the documentation, and chromatic can happily test the palette colours?

packages/@guardian/source-foundations/src/colour/palette.stories.mdx
tjsilver: I agree, and it should ideally be reflected as such the menu/heading as lowercase.

.storybook/preview-head.html
joecowton1: Resolved this by adding the text styles manually.

packages/@guardian/source-foundations/src/colour/palette.tsx
```diff
@@ -92,13 +92,6 @@ const colors = {
 	],
 } as const;
 
-/**
- * [Storybook](https://guardian.github.io/source/?path=/docs/packages-source-foundations-palette--page) •
- * [Design System](https://theguardian.design/2a1e5182b/p/492a30-light-palette)
- *
- * Colour palette organised by type e.g. 'error', or pillar e.g. 'opinion'
- */
-
```
sndrs, line 101: if we remove this we'll lose vscode's intellisense tips for palette for consumers:

image

joecowton1: Reverted this change.

.changeset/red-coins-bow.md
joecowton1: Removed.

packages/@guardian/source-foundations/src/colour/palette.stories.tsx
joecowton1: This one is complex for a number of reasons. Let's discuss and report back actions here.

.storybook/preview-head.html
joecowton1: not sure what it's purpose was currently, but if we leave it in it pulls the category titles off the left of the viewport so i think it makes sense to remove.

Screenshot 2022-07-26 at 12 57 55

packages/@guardian/source-foundations/src/colour/palette.stories.tsx
joecowton1: Discussed in person, outcome was use `mdx` rather than `tsx` story which resolves above points, apart from the different icon. We think though it makes sense to add chromatic examples using the `<Story />` tag for other similar `source-foundations`.

.storybook/preview-head.html
sndrs: is this still needed? it will affect every story (i think it may have been for embedding in zeroheight – @SiAdcock?

joecowton1: IIUC it's only used in the `palette` story

packages/@guardian/source-foundations/src/colour/palette.stories.tsx
tjsilver: I'm still not getting the 'docs' view by default when I run locally - so no heading or example. I also don't see any styling on the sidebar. I wonder if there's a way to 'force' the docs view? 

image

.storybook/preview-head.html
sndrs: apologies - i didn't notice the classname it's restricted to

packages/@guardian/source-foundations/src/colour/palette.stories.tsx
tjsilver: Update: I tried upgrading storybook to latest but that didn't fix the issue.

sndrs: that's annoying - could you filter out props that begin with `__`? and then maybe add a comment explaining why so we can remove at some point when it's fixed we wonder why we're doing something so strange?

sndrs: i guess this is because we're doing something quite weird with the story

joecowton1: Yeah I think that makes sense. I'd been wondering if there was a way to add a type check but this is probably simpler esp if it's a know bug.

joecowton1: I was able to resolve this by specifying that we use `react-docgen` rather then `react-docgen-typescript` in `main.js`
https://github.com/storybookjs/storybook/issues/9832

tjsilver (approved):
Great work!

PR #1491, opened 2022-08-22
Delete strange file

SiAdcock:

I think this file (introduced in #1475) was created erroneously, please forgive me if it's serving some arcane purpose! 🙏

## What does this change?

-   Delete the `,` file


ob6160 (approved):
goodbye `,` — you will be missed 🙏

ob6160:
@SiAdcock not sure why there's a Chromatic diff with this change?

joecowton1:
> @SiAdcock not sure why there's a Chromatic diff with this change?

I think there are just some un-approved changes that have carried over from previous PR's. We should probably just approve them so they stop carrying over.

PR #1492, opened 2022-08-23
Render fonts in Storybook `typography` docs

joecowton1:
## What is the purpose of this change?

To surface the typography variations (fonts, font-weights, italics, line-heights etc) in the Storybook `typography` docs. 

By providing visual examples of each variation rendered as code we're able to use Chromatic to check for regressions.

Note:
In order to surface the font styles in Chromatic we have to wrap each story in a `<Story />` tag - this has the unwanted side-effect of adding each story as a sub-item in the sidebar. 

There is a long running but as yet unimplemented request to hide sidebar items here: https://github.com/storybookjs/storybook/issues/9209
 
## What does this change?

- render typography variations in `typography` Storybook

## Screenshots

**Before**
Screenshot 2022-08-23 at 16 22 55

**After**

Screenshot 2022-08-24 at 16 19 47

PR #1493, opened 2022-08-23
Line height storybook documentation

ob6160:
In this PR we provide four extra stories for our typography documentation in Storybook so any future changes to default line height will be captured by our Chromatic visual regression testing.

## What does this change?

- Introduces four new stories to the Storybook Typography documentation improved by @joecowton1 in #1492 

## Screenshots

<table>
<tr>
	<th>Text Sans</th>
	<th>Text Egyptian</th>
	<th>Guardian Headline</th>
	<th>Guardian Titlepiece</th>
<tr>
Screenshot 2022-08-25 at 09 45 33
	

<tr>
Screenshot 2022-08-25 at 09 45 42

<tr>
Screenshot 2022-08-25 at 09 45 46

<tr>
Screenshot 2022-08-25 at 09 45 38

</table>

A more general documentation section detailing the line height calculation and aliases was also added:
Screenshot 2022-08-25 at 17 12 54

### DIFF
--- a/.changeset/smooth-goats-move.md
+++ b/.changeset/smooth-goats-move.md
@@ -1,7 +1,5 @@
 ---
-'@guardian/source-foundations': major
+'@guardian/source-foundations': patch
 ---
 
-Change the default line-height for text-sans to regular instead of loose.
-
-Please check that this doesn't break your existing layouts that use text-sans with the default line height.
+Extends the Typography story so that we better capture changes to the lineHeight of our fonts
--- a/.changeset/smooth-goats-move.md
+++ b/.changeset/smooth-goats-move.md
@@ -1,0 +1,7 @@
+---
+'@guardian/source-foundations': major
+---
+
+Change the default line-height for text-sans to regular instead of loose.
+
+Please check that this doesn't break your existing layouts that use text-sans with the default line height.
--- a/.storybook/typographyRenderers.tsx
+++ b/.storybook/typographyRenderers.tsx
@@ -4,24 +4,26 @@ import {
 	headlineObjectStyles,
 	textSansObjectStyles,
 } from '@guardian/source-foundations/src';
-import { fontWeightMapping, lineHeightMapping } from '@guardian/source-foundations/src/typography/data';
+import { fontWeightMapping } from '@guardian/source-foundations/src/typography/data';
 import type {
 	Category,
 	FontScaleFunction,
 	FontWeight,
-	LineHeight,
 } from '@guardian/source-foundations/src/typography/types';
 
 type FontFunctions = {
 	[key in Category]: FontScaleFunction;
 };
 
-interface Props {
+interface FontStylesRendererProps {
 	fontName: string;
 	fontStyles: FontFunctions;
 }
 
-export const FontStylesRenderer = ({ fontName, fontStyles }: Props) => {
+export const FontStylesRenderer: React.FC<FontStylesRendererProps> = ({
+	fontName,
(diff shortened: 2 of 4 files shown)


ob6160:
[Fonts: Define a default line height for each font](https://trello.com/c/ZzfzHqL1/1056-fonts-define-a-default-line-height-for-each-font)

joecowton1:
This is great @ob6160 ! My only thought is whether there would ever be an instance where it would be useful to document the line height options available as `options` (and how to use these) outside of the font defaults?

.storybook/typographyRenderers.tsx
```diff
@@ -4,24 +4,26 @@ import {
...
 };
 
-interface Props {
+interface FontStylesRendererProps {
 	fontName: string;
 	fontStyles: FontFunctions;
 }
 
-export const FontStylesRenderer = ({ fontName, fontStyles }: Props) => {
+export const FontStylesRenderer: React.FC<FontStylesRendererProps> = ({
```
joecowton1, line 23: In Source / DCR we prefer not to use `React.FC`, details on why are in here if you're interested: https://kentcdodds.com/blog/how-to-write-a-react-component-in-typescript

.storybook/typographyRenderers.tsx
```diff
@@ -37,23 +39,20 @@ export const FontStylesRenderer = ({ fontName, fontStyles }: Props) => {
...
-				{lineHeight} {'->'} {value}
-			</li>
-		))}
-	</ul>
-);
+interface LineHeightRendererProps {
+	getFontStyles: FontScaleFunction;
+}
+
+export const LineHeightRenderer: React.FC<LineHeightRendererProps> = ({
```
joecowton1, line 46: See earlier comment on `React.FC`

.storybook/typographyRenderers.tsx
ob6160: 👍 makes sense - will change back

joecowton1:
> > This is great @ob6160 ! My only thought is whether there would ever be an instance where it would be useful to document which line height option ares available (and how to use these) outside of the font defaults? Maybe we still need to list these below the font styles as previously?
> 
> i agree - will reinstate the documentation of the line height options that are available as well.

Nice, yeah i think no harm in having both in font line heights and line height options documented

joecowton1 (dismissed):

.changeset/smooth-goats-move.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-foundations': patch
+---
+
+change the default line-height for textsans to regular instead of loose
```
SiAdcock, line 5: Would we consider this to be [a breaking change](https://docs.google.com/document/d/1Vi7m35w46F72z6n4fucsxm6i28xbmn2iMfPGM7zWGz8/edit#heading=h.xoh3rjy9fctr) (i.e. a major version bump)?

ob6160: yep — I'm about to push a change to make this a breaking change after asking @sndrs this morning. I wasn't sure about how we categorise these changes previously

joecowton1 (dismissed):

.changeset/smooth-goats-move.md
ob6160: I've swapped this back to a `patch` change, because we're going to decouple the change in line height and the Storybook changes because of a large diff

joecowton1 (approved):

PR #1496, opened 2022-08-30
Use storybook doc-block for color palette story

joecowton1:

Use the Storybook Color Palette doc block to render our palette colours: https://storybook.js.org/docs/react/writing-docs/doc-block-colorpalette

## What does this change?
-  Implement `<ColorPalette />` and `<ColorItem />` tags
-  remove old custom layout and corresponding styles from preview-head.html
-  use css selector to hide hex values

## Screenshots

**Before**
Screenshot 2022-08-30 at 12 50 23

**After**
Screenshot 2022-08-30 at 12 09 45


joecowton1:
Replaced by: https://github.com/guardian/source/pull/1498

PR #1497, opened 2022-08-30
Dev dependency upgrades (minor/patch versions, 30th Aug 2022)

ob6160:
We'd like to keep our dependencies up to date. This PR bumps versions for dev dependencies that can be upgraded which have a minor/patch change.

As part of this PR I attempted to upgrade `@storybook/*` but ran into some issues, so will tackle these packages separately.

`major` upgrades will be attempted in a separate series of PRs


package.json
```diff
@@ -34,15 +34,15 @@
     "trim": ">=0.0.3"
   },
   "devDependencies": {
-    "@babel/core": "^7.17.10",
-    "@babel/plugin-proposal-class-properties": "^7.14.5",
-    "@babel/preset-env": "^7.17.10",
-    "@babel/preset-typescript": "^7.14.5",
-    "@changesets/cli": "^2.19.0",
-    "@emotion/babel-plugin": "^11.3.0",
-    "@emotion/react": "11.0.0",
```
sndrs, line 43: since this is a peerdep, we should keep this pinned at the minimum version in the peerdep range

sndrs (changes requested):

package.json
ob6160: good spot - have pushed a change

sndrs (approved):

ob6160:
[Bump deps](https://trello.com/c/f8uCebjE/1060-bump-deps)

PR #1498, opened 2022-08-31
Create new colour palette wrapper

joecowton1:

Use a copy of the Storybook Color Palette doc block to render our palette colours: https://storybook.js.org/docs/react/writing-docs/doc-block-colorpalette

We can't use the original as it doesn't work with Storybook `<Story />` wrappers in `.mdx` stories, so we can't use it to regression test with Chromatic. 

The error is: `TypeError: Cannot read property 'fonts' of undefined`

See this issues for details: https://github.com/storybookjs/storybook/issues/12016

None of the fixes suggested here seem to work, but creating a version that uses Emotion rather than StyledComponents seems to get around the problem.

## What does this change?
-  Copies `<ColorPalette />` and `<ColorItem />` from Storybook src code
-  Implement `<ColorPalette />` and `<ColorItem />` tags
-  remove old custom layout and corresponding styles from preview-head.html

## Screenshots

**Before**
Screenshot 2022-08-30 at 12 50 23

**After**
Screenshot 2022-08-30 at 12 09 45


ob6160 (approved):
great work - a really impactful change 👏

PR #1500, opened 2022-08-31
Typography refactor | Simplify theme token data structure

ob6160:

We'd like to simplify the internals of our Typography infrastructure as it has been noted over time that it's becoming quite difficult for developers unfamiliar with the codebase to make modifications and, more generally, to understand what is going on. 

Before beginning this work, we spent some time digging into the reasons *why* developers might find it difficult to approach this part of our codebase and whittled it down to a few key areas that this sequence of changes aims to address.

This pull request covers the following subset of these areas:

* A level of indirection exists in our typography theme token definitions; making it hard to quickly find the value required because of a mapping from each value to an array containing the scale.
  - The file also seemed large with some unnecessary abstractions and comments describing the mapping instead of it being described statically in our code.
* We noticed that in some areas, the code could be better documented, both in terms of doc comments and self documentation

We plan to follow this PR with some further changes to how we internally compute and define the final font styles (specifically focusing on the `fs` method of legend and looking at how to simplify the distinction between `api.ts` and `index.ts` where the same font style object is exposed as a string and as an object).

## What does this change?

The set of changes captured in this PR seek to address the items raised in the list above.

### Flattening of the token data structure 

Each group of design tokens (fontWeights, lineHeights, textSizes, etc...) was defined as a two-level mapping; with a lookup pulling each value from an array of every possible value in the given scale.

To illustrate with an example. A lot of our mappings looked like this:

```ts
const fontSizes = [12, 14, 15, 17, 20, 24, 28, 34, 42, 50, 70] as const;

const titlepieceSizes = {
	small: fontSizes[8], //42px
	medium: fontSizes[9], //50px
	large: fontSizes[10], //70px
} as const;
```

We noticed that we've introduced a potentially confusing of abstraction here. Over time as the file grew, more font definitions were added, creating a scenario in which developers were not able to easily determine what the value for each of the size aliases was without an adjacent comment in place to stop the necessity to continually scroll back up to refer back to the `fontSizes` scale.

This mapping is liable to easily fall out of date over time and potentially cause real confusion due to simple human error, because we have no way of checking to see if the comments are correct.

To address this, we instead take the step of concretely defining the values for each size alias at the point of defining the mapping. The benefit of this approach is that the values are kept in the same place as the alias at the point of definition, reducing mental load on the developer when they need to understand/modify an element of the file.

A downside of this approach is that we lose our tight adherence to a well defined scale, as the `fontSizes` array ceases to exist. This can be remedied in the future at the level of our type checker, ensuring that any values entered into the data structure are valid and belong to the scale as defined in our design system.

### Reorganisation of objects and their exports

- Some objects were exported in a way which didn't allow us to see the references to them. We change the way we export some objects so that, in the editor, we can get a clearer picture of the use of each object.
- The mapping objects: `fontSizes`, `remFontSizes` and `availableFonts` were only used by `fs.ts` and have been moved into this file. 
- The type `AvailableFontsMapping` was moved to `types.ts`
- Three exported references to "mapping" were renamed to: `fonts`, `fontWeights`, and `lineHeights` now that the mapping relation has been removed in favour of defining the values as constants.

### DIFF
--- a/.changeset/twenty-onions-fetch.md
+++ b/.changeset/twenty-onions-fetch.md
@@ -1,0 +1,5 @@
+---
+'@guardian/source-foundations': patch
+---
+
+refactoring of internal typography definitions to directly map to a concrete value instead of pulling from an array of values; a general simplification of the typography data file, and of its exports
--- a/packages/@guardian/source-foundations/src/typography/data.ts
+++ b/packages/@guardian/source-foundations/src/typography/data.ts
@@ -6,8 +6,8 @@ import { pxToRem } from '../utils/px-to-rem';
  * These fit the following px size scale:
  * [12, 14, 15, 17, 20, 24, 28, 34, 42, 50, 70]
  */
-const pxTextSizes = {
-	textSansSizes: {
+export const pxTextSizes = {
+	textSans: {
 		xxsmall: 12,
 		xsmall: 14,
 		small: 15,
@@ -17,12 +17,12 @@ const pxTextSizes = {
 		xxlarge: 28,
 		xxxlarge: 34,
 	},
-	bodySizes: {
+	body: {
 		xsmall: 14,
 		small: 15,
 		medium: 17,
 	},
-	headlineSizes: {
+	headline: {
 		xxxsmall: 17,
 		xxsmall: 20,
 		xsmall: 24,
@@ -31,7 +31,7 @@ const pxTextSizes = {
 		large: 42,
 		xlarge: 50,
 	},
-	titlepieceSizes: {
+	titlepiece: {
 		small: 42,
 		medium: 50,
 		large: 70,
@@ -48,8 +48,8 @@ const pxTextSizes = {
  *
  * See {@link [pxToRem](../utils/px-to-rem.ts)} for more details.
  */
-const remTextSizes = {
-	textSansSizes: {
+export const remTextSizes = {
+	textSans: {
 		xxsmall: pxToRem(12),
 		xsmall: pxToRem(14),
 		small: pxToRem(15),
@@ -59,12 +59,12 @@ const remTextSizes = {
 		xxlarge: pxToRem(28),
 		xxxlarge: pxToRem(34),
 	},
-	bodySizes: {
+	body: {
 		xsmall: pxToRem(14),
 		small: pxToRem(15),
 		medium: pxToRem(17),
 	},
-	headlineSizes: {
+	headline: {
 		xxxsmall: pxToRem(17),
 		xxsmall: pxToRem(20),
 		xsmall: pxToRem(24),
@@ -73,7 +73,7 @@ const remTextSizes = {
 		large: pxToRem(42),
 		xlarge: pxToRem(50),
 	},
-	titlepieceSizes: {
+	titlepiece: {
 		small: pxToRem(42),
 		medium: pxToRem(50),
 		large: pxToRem(70),
@@ -134,20 +134,20 @@ export const underlineThickness = {
 } as const;
 
 // Pixel font size exports
-export const textSansSizes = pxTextSizes.textSansSizes;
+export const textSansSizes = pxTextSizes.textSans;
 
-export const bodySizes = pxTextSizes.bodySizes;
+export const bodySizes = pxTextSizes.body;
 
-export const headlineSizes = pxTextSizes.headlineSizes;
+export const headlineSizes = pxTextSizes.headline;
 
-export const titlepieceSizes = pxTextSizes.titlepieceSizes;
+export const titlepieceSizes = pxTextSizes.titlepiece;
 
 // Computed rem font size exports
 
-export const remTitlepieceSizes = remTextSizes.titlepieceSizes;
+export const remTitlepieceSizes = remTextSizes.titlepiece;
 
-export const remHeadlineSizes = remTextSizes.headlineSizes;
+export const remHeadlineSizes = remTextSizes.headline;
 
-export const remBodySizes = remTextSizes.bodySizes;
+export const remBodySizes = remTextSizes.body;
 
-export const remTextSansSizes = remTextSizes.textSansSizes;
+export const remTextSansSizes = remTextSizes.textSans;
--- a/packages/@guardian/source-foundations/src/typography/data.ts
+++ b/packages/@guardian/source-foundations/src/typography/data.ts
@@ -1,6 +1,3 @@
-// eslint-disable-next-line @typescript-eslint/no-unused-vars -- imported so it can be used for our doc comment
-import { pxToRem } from '../utils/px-to-rem';
-
 /**
  * Pixel size values for each font that we use in the design system.
  *
@@ -46,7 +43,7 @@ const pxTextSizes = {
  * The resulting scale that we draw from is:
  * [0.75, 0.875, 0.9375, 1.0625, 1.25, 1.5, 1.75, 2.125, 2.625, 3.125, 4.375]
  *
- * See {@link pxToRem} for more details.
+ * See {@link [pxToRem](../utils/px-to-rem.ts)} for more details.
  */
 const remTextSizes = {
 	titlepieceSizes: {
--- a/packages/@guardian/source-foundations/src/typography/data.ts
+++ b/packages/@guardian/source-foundations/src/typography/data.ts
@@ -1,3 +1,12 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars -- imported so it can be used for our doc comment
+import { pxToRem } from '../utils/px-to-rem';
+
+/**
+ * Pixel size values for each font that we use in the design system.
+ *
+ * These fit the following px size scale:
+ * [12, 14, 15, 17, 20, 24, 28, 34, 42, 50, 70]
+ */
 const pxTextSizes = {
 	textSansSizes: {
 		xxsmall: 12,
@@ -30,6 +39,15 @@ const pxTextSizes = {
 	},
 } as const;
 
+/**
+ * Relative font sizes, calculated from the pixel sizes above;
+ * using the pxToRem method.
+ *
+ * The resulting scale that we draw from is:
+ * [0.75, 0.875, 0.9375, 1.0625, 1.25, 1.5, 1.75, 2.125, 2.625, 3.125, 4.375]
+ *
+ * See {@link pxToRem} for more details.
+ */
 const remTextSizes = {
 	titlepieceSizes: {
 		small: 2.625,
@@ -45,12 +63,12 @@ const remTextSizes = {
 		large: 2.625,
 		xlarge: 3.125,
 	},
-	remBodySizes: {
+	bodySizes: {
 		xsmall: 0.875,
 		small: 0.9375,
 		medium: 1.0625,
 	},
-	remTextSansSizes: {
+	textSansSizes: {
 		xxsmall: 0.75,
 		xsmall: 0.875,
 		small: 0.9375,
@@ -60,7 +78,7 @@ const remTextSizes = {
 		xxlarge: 1.75,
 		xxxlarge: 2.125,
 	},
-};
+} as const;
 
 export const fonts = {
 	titlepiece: 'GT Guardian Titlepiece, Georgia, serif',
@@ -124,12 +142,12 @@ export const headlineSizes = pxTextSizes.headlineSizes;
 
 export const titlepieceSizes = pxTextSizes.titlepieceSizes;
 
-// Rem font size exports
+// Computed rem font size exports
 
 export const remTitlepieceSizes = remTextSizes.titlepieceSizes;
 
 export const remHeadlineSizes = remTextSizes.headlineSizes;
 
-export const remBodySizes = remTextSizes.remBodySizes;
+export const remBodySizes = remTextSizes.bodySizes;
 
-export const remTextSansSizes = remTextSizes.remTextSansSizes;
+export const remTextSansSizes = remTextSizes.textSansSizes;
--- a/packages/@guardian/source-foundations/src/typography/data.ts
+++ b/packages/@guardian/source-foundations/src/typography/data.ts
@@ -1,6 +1,6 @@
 import type { Category, FontWeight, FontWeightDefinition } from './types';
 
-const textSansSizes = {
+export const textSansSizes = {
 	xxsmall: 12,
 	xsmall: 14,
 	small: 15,
@@ -10,12 +10,14 @@ const textSansSizes = {
 	xxlarge: 28,
 	xxxlarge: 34,
 } as const;
-const bodySizes = {
+
+export const bodySizes = {
 	xsmall: 14,
 	small: 15,
 	medium: 17,
 } as const;
-const headlineSizes = {
+
+export const headlineSizes = {
 	xxxsmall: 17,
 	xxsmall: 20,
 	xsmall: 24,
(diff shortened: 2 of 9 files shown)


ob6160:
[Fonts: Refactor internal typography API in Source](https://trello.com/c/S4MTZj2k/1057-fonts-refactor-internal-typography-api-in-source)

packages/@guardian/source-foundations/src/typography/data.ts
```diff
@@ -1,243 +1,153 @@
...
-Object.freeze(availableFonts);
-Object.freeze(titlepieceUnderlineThickness);
-Object.freeze(headlineUnderlineThickness);
-Object.freeze(bodyUnderlineThickness);
-Object.freeze(textSansUnderlineThickness);
-Object.freeze(underlineThicknessMapping);
+export const remBodySizes = remTextSizes.bodySizes;
 
-export {
-	titlepieceSizes,
```
joecowton1, line 223: Is there a case here that listing all exports gives an overview of what's within the file? Not a strong opinion..

packages/@guardian/source-foundations/src/typography/data.ts
```diff
@@ -1,243 +1,153 @@
...
 
-const underlineThicknessMapping = {
-	titlepiece: titlepieceUnderlineThickness,
-	headline: headlineUnderlineThickness,
-	body: bodyUnderlineThickness,
-	textSans: textSansUnderlineThickness,
-};
+export const remHeadlineSizes = remTextSizes.headlineSizes;
 
-Object.freeze(titlepieceSizes);
```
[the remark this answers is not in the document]
ob6160: it's an interesting one, i wasn't 100% clear on what `Object.freeze()` does, here's the docs (just putting here for posterity):
> A frozen object can no longer be changed: new properties cannot be added, existing properties cannot be removed, their enumerability, configurability, writability, or value cannot be changed, and the object's prototype cannot be re-assigned. freeze() returns the same object that was passed in. [^1]

So, it makes the export *really safe* in the sense that any potential user of the object won't be able to modify it.

The thinking behind removing it in this PR is that—from a quick search of the codebase—this is the *only* theme token definition file where we do this. So the reasoning was that, if we don't require that level of safety for any other object that we export, why have the special case here? And - as far as I can tell - there isn't a strong reason to have this special case.

Whether we should use `Object.freeze` across the board is an interesting discussion, and one we should probably have separate to this PR, but as this was a one-off instance it seemed prudent to just remove the inconsistency.

[^1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze

packages/@guardian/source-foundations/src/typography/data.ts
ob6160: I agree - having the overview down below is really useful.

I changed it to a per-line export because with that system you can have an overview of the references to each object.. which is massively helpful when doing stuff like refactors, or any modification in general.


Here's a demo of the difference:
Screenshot 2022-09-05 at 14 44 59

Screenshot 2022-09-05 at 14 44 22

packages/@guardian/source-foundations/src/typography/typography.test.ts
```diff
@@ -1,9 +1,9 @@
 import {
-	fontMapping as fonts,
-	fontWeightMapping as fontWeights,
+	fonts,
```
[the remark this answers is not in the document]
joecowton1: Approve of this.

packages/@guardian/source-foundations/src/typography/data.ts
joecowton1: Yeah that nice to have visibility over references - I guess with this one again maybe consistency could be our guide.. if we use grouped exports elsewhere maybe its good to do that here? 

Happy with whatever the consensus is thouugh.

packages/@guardian/source-foundations/src/typography/fs.ts
```diff
@@ -1,13 +1,73 @@
...
-import type { FontStyle, FontWeightDefinition, Fs, Option } from './types';
+import type {
+	AvailableFontsMapping,
+	FontStyle,
+	FontWeightDefinition,
+	Fs,
+	Option,
+} from './types';
+
+const fontSizes = {
```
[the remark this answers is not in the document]
ob6160: https://github.com/guardian/source/pull/1525 will capture the work I’m  doing on the side to clean these up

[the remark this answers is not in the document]
ob6160: i've decided to backport a change from this PR to here as it's quite small to be able to remove `fontSizes` and `remFontSizes` completely from this file. Will provide more explanation in a new comment

packages/@guardian/source-foundations/src/typography/fs.ts
```diff
@@ -27,31 +67,27 @@ function getFontStyle(
...
-				: // @ts-expect-error -- the types actually overlap, see https://gist.github.com/mxdvl/5e31fd5b13670b6a41ddac6c65efeee4
-				  `${Number(remFontSizeMapping[category][level])}rem`;
+				? Number(pxTextSizes[category][level])
+				: `${Number(remTextSizes[category][level])}rem`;
 		const lineHeightValue: `${number}px` | number =
 			unit === 'px'
 				? // line-height is defined as a unitless value, so we multiply
 				  // by the element's font-size in px to get the px value
 				  `${
-						lineHeightMapping[lineHeight] *
```
ob6160, line 41: We are able to completely remove `lineHeightMapping` and `fontSizeMapping` by referencing the definition directly from data.ts

packages/@guardian/source-foundations/src/typography/data.ts
ob6160: agreed — there might be a way to have grouped exports and retain the references, will do a little digging

packages/@guardian/source-foundations/src/typography/data.ts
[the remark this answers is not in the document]
ob6160: yeah, it's an interesting one around a regression.. if we don't require this standard elsewhere it does make me lean on the side of removing it; unless there is a strong justification to keep it (in which case we should probably ask ourselves why it isn't in place anywhere else).

packages/@guardian/source-foundations/src/typography/data.ts
```diff
@@ -1,243 +1,153 @@
...
-	xxsmall: remFontSizes[0], //12px
-	xsmall: remFontSizes[1], //14px
-	small: remFontSizes[2], //15px
-	medium: remFontSizes[3], //17px
-	large: remFontSizes[4], //20px
-	xlarge: remFontSizes[5], //24px
-	xxlarge: remFontSizes[6], //28px
-	xxxlarge: remFontSizes[7], //34px
+/**
+ * Pixel size values for each font that we use in the design system.
```
sndrs, line 4: how would feel about typing these things? something like:

```ts
type TextSize = 12 | 14 // etc
type FontSizeName = 'xxsmall' | 'xsmall' // etc

type PxTextSize = Record<FontSizeName, TextSize>
type PxTextSizes = Record<string, PxTextSize>

export const pxTextSizes: PxTextSizes = {
	textSans: {
		xxsmall: 12,
		xsmall: 14,
    }
}
```

ob6160: Completely agree — I had this in the back of my mind while doing this work. I added a note about this to the PR description too. My plan was to add as a followup PR so this one doesn't get too big :)

SiAdcock (approved):
A blessed refactor 🙏 👼 💎

packages/@guardian/source-foundations/src/typography/data.ts
sndrs: i think it would be good to add them here, since they'll catch any accidental incorrect sizes etc in _this_ PR too

packages/@guardian/source-foundations/src/typography/data.ts
```diff
@@ -1,241 +1,153 @@
...
+ *
+ * The resulting scale that we draw from is:
+ * [0.75, 0.875, 0.9375, 1.0625, 1.25, 1.5, 1.75, 2.125, 2.625, 3.125, 4.375]
+ *
+ *
+ * See {@link [pxToRem](../utils/px-to-rem.ts)} for more details.
+ */
+export const remTextSizes = {
+	textSans: {
+		xxsmall: pxToRem(12),
```
sndrs, line 53: does this become safer doing something like this?

```suggestion
		xxsmall: pxToRem(pxTextSizes.textSans.xxsmall),
```

packages/@guardian/source-foundations/src/typography/data.ts
```diff
@@ -1,241 +1,153 @@
...
-	titlepiece: remTitlepieceSizes,
-	headline: remHeadlineSizes,
-	body: remBodySizes,
-	textSans: remTextSansSizes,
+/**
+ * Relative font sizes, calculated from the pixel sizes above;
+ * using the pxToRem method.
+ *
+ * The resulting scale that we draw from is:
+ * [0.75, 0.875, 0.9375, 1.0625, 1.25, 1.5, 1.75, 2.125, 2.625, 3.125, 4.375]
```
sndrs, line 46: is this line helpful to people? i'm worried it could become out of date without anyone noticing. do we need to know these numbers?

.changeset/twenty-onions-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-foundations': patch
```
mxdvl, line 2: Would this not be a feature? If so, it’s `minor`.

mxdvl (dismissed):
This looks a lot cleaner.

packages/@guardian/source-foundations/src/typography/data.ts
ob6160: nice point, i agree it is safer — it made me think of the point I raised in chat around using concrete values instead of computing them at all. would be great to get feedback on that idea too :)

packages/@guardian/source-foundations/src/typography/data.ts
ob6160: agreed, easily made out of date 👍

packages/@guardian/source-foundations/src/typography/data.ts
ob6160: 👍  will include in this PR

.changeset/twenty-onions-fetch.md
sndrs: i think it's more a internal refactor? the api shouldn't change (iiuc)

packages/@guardian/source-foundations/src/typography/data.ts
ob6160: yep 👍 subsequent PR makes sense

.changeset/twenty-onions-fetch.md
ob6160: i'll do a final check after we've resolved the other discussion points to make sure the api doesn't change

packages/@guardian/source-foundations/src/typography/data.ts
mxdvl: You will lose some information by dropping the `as const` assertion on `pxTextSizes`. If you type it as `PxTextSizes` you cannot then know that `xxxsmall` is not available on `headline`, for example.

One fix for this is to actually capture these in a test rather than via the compiler.

```ts
type sizes = [12, 14, 17,  /* … */];
type names = ["small", "medium", "large", /* … */ ];

test.each(Object.entries(pxTextSizes))("pxTextSizes has valid values for %s", (key, value) => {
	for (const [name, size] of Object.entries(value)) {
		expect(names).toContain(name);
		expect(sizes).toContain(size);
	}
})
```

ob6160: thanks for raising the PR - looks great!

sndrs: @mxdvl ah yeah that's a shame, i didn't realise that. (feels weird that you can't limit the values of a `const` assertion).

i wonder if that is more useful than pinning the pixels value down?

mxdvl: That's all down to how the TypeScript compiler works. You cannot have it both ways, unfortunately. 

Either an object has a generic shape `Record<string, string>`, or it has a specific implementation `{ a: 0 } as const`. In this case I think knowing the exact mappings is valuable. It does mean you have to tell the compiler to chill whilst performing your tests.

ob6160: I've merged your PR @mxdvl  - thanks again for the contribution

.changeset/twenty-onions-fetch.md
sndrs: the e2e tests should catch anything too - as long as `[source-foundations/src/index.ts](https://github.com/guardian/source/pull/1500/files#diff-4d48fdc382593a97a362bf40e2b2156e09271531b8776cb413d957fd1f422d10)` doesn't have any changes, should be good

packages/@guardian/source-foundations/src/typography/data.ts
ob6160: changed in: [`ac5e32f` (#1500)](https://github.com/guardian/source/pull/1500/commits/ac5e32f9f6a7e3f4ee4d2eb0ac6fc55fec01ab35)

packages/@guardian/source-foundations/src/typography/data.ts
```diff
@@ -1,241 +1,149 @@
...
-	medium: remFontSizes[3], //17px
-	large: remFontSizes[4], //20px
-	xlarge: remFontSizes[5], //24px
-	xxlarge: remFontSizes[6], //28px
-	xxxlarge: remFontSizes[7], //34px
+/**
+ * Pixel size values for each font that we use in the design system.
+ *
+ * These fit the following px size scale:
+ * [12, 14, 15, 17, 20, 24, 28, 34, 42, 50, 70]
```
joecowton1, line 7: Did we wonder about removing these commented values?

packages/@guardian/source-foundations/src/typography/data.ts
joecowton1: I'm a bit confused here by all the comments so apologies if this is irrelevant, but where we going to type some of these? I couldn't see the typings in the final diff?

packages/@guardian/source-foundations/src/typography/data.ts
ob6160: yep, missed this 👍

packages/@guardian/source-foundations/src/typography/data.ts
ob6160: as I understood it, the resolution for this in the end was to check that the values are correct by making an assertion in the unit tests that Max put together in: https://github.com/guardian/source/pull/1535

ob6160: I've added to the comments on each (px and rem) object so people are aware of these tests: [`d727d94` (#1500)](https://github.com/guardian/source/pull/1500/commits/d727d9434739954e505b749a32592ce060caca20#diff-9f53feda13d0210df1793144b9787a8f820113e01c8b97eddaa9e391622d60eaR6)

joecowton1: Nice one!

joecowton1 (approved):

mxdvl (approved):

packages/@guardian/source-foundations/src/typography/data.ts
```diff
@@ -1,241 +1,152 @@
...
-	small: remFontSizes[2], //15px
-	medium: remFontSizes[3], //17px
-	large: remFontSizes[4], //20px
-	xlarge: remFontSizes[5], //24px
-	xxlarge: remFontSizes[6], //28px
-	xxxlarge: remFontSizes[7], //34px
+/**
+ * Pixel size values for each font that we use in the design system.
+ *
+ * We assert that the values match those we expect in our
```
sndrs, line 6:
```suggestion
 * We assert that the values match the guardian type scale in our
```

sndrs (approved):

PR #1525, opened 2022-09-05
Typography refactor | fs method and related types/objects

ob6160:
## Background

The purpose of the `fs` method as it currently stands isn't very clear, this PR refactors it so it is self documenting and simplifies its usage, related configuration objects and types.

## Change summary:

* Renames `fs` to `fontStyleFunction` for clarity
* Doc comment included with a usage example
* [`availableFonts`](https://github.com/guardian/source/pull/1525/files#diff-0d8826eef2671fa5bfd4f50f4f6a2498c6dd44255c297e042084c173431522b4L17), was broken down into two objects to define supported [font weights and italics separately](https://github.com/guardian/source/pull/1525/files#diff-9f53feda13d0210df1793144b9787a8f820113e01c8b97eddaa9e391622d60eaR107)
* Refactor of how we use `fontStyleFunction` so that we don't have to nest within a container function at the point of definition in api.ts, really aiding readability in [this file](https://github.com/guardian/source/pull/1525/files#diff-08eac9baba951f548f754aa137c47aed2160c1f53150740dc2f73becbef43c7eR31). Going from this:
```typescript
xxsmall: (options?: FontScaleArgs) =>
		textSansFs('xxsmall', Object.assign({}, textSansDefaults, options))
```
to this:
```typescript
xxsmall: fontStyleFunction('textSans', 'xxsmall', textSansDefaults)
```
* Refactor of how we define the font styles as string methods in [`index.ts`](https://github.com/guardian/source/blob/ob/typography-refactor-fs/packages/@guardian/source-foundations/src/typography/index.ts) — the `objectStylesToString` method is moved and renamed to `fontStyleToStringFunction` so its purpose is clear, and it is localised to the core `fontStyleFunction` method. Similar to above, we go from this:
```typescript
xsmall: (options?: FontScaleArgs) =>
	objectStylesToString(bodyAsObj.small(options)),
```
to this:
```typescript
xsmall: fontStyleToStringFunction(bodyAsObj.small),
```


packages/@guardian/source-foundations/src/typography/api.ts
```diff
@@ -1,110 +1,68 @@
...
-		textSansFs('medium', Object.assign({}, textSansDefaults, options)),
-	large: (options?: FontScaleArgs) =>
-		textSansFs('large', Object.assign({}, textSansDefaults, options)),
-	xlarge: (options?: FontScaleArgs) =>
-		textSansFs('xlarge', Object.assign({}, textSansDefaults, options)),
-	xxlarge: (options?: FontScaleArgs) =>
-		textSansFs('xxlarge', Object.assign({}, textSansDefaults, options)),
-	xxxlarge: (options?: FontScaleArgs) =>
-		textSansFs('xxxlarge', Object.assign({}, textSansDefaults, options)),
+	xxsmall: fontStyleFunction('textSans', 'xxsmall', textSansDefaults),
```
joecowton1, line 60: Much easier to read!

packages/@guardian/source-foundations/src/typography/font-styles.ts
```diff
@@ -0,0 +1,89 @@
...
+	<
+		Category extends keyof Categories,
+		Level extends keyof Categories[Category],
+	>(
+		category: Category,
+		level: Level,
+		defaults: TypographyOptions,
+	) =>
+	(options?: FontScaleArgs): TypographyStyles => {
+		const finalOptions = {
```
joecowton1, line 45: could this be something like `fontSpec` or similar?

packages/@guardian/source-foundations/src/typography/data.ts
```diff
@@ -101,6 +105,20 @@ export const fontWeights = {
 	bold: 700,
 } as const;
 
+export const fontWeightsAvailable = {
```
joecowton1, line 108: I think something about this naming makes it harder to grasp whats going on here :) could this be `AvailableFontWeights` and the type called something else ie `AvailableFontWeightsType` or something?

packages/@guardian/source-foundations/src/typography/types.ts
```diff
@@ -48,12 +48,51 @@ type Categories = {
...
-	};
+export type AvailableFontWeights = {
+	[cat in Category]?: { [weight in FontWeight]?: boolean };
 };
 
+export type ItalicsAvailableForFontWeight = {
+	[cat in Category]?: { [weight in FontWeight]?: boolean };
+};
+
+export type TypographyOptions = {
```
joecowton1, line 59: Again i wonder if using `spec` rather than `options` makes mores sense? I'm thinking of `options` in terms of what the consumer uses to change the spec rather than the spec itself

packages/@guardian/source-foundations/src/typography/data.ts
joecowton1: Also for my own interest were these defined previously or is this something we've added in this PR?

packages/@guardian/source-foundations/src/typography/types.ts
```diff
@@ -48,12 +48,51 @@ type Categories = {
...
+> & {
+	fontStyle?: FontStyle;
+};
+
+export type FontScaleFunction = (options?: FontScaleArgs) => TypographyStyles;
+
+// returns styles as a template literal
+export type FontScaleFunctionStr = (options?: FontScaleArgs) => string;
+
+export type TitlepieceFunctions = {
```
joecowton1, line 77: Could we define this as a common type? ie:
```ts
type TypographyFunction<T> = {
	[key in keyof T]: FontScaleFunction;
};

export type HeadlineFunctions = TypographyFunction<HeadlineSizes>

export type TitlepieceFunctions = TypographyFunction<TitlepieceSizes>
```
and so on..

packages/@guardian/source-foundations/src/typography/data.ts
ob6160: good point, I agree about the name change 👍 

they (and the italic object below) were both defined previously as part of this object: https://github.com/guardian/source/pull/1525/files#diff-9f53feda13d0210df1793144b9787a8f820113e01c8b97eddaa9e391622d60eaL124

it was kind of a "two in one" scenario, where the presence of the font weight in that object meant that it was supported *and* then the italic boolean was defined a level below. 

so, this was an attempt to separate the two out and make it a bit clearer what the intent was. that said, I'm still not very happy with this solution - going to have a very quick play around with the shape of it to see if I can come up with a clearer way of expressing this

packages/@guardian/source-foundations/src/typography/types.ts
ob6160: agreed - what d'you think about `TypographyConfiguration`?

packages/@guardian/source-foundations/src/typography/types.ts
ob6160: nice, that's a lot cleaner

packages/@guardian/source-foundations/src/typography/font-styles.ts
ob6160: i've put in a change to `finalFontConfiguration`, let me know if that's clearer - or if `spec` (i'm guessing for "specification"?) would be clearer

packages/@guardian/source-foundations/src/typography/types.ts
ob6160: https://github.com/guardian/source/pull/1525/files#diff-ecf8ea629817e204e3ed85acdc26e07fb6524496ff250307ddfede20929a16d4R82-R88

packages/@guardian/source-foundations/src/typography/data.ts
ob6160: couldn't think of a cleaner way to do this in the time box I gave myself — so i've renamed them as you suggested :)
Change commit here: [`d3bc35f` (#1525)](https://github.com/guardian/source/pull/1525/commits/d3bc35fb80b0d19c12b6b093992b63e2ab3f168c)

packages/@guardian/source-foundations/src/typography/data.ts
```diff
@@ -101,6 +105,20 @@ export const fontWeights = {
...
+	titlepiece: { bold: true },
+	headline: { light: true, medium: true, bold: true },
+	body: { regular: true, bold: true },
+	textSans: { regular: true, bold: true },
+} as AvailableFontWeights;
+
+export const italicsAvailableForFontWeight = {
```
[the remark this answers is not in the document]
ob6160: couldn't think of a cleaner way to do this in the time box I gave myself — so i've renamed them as you suggested :)
Change commit here: [`d3bc35f` (#1525)](https://github.com/guardian/source/pull/1525/commits/d3bc35fb80b0d19c12b6b093992b63e2ab3f168c)

packages/@guardian/source-foundations/src/typography/api.ts
```diff
@@ -1,110 +1,69 @@
...
-type BodyFunctions = {
-	[key in keyof BodySizes]: FontScaleFunction;
+export const headline: TypographyFunctions<HeadlineSizes> = {
+	xxxsmall: fontStyleFunction('headline', 'xxxsmall', headlineDefaults),
+	xxsmall: fontStyleFunction('headline', 'xxsmall', headlineDefaults),
+	xsmall: fontStyleFunction('headline', 'xsmall', headlineDefaults),
+	small: fontStyleFunction('headline', 'small', headlineDefaults),
+	medium: fontStyleFunction('headline', 'medium', headlineDefaults),
+	large: fontStyleFunction('headline', 'large', headlineDefaults),
+	xlarge: fontStyleFunction('headline', 'xlarge', headlineDefaults),
```
tjsilver, line 38: Just wondering if there's a way to iterate over the sizes, set the key to each size and pass each size to the `fontStyleFunction` without having to write each one out individually?

ob6160: good idea! I'll have a go at implementing that tomorrow and we can compare the two approaches.
I think it'll work well, but there's a small part of me that kinda likes how declarative setting them out individually is.

My hope is that with both of them side by side it'll be an easier comparison to make :)

ob6160:
[Fonts: Refactor internal typography API in Source](https://trello.com/c/S4MTZj2k/1057-fonts-refactor-internal-typography-api-in-source)

packages/@guardian/source-foundations/src/typography/api.ts
ob6160: @tjsilver I had a go at doing this in: [this commit: `af1ea2e` (#1525)](https://github.com/guardian/source/pull/1525/commits/af1ea2e00d1a04f7d6ecda7676556b60cda2ab7e) 

it has the added benefit of adding a strong coupling between `index.ts` and `api.ts`, meaning that if api.ts is updated, this file will reflect those changes too.

The method: `getStringStyleFunctions` takes a bit of headspace to read though, I added a doc comment which should help to explain what it does, but defining line by line does seem a bit clearer.

I'm still on the fence, what d'you think?

tjsilver: I'm a bit confused, looks like your new `getStringStyleFunctions` works on the `fontStyleToStringFunction` in index.ts (which is also great), but the implementation of `fontStyleFunction` in api.ts (which I was referring to above) is left unchanged? Or is that the two options being shown together? 

Also, I see what you mean about the `getStringStyleFunctions` being a bit gnarly - could a map be used here instead of reduce, for clarity?

ob6160: oh - sorry, i completely missed that your comment was in this file and not index.ts - let's discuss on that call

SiAdcock (approved):
Great changes, so much clearer! 🙏💎

packages/@guardian/source-foundations/src/typography/api.ts
ob6160: update:
we discussed the feasibility of defining the string and object style functions in the same file for simplicity, but the refactoring work would have taken longer than the time box we had decided upon.

tjsilver (approved):
This is a great improvement!

packages/@guardian/source-foundations/src/typography/data.ts
joecowton1: Great, that works nicely now i think.

packages/@guardian/source-foundations/src/typography/types.ts
joecowton1: Nicely done!

packages/@guardian/source-foundations/src/typography/data.ts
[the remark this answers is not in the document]
joecowton1: Perfect.

packages/@guardian/source-foundations/src/typography/font-styles.ts
joecowton1: Yeah i think thats fine.

packages/@guardian/source-foundations/src/typography/types.ts
joecowton1: Yeah sounds good to me.

joecowton1 (changes requested):
Excellent work, think this had made a real difference.

joecowton1 (approved):

PR #1529, opened 2022-09-07
Remove `get-usage`

sndrs:
we dont run or maintain it. it should live in tracker really, but having it here adds overhead for linting/ts updates even though we don't ever run it.


ob6160 (approved):
looks good!  i've not come across `get-usage` before, what was it doing when it was still in place? does it live in its own repo now?

sndrs:
it tracked the usage of source in our github repo, iirc. really it should live in tracker and be automated. @tjsilver is looking into that currently.

maybe this _should_ live here, but currently it's cobwebby and unused. we can always revert this PR if we decide that in the future

PR #1531, opened 2022-09-08
Make `textSans` default line-height "regular"

joecowton1:

We'd like to set a default line height for each main font that we provide in Source.

Text Sans has had a line height of 150 as a default for many years. In design, we mostly use Text Sans with a line height of 135.

This work is centred around changing that line height and managing the knock-on effects of doing so. We'd like to have as little impact on the external teams as possible.

## What does this change?

- Updates textSans line-height default to `regular` 
- Removes overrides which were setting line-height to `regular` on a per-component basis
- Removes additional padding in `Button` component which was added to correct layout with incorrect line-height
- Removes additional margin which was used to correct alignment in `Summary` component

## UI Changes

The intention behind this PR is that we should update the line height but not cause any UI changes. See Chromatic for a full list of components effected.


packages/@guardian/source-react-components/src/button/styles.ts
```diff
@@ -100,23 +100,12 @@ const subdued = (
 	border-radius: 0;
 `;
 
-/*
```
joecowton1, line 103: This was correcting the position of the text, setting the line-height to 1.35 rather than 1.5 removes the need for this code.

packages/@guardian/source-react-components-development-kitchen/src/summary/styles.ts
```diff
@@ -12,7 +12,6 @@ export const wrapperStyles = (color: string): SerializedStyles => css`
 export const iconStyles = (color: string): SerializedStyles => css`
 	display: flex;
 	flex: 0 1 auto;
-	margin-top: 1px;
```
joecowton1, line 15: This is causing issues with Chromatic diffs but seems the correct thing to do here.

joecowton1: Doesn't work with Chromatic testing, have reverted to the old approach.

packages/@guardian/source-react-components-development-kitchen/src/summary/styles.ts
```diff
@@ -24,7 +24,10 @@ export const messageStyles = (
 	color: string,
 	isBold = true,
 ): SerializedStyles => css`
-	${textSans.medium({ fontWeight: isBold ? 'bold' : 'regular' })}
+	${textSans.medium({
+		fontWeight: isBold ? 'bold' : 'regular',
+		lineHeight: 'loose',
```
joecowton1, line 29: Overriding the line height here so we don't break existing styling.

.changeset/cool-deers-chew.md
```diff
@@ -0,0 +1,9 @@
+---
+'@guardian/eslint-plugin-source-foundations': major
+'@guardian/eslint-plugin-source-react-components': major
+'@guardian/source-foundations': major
+'@guardian/source-react-components': major
+'@guardian/source-react-components-development-kitchen': major
+---
+
+Set default textSans line height to 'regular'
```
sndrs, line 9: It would be really useful to outline what consumers will need to do to accommodate this

packages/@guardian/source-react-components/src/checkbox/styles.ts
```diff
@@ -84,7 +84,7 @@ export const checkbox = (
 				color: ${checkbox.textIndeterminate};
 				content: '-';
 				position: absolute;
-				top: -10px;
+				top: -${space[2]}px;
```
sndrs, line 87: 👍

.changeset/cool-deers-chew.md
joecowton1: The intention is that consumers shouldn't need to make any changes, but we could say that they should expect any textSans copy to now have a tighter line height?

.changeset/cool-deers-chew.md
```diff
@@ -0,0 +1,9 @@
+---
+'@guardian/eslint-plugin-source-foundations': major
+'@guardian/eslint-plugin-source-react-components': major
+'@guardian/source-foundations': major
+'@guardian/source-react-components': major
+'@guardian/source-react-components-development-kitchen': major
+---
+
+Set default textSans line height to 'regular', there should be no change to Source components but any consumers using textSans should expect a default line height of 1.35 rather than 1.5
```
sndrs, line 9: Great!

sndrs (approved):
Great stuff

mxdvl:
I’m afraid this makes our text inaccessible: [Success Criterion 1.4.12: Text Spacing](https://www.w3.org/WAI/WCAG22/Understanding/text-spacing.html)

joecowton1:
Thanks @mxdvl  that's very useful to know! @akemitakagi is going to consult with DAC.

In the meantime we'll roll back these changes.

PR #1538, opened 2022-09-13
Revert "Release package updates"

joecowton1:
Reverts guardian/source#1536


sndrs:
these versions are unpublished from NPM, so we'll need to manually increment the next major bumps for these packages in the release PR (i guess by a patch version?)

PR #1547, opened 2022-09-15
Set button loading spinner width to 24px

ob6160:

It was noted that the loading spinner on our buttons was too big and a width of `24px` was requested. That falls outside of the icon size scale used for the `LoadingSpinner` component, so this PR sets the width directly as a unique case for the spinner when used on the button.

### Before

Screenshot 2022-09-16 at 16 59 17

### After

Screenshot 2022-09-16 at 16 58 34


ob6160:
[Loading button: The loading animated icon is too big.](https://trello.com/c/wHjJ6eka/1080-loading-button-the-loading-animated-icon-is-too-big)

.changeset/swift-flies-trade.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-react-components': minor
```
sndrs, line 2: should this be a patch - a minor should offer new (non-breaking) functionality but does this do that?

```suggestion
'@guardian/source-react-components': patch
```

tjsilver (commented):
How does this work with smaller button sizes (is it even meant to?)?

ob6160:
[Loading button: The loading animated icon is too big.](https://trello.com/c/wHjJ6eka/1080-loading-button-the-loading-animated-icon-is-too-big)

ob6160:
> > How does this work with smaller button sizes (is it even meant to?)?
> 
> Great point — hm, I was going off of our Chromatic diff which doesn't appear to include an example of a smaller button + spinner combination. I'll take a look at how this looks and get back to this thread!

@tjsilver I've confirmed that this is a bug with the existing implementation and raised a new ticket to track it. I'm planning to raise a separate PR to resolve this after talking to @akemitakagi ; as it's a problem with the existing implementation and the scope of this one is to change the default spinner size. Hope that's okay!

tjsilver (approved):
💯

PR #1551, opened 2022-09-21
Apply visually hidden css rules recommended by the DAC 

ob6160:

Modernise our visually hidden rules so that they follow DAC recommendations.

What is the purpose of the class?

> Smushes, clips, de-flows and hides content so that it is only perceptible to assistive technology like a screen reader [^1]

An annotated version of the CSS rules, explaining why they are there (thanks @SiAdcock for providing some of the context around these!):

[^1]: https://medium.com/@jessebeach/beware-smushed-off-screen-accessible-text-5952a4c2cbfe#:~:text=Basically%2C%20the%20technique%20smushes%2C%20clips,technology%20like%20a%20screen%20reader.

```css
	position: absolute !important;
	overflow: hidden !important; /* gets rid of horizontal scrollbar that appears in some circumstances */
	white-space: nowrap !important; /* The white-space property forces the content to render on one line. */
	width: 1px !important;  /* ensures content is announced by VoiceOver. */
	height: 1px !important; /* ensures content is announced by VoiceOver. */
	margin: -1px !important; /* hide or clip content that does not fit into a 1-pixel visible area. */
	padding: 0 !important; /* hide or clip content that does not fit into a 1-pixel visible area. */
	border: 0 !important;
	clip: rect(1px, 1px, 1px, 1px) !important; /* clip removes any visible trace of the element */
	-webkit-clip-path: inset(50%) !important; /* clip removes any visible trace of the element */
	clip-path: inset(50%) !important; /* clip removes any visible trace of the element */
```

### DIFF
--- a/.changeset/tidy-suits-pay.md
+++ b/.changeset/tidy-suits-pay.md
@@ -1,5 +1,5 @@
 ---
-'@guardian/source-foundations': minor
+'@guardian/source-foundations': major
 ---
 
 - change the visually hidden CSS to conform to the DAC standard. **Note:** this may cause changes to your layout, please ensure that it is not affected after updating to this version
--- a/.changeset/tidy-suits-pay.md
+++ b/.changeset/tidy-suits-pay.md
@@ -1,0 +1,5 @@
+---
+'@guardian/source-foundations': minor
+---
+
+- change the visually hidden CSS to conform to the DAC standard. **Note:** this may cause changes to your layout, please ensure that it is not affected after updating to this version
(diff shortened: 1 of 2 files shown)


joecowton1 (dismissed):
Nice work.

ob6160:
[Modernize the visuallyHidden CSS in Source Foundations](https://trello.com/c/j9gtnG5F/781-modernize-the-visuallyhidden-css-in-source-foundations)

.changeset/tidy-suits-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-foundations': minor
```
sndrs, line 2: if we think this could break layouts, it should be a major (see https://github.com/guardian/source/blob/main/docs/contributing/overview.stories.mdx#versioning)

```suggestion
'@guardian/source-foundations': major
```

sndrs (changes requested):

sndrs (approved):
greta!

PR #1552, opened 2022-09-21
Remove two deprecated colours from our palette

ob6160:

The colours `opinion.300` and `culture.350` are duplicates of `opinion.400` and `culture.400` and have been deprecated for a while.

This pull request removes both values and changes one instance where we were using `opinion.300` in the dev kitchen.


ob6160:
We're not ready to remove these yet

PR #1554, opened 2022-09-22
Fix invalid CSS; add assertions that CSS is valid

ob6160:

In some cases, Emotion CSS syntax has started to slip into `source-foundations`.

This means that the latest version of foundations is breaking the build in projects like dotcom rendering [(AMP specifically)](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/amp/server/document.tsx#L98) when upgrading the version where the CSS blocks that we export are being rendered directly to the page.

This change fixes some invalid CSS comments that use `//` vs `/* .. */` in our reset which could affect the cascade and adds unit tests which make an assertion using the [`lightningcss`](https://github.com/parcel-bundler/lightningcss/) parser so we have confidence that we are shipping valid CSS.

We also clarify the use of the parent selector in `resets.input` and in the Typography module — by making two exports for each font type. The difference is explained below:

- `headline.medium()` — This will now return a `SerializedStyles` object, because it returns a parent selector. It is a breaking change to the API because this used to return an invalid CSS string.
- `headlineString.medium()` — This will return a CSS string as before, but without including the parent selector.

These assertions are implemented as a custom Jest matcher called `toBeValidCSS`, which accepts an option for when the CSS to check is a fragment and not surrounded by a selector.

I was wary of adding a new dependency here, but decided to include this in the pull request proposal because it seems very easy to let invalid CSS syntax slip into what is (I think at least?) a package that should be framework agnostic.

Thanks @SiAdcock & @joecowton1  for a great discussion around this!


packages/@guardian/source-foundations/src/utils/resets.test.ts
```diff
@@ -0,0 +1,56 @@
...
+					${resets.fieldset}
+				}`,
+				'utf8',
+			),
+			filename: '',
+		}),
+	).not.toThrowError();
+});
+
+it('should provide SCSS for resets.input', () => {
```
ob6160, line 24: we expect this to error, because for this case we export SCSS — but fixing this will be more involved and arguably beyond the scope of this work

ob6160:
[Fix invalid CSS syntax exported from source-foundations](https://trello.com/c/Zrkv4oox/1091-fix-invalid-css-syntax-exported-from-source-foundations)

packages/@guardian/source-foundations/src/utils/resets.test.ts
ob6160: I added a comment above that css block where it’s defined to make this intention clear

packages/@guardian/source-foundations/src/accessibility/accessibility.test.ts
```diff
@@ -0,0 +1,45 @@
...
+		transform({
+			code: Buffer.from(
+				`* {
+					${focusHalo}
+				}`,
+				'utf8',
+			),
+			filename: '',
+		}),
+	).not.toThrowError();
```
mxdvl, line 16: I would much rather this read like: `expect(isValidCSS(focusHalo)).toBe(true)`.

If you want to be really fancy, you could write [a custom matcher](https://jestjs.io/docs/expect#custom-matchers-api) and have `expect(focusHalo).toBeValidCSS()`.

packages/@guardian/source-foundations/src/utils/resets.ts
```diff
@@ -17,6 +17,7 @@ const fieldset = `
 `;
 
 // remove styling of invalid input elements that gets applied in Firefox
+// this uses the SCSS parent selector and in the future we should look to make this just CSS
```
mxdvl, line 20: If you move to JSDoc here, this will make it clearer to consumers that this is Sass and not CSS!

mxdvl (approved):
This is a great catch. I have an issue with the fact that this relies on a specific library’s behaviour when parsing CSS.

We want to avoid early abstraction, but as we are already using the `transform` four times already, I would strongly suggest creating a helper method like `isValidCSS` that is tested with valid and invalid inputs. While I am quite familiar with `lightningcss` having used if very recently, I think that the indirection of creating a buffer, specifying a filename and expecting errors to be thrown makes this code harder to maintain.

packages/@guardian/source-foundations/src/accessibility/accessibility.test.ts
ob6160: completely agree — a lot clearer and concise

.vscode/extensions.json
```diff
@@ -5,6 +5,7 @@
 		"styled-components.vscode-styled-components",
 		"stylelint.vscode-stylelint",
 		"stkb.rewrap",
-		"naumovs.color-highlight"
+		"naumovs.color-highlight",
+		"tobermory.es6-string-html"
```
sndrs, line 9: why is this added?

packages/@guardian/source-foundations/src/accessibility/accessibility.test.ts
```diff
@@ -0,0 +1,45 @@
+import { transform } from 'lightningcss';
+import { focusHalo, focusHaloSpaced } from './focus-halo';
+import { visuallyHidden } from './visually-hidden';
+
+it('should provide valid CSS for focusHalo', () => {
```
sndrs, line 5: a bit nit picky, but how do you feel about something like this? (it took me a sec to understand the intention of the test)
```suggestion
test('focusHalo is valid CSS', () => {
```

packages/@guardian/source-foundations/src/utils/resets.ts
sndrs: if input is not valid CSS, we should probably rename it to make that clear.

given we export `resets.resetCSS`, maybe something like `resets.inputSCSS`? i don't think it would be clear (or expected) that we'd export scss from this package (it's news to me!)

packages/@guardian/source-foundations/src/utils/resets.test.ts
sndrs: is there a similar package we can use to verify that it's valid scss?

sndrs: given that scss is a superset of css, being invalid css is not proof that we have valid scss :) it could be invalid css _and_ scss

.vscode/extensions.json
ob6160: this (and the associated `/* css */` comments add syntax highlighting to our css template strings — I need to update the PR description as I just added this after a conversation with @mxdvl

packages/@guardian/source-foundations/src/utils/resets.ts
ob6160: agreed 👍

packages/@guardian/source-foundations/src/accessibility/accessibility.test.ts
ob6160: agreed 👍

.vscode/extensions.json
mxdvl: Yes, I think this should live in a separate PR!

packages/@guardian/source-foundations/src/accessibility/focus-halo.ts
```diff
@@ -5,14 +5,14 @@ import { palette } from '../colour/palette';
  *
  * CSS rules that provide a clear visual focus state to support keyboard navigation.
  */
-export const focusHalo = `
+export const focusHalo = /* css */ `
```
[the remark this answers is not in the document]
ob6160: see: https://github.com/guardian/source/pull/1554#discussion_r977813699

packages/@guardian/source-foundations/src/utils/resets.ts
```diff
@@ -28,16 +29,16 @@ const input = `
...
 	*, *:before, *:after {
         box-sizing: border-box;
     }
     html {
         -moz-osx-font-smoothing: grayscale;
         -webkit-font-smoothing: antialiased;
-        // always show the vertical scroll bar to stop the page
-        // jumping about when navigating between pages where 
-        // one has content shorter than the viewport
+        /* always show the vertical scroll bar to stop the page
```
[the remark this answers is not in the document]
ob6160: it's such a sneaky one that slipped through the net!

there's another one in typography.ts which is coming as a followup :)

packages/@guardian/source-foundations/src/utils/resets.test.ts
ob6160: hahaha, yes good point. 
Although useful for cases like this.. My immediate thought is that by adding `xyzSCSS` are we implicitly saying that SCSS is okay to go ahead and use in `source-foundations`?
If this is something we don't want, we might want to signal that quite strongly

packages/@guardian/source-foundations/src/utils/resets.ts
mxdvl: Could we make this valid CSS by simply doing:

```css
input:invalid {
  box-shadow: none;
}
```

.vscode/extensions.json
ob6160: 👍  will add as a followup

ob6160: have now removed this from the PR

packages/@guardian/source-foundations/src/utils/jest-matchers/toBeValidCSS.ts
```diff
@@ -0,0 +1,53 @@
...
+			return {
+				pass: true,
+				message: () => '',
+			};
+		} catch (e) {
+			return {
+				pass: false,
+				message: () => (e as Warning).message,
+			};
+		}
```
mxdvl, line 51: You can probably check that `e instanceof Warning` and otherwise throw that again.

```ts
		catch (e: unknown) {
			if(!(e instanceof Warning) throw e;
			return {
				pass: false,
				message: () => (e as Warning).message,
			};
		}
```

packages/@guardian/source-foundations/src/accessibility/focus-halo.ts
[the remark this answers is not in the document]
ob6160: removed from PR to be proposed in a followup

packages/@guardian/source-foundations/src/utils/resets.ts
ob6160: Screenshot 2022-09-22 at 17 04 20

it's individually applied to the `TextArea` (and `TextInput`) in Source.. not sure about if it's used by any of our consumers..

I suppose that would be fine to move into our main reset as long as we're sure we'd like this behaviour everywhere?

mxdvl: If it is only used twice… is is useful to have this as an abstraction? Should we simply inline it in both of these styles?

packages/@guardian/source-foundations/jest.config.js
```diff
@@ -2,4 +2,5 @@
 module.exports = {
 	preset: 'ts-jest',
 	testEnvironment: 'node',
+	setupFilesAfterEnv: ['./src/utils/jest-matchers/index.ts'],
```
sndrs, line 5: could we move this to `packages/@guardian/source-foundations/lib`? the other things that live in `utils` are actually code that ships, whereas this is code that analyses that code.

appreciate the naming of these folders is not super clear, but we do have a pattern of using `libs` (e.g. at the root) for this type of thing, so it would be consistent, if not clear 🙃 

we should probably make this stuff a bit clearer subsequently

packages/@guardian/source-foundations/src/utils/resets.test.ts
```diff
@@ -0,0 +1,23 @@
+import { resets } from './resets';
+
+test('resets.defaults should be valid CSS', () => {
+	expect(resets.defaults).toBeValidCSS();
+});
```
sndrs, line 5: these read so clearly!

sndrs: ❤️

packages/@guardian/source-foundations/src/utils/resets.ts
sndrs: @SiAdcock do you know why this is not in the global reset already? is there a good reason for it?

packages/@guardian/source-foundations/src/utils/resets.test.ts
sndrs: i guess the alternative is to mask that something _is_ scss, and therefore may break if you use it as css, which seems like a bad situation to me!

i'm not sure we _should_ be providing scss, but given that currently we _do_, making sure it's only used in the correct places has value.

we could mark it as `@deprecated` if you like, so at least we can convey that the use it not encouraged

packages/@guardian/source-foundations/src/accessibility/accessibility.test.ts
sndrs: llovely thanks

packages/@guardian/source-foundations/src/utils/resets.test.ts
sndrs: we have a pattern for this in `react-components` https://github.com/guardian/source/blob/main/packages/%40guardian/source-react-components/src/deprecated-exports.ts

packages/@guardian/source-foundations/src/utils/jest-matchers/toBeValidCSS.ts
ob6160: couldn't check to see if it was a `Warning` because it's only supplied as an interface which we can't do an `instanceof` check against.

Instead I've put a check in for the `message` property and throw if that is not present

packages/@guardian/source-foundations/src/utils/resets.test.ts
ob6160: I've added a `deprecated` rule above this

packages/@guardian/source-foundations/jest.config.js
ob6160: moved!

.changeset/popular-adults-behave.md
```diff
@@ -0,0 +1,6 @@
+---
+'@guardian/source-foundations': patch
+---
+
+- fix invalid CSS syntax in the exported reset.
```
sndrs, line 5:
```suggestion
- remove invalid syntax from the css reset
```

.changeset/popular-adults-behave.md
```diff
@@ -0,0 +1,6 @@
...
+- add assertions and new unit tests for areas where we export CSS and want to ensure that SCSS does not slip into the code.
```
sndrs, line 6: as this is an internal implementation detail, i'd be inclined to remove this. knowing that tests have been added doesn't affect how i would consume the package, so we maybe don't need to add this to the changelog?

```suggestion
- add assertions and new unit tests for areas where we export CSS and want to ensure that SCSS does not slip into the code.
```

packages/@guardian/source-foundations/src/utils/resets.ts
```diff
@@ -16,6 +16,10 @@ const fieldset = `
 	margin: 0;
 `;
 
+/**
+ * @deprecated this uses the SCSS parent selector. In the future we
+ * should look to make this just CSS or move into into our main reset.
+ */
 // remove styling of invalid input elements that gets applied in Firefox
 const input = `
```
sndrs, line 24: are we deprecating the name, the language or the functionality here? 

i.e. if this become pure css, what is deprecated about this?

i wonder if we should rename this to rename this to `inputSCSS` and re-export it as a deprecated `input` with the deprecation message?

```suggestion
/**
 * @deprecated use `resets.inputSCSS`
 */
 const input = inputSCSS;
 
// remove styling of invalid input elements that gets applied in Firefox
const inputSCSS = `
```

that effectively adds a new export (`resets.inputSCSS`) while flagging up the old, ambiguous name as something we want to remove.

that would then decouple the misleading name issue from any future work we might want to do on making the reset available as pure css.

.changeset/popular-adults-behave.md
ob6160: good point 👍

packages/@guardian/source-foundations/src/utils/resets.ts
ob6160: Agreed, that’s a lot clearer - thanks for the suggestion!

packages/@guardian/source-foundations/src/utils/resets.ts
sndrs: let leave this for now, renaming should clear up confusion and we can work on an improvement as a separate change

packages/@guardian/source-foundations/src/utils/resets.ts
ob6160: As discussed yday - I've changed `input` to just have the type: `SerialisedStyles` which should indicate this properly

packages/@guardian/source-foundations/src/typography/font-styles.ts
```diff
@@ -103,37 +106,54 @@ export const fontStyleFunction =
...
+const fontStyleToCssString = (styles: TypographyStyles) => {
+	const { fontFamily, fontSize, fontStyle, fontWeight, lineHeight } = styles;
+
+	return `
+			font-family: ${fontFamily};
+			font-size: ${typeof fontSize === 'number' ? `${fontSize}px` : fontSize};
+			line-height: ${lineHeight};
+			${fontWeight ? `font-weight: ${fontWeight};` : ''}
+			${fontStyle ? `font-style: ${fontStyle};` : ''}
+		`;
```
mxdvl, line 121: This adds tabs to the string and makes it multi-line. Maybe simply an array that’s joined with return carriages?

```ts
return [
	`font-family: ${fontFamily};`,
	`font-size: ${typeof fontSize === 'number' ? `${fontSize}px` : fontSize};`,
	`line-height: ${lineHeight};`,
	fontWeight && `font-weight: ${fontWeight};`,
	fontStyle && `font-style: ${fontStyle};`,
].filter(Boolean).join("\n");
```

ob6160: If we change this I feel like we'd need to change all examples of CSS defined as template strings, as this is a common pattern throughout this package (see: https://github.com/guardian/source/blob/main/packages/%40guardian/source-foundations/src/utils/resets.ts#L64) — my feeling is that by switching to an array in this one case we might be introducing something that could surprise any future developers working on source-foundations.

There might also be a discussion to be had around readability vs the benefits of shipping the CSS without tabs

ob6160:
Closing to reopen as two distinct changes, one for the CSS validity assertions and another for our Emotion exports

ob6160:
Closing to reopen as two distinct changes, one for the CSS validity assertions and another for our Emotion exports

ob6160:
[Fix invalid CSS syntax exported from source-foundations](https://trello.com/c/Zrkv4oox/1091-fix-invalid-css-syntax-exported-from-source-foundations)

PR #1558, opened 2022-09-26
Apply correct loading spinner size on buttons

ob6160:

The loading spinner is currently not displaying at the correct width for button sizes `small`, and `xsmall`. We'd like to correct this so the spinner renders at the correct size for every provided button size.

## What does this change?

This change defines the desired spinner width in pixels as a mapping for each possible button size and applies it when rendering the button.

## Screenshots

**Before**
Screenshot 2022-09-26 at 14 11 40

**After**
Screenshot 2022-09-26 at 14 10 21

### DIFF
--- a/.changeset/smooth-garlics-swim.md
+++ b/.changeset/smooth-garlics-swim.md
@@ -1,0 +1,5 @@
+---
+'@guardian/source-react-components': patch
+---
+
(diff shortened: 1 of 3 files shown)


joecowton1 (approved):
Looks good to me!

PR #1559, opened 2022-09-27
Pass value to textarea component

joecowton1:

@guardian/investigations reported a bug where `value` was not being passed into the `textArea` when the component loaded. 

## What does this change?

- adds `value` field to `textArea` and populates with `value` prop
- adds a change handler to the text area stories


ob6160 (dismissed):
🔧

sndrs:
worth a story?

joecowton1:
> worth a story?

Added, that threw up some odd behaviour so have amended the original PR, see description for an explanation.

ob6160 (approved):
good catch with the `defaultValue` 👍

packages/@guardian/source-react-components/src/text-area/TextArea.tsx
```diff
@@ -131,6 +131,7 @@ export const TextArea = ({
 				required={!optional}
 				rows={rows}
 				className={getClassName()}
+				defaultValue={value}
```
ob6160, line 134: Should the name of the prop passed in also be `defaultValue`? It confused me when I read the story as I expected the names to match up as the behaviour is different

joecowton1: I think maybe we could add another prop called `defaultValue`, but if we were to change `value` to `defaultValue` we'd be at cross purposes with the way this component is [generally used](https://github.com/guardian/dotcom-rendering/blob/3ac24426c459fb8456c81b5148da6c883872189e/dotcom-rendering/src/web/components/Callout/TextArea.tsx#L23).

That might make sense though?

joecowton1: I've realised the issue was actually just with the way the story was composed - if we have a controlled component we need to pass a change handler in, this was missing in the story.

Now its added in we can use the value field `value={value}` and everything is nice.

ob6160: Sounds good 👍

.changeset/purple-countries-think.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-react-components': patch
+---
+
+Add value field to textArea
```
ob6160, line 5: maybe this could be slightly reworded to mention that it passes through the value field as expected so the component works as a controlled component . As the value field was already added on the component level, just not wired up internally to the inner `textarea`

joecowton1: I'd thought the lower case `t` indicated that but happy to change

ob6160 (approved):

PR #1566, opened 2022-10-03
Check that we are exporting valid CSS

ob6160:

In some cases, Emotion style CSS is exported from `source-foundations`. Applications that consume these exported styles can have an expectation that only CSS is provided, but right now we make no guarantee that this is the case.

This breaks the build in projects like dotcom rendering [(AMP specifically)](https://github.com/guardian/dotcom-rendering/blob/main/dotcom-rendering/src/amp/server/document.tsx#L98), because these invalid CSS rules are being rendered directly onto the page, without going through the Emotion stack.

This change fixes some invalid CSS comments that use `//` vs `/* .. */` in our reset which could affect the cascade. We also add unit tests using the [`lightningcss`](https://github.com/parcel-bundler/lightningcss/) project to parse our exported CSS so we have confidence that we are shipping valid code.

## What does this change?

- Adds a custom "Jest matcher" called `.toBeValidCSS()` so we can make assertions in a natural way against the CSS that we export using the `lightningcss` library.
- Adds these assertions to our unit tests for the CSS resets and accessibility classes that we export.

This PR was split out from https://github.com/guardian/source/pull/1554 so we could make it distinct from other changes to export emotion styles from the typography module.


joecowton1 (approved):
Much clearer! Thanks for breaking it out.

ob6160:
[Fix invalid CSS syntax exported from source-foundations](https://trello.com/c/Zrkv4oox/1091-fix-invalid-css-syntax-exported-from-source-foundations)

PR #1570, opened 2022-10-05
Update `ToggleSwitch` component styles

joecowton1:

This PR updates the `ToggleSwitch` component with the latest design changes.

refs #1575

## What does this change?

The new designs introduce a number of changes to the component API:
- `fontSize` is now a prop and limited to `small` and `medium`
- `fontWeight` is now a prop and limited to `regular` and `bold`
- `labelBorder` is now a prop (boolean)
- `format` is now a prop, providing this in the consumer will render an alternative colour toggle suitable for a coloured background.
- ~`noJs` is now a prop, this needs to be provided by the consumer. This replaces the previous internal check.~

Each of these changes is demonstrated in the corresponding stories.

There are also updates to the component layout and a fix for the tick loading to soon when the toggle is checked.

## Screenshots

**Label border**
Screenshot 2022-10-05 at 16 49 07

**Label border medium**
Screenshot 2022-10-05 at 16 54 58

**Label border bold**
Screenshot 2022-10-05 at 16 49 35

**Label border bold medium**
Screenshot 2022-10-05 at 16 49 40

**With format prop (news)**
Screenshot 2022-10-05 at 16 50 32

**With format prop (news) (selected)**
Screenshot 2022-10-05 at 16 58 29

### Accessibility

- [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
- [x] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
- [x] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
- [x] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

- [x] Tested with touch screen device

### Responsiveness

- [x] Tested at all breakpoints
- [x] Tested with with long text
- [x] Stretched to fill a wide container

### Documentation

- [x] Full API surface area is documented in the README
- [x] Examples in Storybook

### DIFF
--- a/packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/ToggleSwitch.tsx
+++ b/packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/ToggleSwitch.tsx
@@ -1,6 +1,7 @@
 import type { EmotionJSX } from '@emotion/react/types/jsx-namespace';
 import type { ArticleFormat } from '@guardian/libs';
 import { descriptionId, generateSourceId } from '@guardian/source-foundations';
+import { useEffect, useState } from 'react';
 import type { Props } from '../../../source-react-components/src/@types/Props';
 import {
 	buttonStyles,
@@ -50,11 +51,6 @@ export interface ToggleSwitchProps extends Props {
 	 */
 	labelBorder?: boolean;
 	/**
-	 * If JS is not available the toggle is not clickable and shows a tooltip.
-	 * The default is false.
-	 */
-	noJs?: boolean;
-	/**
 	 * Sets the font weight to either 'regular' or 'bold'.
 	 * Note this does not take all font weights, only 'regular' and 'bold'.
 	 */
@@ -91,7 +87,6 @@ export const ToggleSwitch = ({
 	label,
 	labelBorder = false,
 	labelPosition = 'right',
-	noJs = false,
 	defaultChecked,
 	cssOverrides,
 	onClick,
@@ -99,16 +94,24 @@ export const ToggleSwitch = ({
 }: ToggleSwitchProps): EmotionJSX.Element => {
 	const buttonId = id ?? generateSourceId();
 	const labelId = descriptionId(buttonId);
+	const [isBrowser, setIsBrowser] = useState(false);
+	let tooltiptext = '';
 
 	const isChecked = (): boolean => {
-		if (noJs) return false;
-
 		if (checked != undefined) {
 			return checked;
 		}
 
 		return !!defaultChecked;
 	};
+
+	useEffect(() => {
+		setIsBrowser(true);
+	});
+
+	if (!isBrowser) {
+		tooltiptext = 'tooltiptext';
+	}
 
 	return (
 		<label
@@ -130,7 +133,7 @@ export const ToggleSwitch = ({
 				onClick={onClick}
 			></button>
 			{labelPosition === 'right' && label}
-			<div className={noJs ? 'show-tooltip' : ''} css={tooltipStyles}>
+			<div className={tooltiptext} css={tooltipStyles}>
 				<span>Please turn on JavaScript to use this feature</span>
 			</div>
 		</label>
--- a/packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/ToggleSwitch.tsx
+++ b/packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/ToggleSwitch.tsx
@@ -61,6 +61,10 @@ export interface ToggleSwitchProps extends Props {
 	 */
 	fontSize?: ToggleSwitchFontSize;
 	/**
+	 * Whether the toggle has a tooltip.
+	 * The default is false. */
+	tooltip?: boolean;
+	/**
 	 * A callback function called when the component is checked or unchecked.
 	 * Receives the click event as an argument.
 	 */
--- a/packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/ToggleSwitch.tsx
+++ b/packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/ToggleSwitch.tsx
@@ -1,15 +1,18 @@
 import type { EmotionJSX } from '@emotion/react/types/jsx-namespace';
+import type { ArticleFormat } from '@guardian/libs';
 import { descriptionId, generateSourceId } from '@guardian/source-foundations';
-import type { Props } from '@guardian/source-react-components';
-import { useEffect, useState } from 'react';
+import type { Props } from '../../../source-react-components/src/@types/Props';
 import {
 	buttonStyles,
+	labelBorderStyles,
 	labelStyles,
 	toggleStyles,
 	tooltipStyles,
 } from './styles';
 
 export type LabelPosition = 'left' | 'right';
+export type ToggleSwitchFontWeight = 'regular' | 'bold';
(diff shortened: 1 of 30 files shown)


packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/ToggleSwitch.tsx
```diff
@@ -65,53 +84,55 @@ export interface ToggleSwitchProps extends Props {
...
+	onClick,
 	...props
 }: ToggleSwitchProps): EmotionJSX.Element => {
 	const buttonId = id ?? generateSourceId();
 	const labelId = descriptionId(buttonId);
-	const [isBrowser, setIsBrowser] = useState(false);
-	let tooltiptext = '';
 
 	const isChecked = (): boolean => {
+		if (noJs) return false;
```
joecowton1, line 104: I'm not sure if it's right to leave this in? It allows us to test but seems slightly odd.

packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/ToggleSwitch.tsx
```diff
@@ -39,12 +46,24 @@ export interface ToggleSwitchProps extends Props {
...
 	/**
-	 * Whether the toggle has a tooltip.
+	 * Displays a border above the toggle switch label and toggle.
+	 */
+	labelBorder?: boolean;
+	/**
+	 * If JS is not available the toggle is not clickable and shows a tooltip.
 	 * The default is false.
 	 */
-	tooltip?: boolean;
```
ob6160, line 45: how come this is being removed? couldn't find a mention of it in the pr description

packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/ToggleSwitch.tsx
```diff
@@ -39,12 +46,24 @@ export interface ToggleSwitchProps extends Props {
...
+	noJs?: boolean;
```
ob6160, line 56: could this name be more descriptive perhaps? maybe something like `showNoJavascriptTooltip`?

packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/styles.ts
```diff
@@ -1,8 +1,25 @@
...
+import type {
+	LabelPosition,
+	ToggleSwitchFontSize,
+	ToggleSwitchFontWeight,
+} from './ToggleSwitch';
 
-export const buttonStyles = (labelPosition: LabelPosition) => css`
+const toggleBackground = 'rgba(255, 255, 255, 0.4)';
+const toggleBorder = 'rgba(255, 255, 255, 0.6)';
+const toggleBorderGreen = '#A7CFB8';
```
ob6160, line 18: is this something that our palette doesn't provide?

packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/styles.ts
```diff
@@ -17,95 +34,112 @@ export const buttonStyles = (labelPosition: LabelPosition) => css`
...
+	border-top: 1px solid ${format ? neutral[100] : neutral[46]};
+	padding-top: ${space[1]}px;
+	width: 100%;
+`;
+
 export const tooltipStyles = css`
 	position: absolute;
-	visibility: hidden;
-	/* display: none; */
+	display: none;
```
ob6160, line 135: I think `display: none;` will remove the tooltip from the accessibility tree, so a screen reader couldn't access it. Do we want them to be read out?

ob6160 (commented):
nice work! a few comments that came up, but nothing major

packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/styles.ts
joecowton1: yes.

packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/styles.ts
joecowton1: Good point - this will be superseded in the next changes to add a fallback so not sure how important it is but shouldn't be introducing accessibility regressions, I'll have a look at it.

packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/ToggleSwitch.tsx
joecowton1: The tooltip is being binned imminently in favour of a fallback so really all this prop is for is for the platform to specify whether or not JS is available, it might be that down the line this changes but i feel like for now it's suitably generic :)

packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/styles.ts
joecowton1: Put it back to using `visibility`

packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/ToggleSwitch.tsx
joecowton1: The `tooltip` prop was also not used anywhere previously, its a bizarre bit of implementation.

joecowton1: reinstated this prop so no longer relevant

packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/styles.ts
ob6160: nice one — I wonder if `visuallyHidden` might be useful here, or if it doesn't work for this use-case 🤔

joecowton1: Sorry maybe wasn't clear enough - this will be replaced in the coming weeks so probably doesn't need changing at this stage.

ob6160: ah, sounds good!  nice one :)

packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/styles.ts
ob6160: okay, maybe something to double check with Akemi whether it belongs in palette? (sorry if you already have)

joecowton1: (its the colour we would get by layering 40% opacity white border over success[400], but because borders don't sit over the thing they surround, we have to pre-calculate that colour and apply as a standard border)

ob6160: thanks for the explanation, that makes sense to me :) 

maybe a small comment above it with that layering method would be helpful, so we remember why it's not coming from the palette

ob6160 (approved):
⭐️

ob6160 (approved):

PR #1572, opened 2022-10-05
Make textSans default line-height "regular"

joecowton1:

We'd like to set a default line height for each main font that we provide in Source.

Text Sans has had a line height of 150 as a default for many years. In design, we mostly use Text Sans with a line height of 135.

This work is centred around changing that line height and managing the knock-on effects of doing so. We'd like to have as little impact on the external teams as possible.

Recreation of: https://github.com/guardian/source/pull/1531

Issues around accessibility have been resolved by @akemitakagi and DAC:

Success Criterion 1.4.12 Text Spacing (Level AA) doesn’t mean all of the line height needs to be at least 1.5 times the font size.

Its author responsibility mentions
‘This SC (Success Criterion) does not dictate that authors must set all their content to the specified metrics. Rather, it specifies that an author's content has the ability to be set to those metrics without loss of content or functionality.’

We can set any line height we want as long as the user can override the styles.
The Guardian supports custom css to override our style.
Therefore we cleared this Success Criterion.

## What does this change?

- Updates textSans line-height default to `regular` 
- Removes overrides which were setting line-height to `regular` on a per-component basis
- Removes additional padding in `Button` component which was added to correct layout with incorrect line-height
- Removes additional margin which was used to correct alignment in `Summary` component

## UI Changes

The intention behind this PR is that we should update the line height but not cause any UI changes. See Chromatic for a full list of components effected.

### DIFF
--- a/.changeset/thin-eyes-wonder.md
+++ b/.changeset/thin-eyes-wonder.md
@@ -1,0 +1,7 @@
+---
+'@guardian/source-foundations': major
+'@guardian/source-react-components': major
+'@guardian/source-react-components-development-kitchen': major
+---
+
+Set default textSans line height to regular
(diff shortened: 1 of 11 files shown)


ob6160 (approved):
✏️ ⭐

PR #1588, opened 2022-10-10
Add e2e tests for eslint plugins

tjsilver:

This PR adds e2e tests to the eslint plugins, so that when they are migrated over to the csnx monorepo, the project e2e script will have something to run. This is linked to the PR to [create skeletons in csnx](https://github.com/guardian/csnx/pull/120).

## What does this change?

- Adds e2e tests for each plugin.
- Upgrades `@guardian/libs` dependency to latest - this resolved a peer dependency error on `web-vitals` that occurred when running the new e2e tests.

### DIFF
--- a/packages/@guardian/eslint-plugin-source-foundations/jest.config.js
+++ b/packages/@guardian/eslint-plugin-source-foundations/jest.config.js
@@ -1,6 +1,4 @@
-/** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
 module.exports = {
 	preset: 'ts-jest',
 	testEnvironment: 'node',
-	setupFilesAfterEnv: ['./lib/jest-matchers/index.ts'],
 };
--- a/packages/@guardian/eslint-plugin-source-foundations/jest.config.js
+++ b/packages/@guardian/eslint-plugin-source-foundations/jest.config.js
@@ -1,0 +1,6 @@
+/** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
+module.exports = {
+	preset: 'ts-jest',
+	testEnvironment: 'node',
+	setupFilesAfterEnv: ['./lib/jest-matchers/index.ts'],
+};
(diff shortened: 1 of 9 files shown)


packages/@guardian/eslint-plugin-source-foundations/jest.config.js
```diff
@@ -0,0 +1,6 @@
+/** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
+module.exports = {
+	preset: 'ts-jest',
+	testEnvironment: 'node',
+	setupFilesAfterEnv: ['./lib/jest-matchers/index.ts'],
```
sndrs, line 5: is this matcher used in these tests?

packages/@guardian/eslint-plugin-source-foundations/package.json
```diff
@@ -28,7 +29,7 @@
 	"devDependencies": {
 		"@guardian/eslint-config": "^1.0.0",
 		"@guardian/eslint-config-typescript": "^1.0.0",
-		"@guardian/libs": "^3.0.0",
+		"@guardian/libs": "^8.0.5",
```
sndrs, line 32: why this bump?

tjsilver: Because otherwise the e2e test throws an error about a missing peer dep of `web-vitals` (see description). Also thought it would make it easier to use the `*workspace` option once the package is migrated as the latest is used in csnx. The other option is to add `web-vitals` as a dependency and then revert the version back to v3. What do you think @sndrs ?

sndrs: i'm not quite sure why this happens? for one thing, v8 of libs still has a peerdep on `web-vitals` https://github.com/guardian/csnx/blob/main/libs/%40guardian/libs/package.json#L17.

how does the missing peerdep affect the test (given it doens't consume the peerdep)?

tjsilver: We chatted about this and established that the latest version of libs imports web vitals dynamically, which is why this error is resolved when we upgrade. We are going to bump libs in all Source packages in a separate PR.

packages/@guardian/eslint-plugin-source-foundations/jest.config.js
tjsilver: No, have removed it!

sndrs (dismissed):

PR #1584, opened 2022-10-08
Apply correct state transitions for form element borders

ob6160:
## What is the purpose of this change?

We'd like the form element hover, focus and active state behaviour to match the rules set out in Figma.

closes #1581
closes #1580 

## What does this change?

- Text area, Text input, and Select active border colour persists when focused.
- Radio and Checkbox error state border behaviour follows the changes set out in our Figma state change diagram

## Demonstration of behaviour

Screen recordings below demonstrate the active, focus and hover border states for each of the modified form elements.

### Text input

**These videos demonstrate how the focus border colour persists while the user is typing in the text input box for default, error, and success.**

<details>
<summary>Text input behaviour changes</summary>

#### Default state

https://user-images.githubusercontent.com/1771189/194708859-6b6e0b42-c77c-4e40-9a78-a82e4c7fa508.mov

#### Success state

https://user-images.githubusercontent.com/1771189/194708877-41f5a877-b7ef-43a2-800c-bdd74fd2c404.mov

#### Error state

https://user-images.githubusercontent.com/1771189/194708888-01e668dc-4aa5-433c-831e-1cfede7e781b.mov
</details>

### Text area

**These videos demonstrate how the focus border colour persists while the user is typing in the text area box for default, error, and success.**

<details>
<summary>Text area behaviour changes</summary>

#### Default state

https://user-images.githubusercontent.com/1771189/194709082-3ca297c4-d138-4541-9960-3336e02e1b74.mov

#### Success state

https://user-images.githubusercontent.com/1771189/194709101-e4d9d2a6-8c1d-4076-a740-12642f841323.mov

#### Error state

https://user-images.githubusercontent.com/1771189/194709114-e2bcf943-0dfe-4a72-91c1-0fe3c66917b9.mov

</details>

### Select

**This video demonstrates how the focus border colour persists while the select menu is open.**

<details>
<summary>Select behaviour changes</summary>

https://user-images.githubusercontent.com/1771189/194709169-18f2b03d-cf73-437c-8066-5159d6de911d.mov
</details>

### Radio button

**This video demonstrates how the radio button border changes on hover when unchecked and remains in the error state on hover when it is checked.**

<details>
<summary>Radio button behaviour changes</summary>

https://user-images.githubusercontent.com/1771189/194709481-2e613251-8ab5-4bba-81fb-636f94682cdb.mov
</details>

### Checkbox

**This video demonstrates how the checkbox border changes on hover when unchecked and remains in the error state on hover when it is checked.**

<details>
[...]

PR #1589, opened 2022-10-10
Fix focus-visible on `ToggleSwitch`

joecowton1:

Inspired by @ob6160's [recent fixes](https://github.com/guardian/source/pull/1584) this PR updates the `focus-visible` behaviour of the `ToggleSwitch` component.

If the `format` prop is provided to the component the focus ring is `neutral[100]` if not, it is `brand[500]`

## Screenshots

**Before**

Screenshot 2022-10-11 at 12 24 01

**After**

Screenshot 2022-10-11 at 12 20 26

**Before**

Screenshot 2022-10-11 at 12 26 40

**After**

Screenshot 2022-10-11 at 12 26 27

### DIFF
--- a/packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/styles.ts
+++ b/packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/styles.ts
@@ -1,8 +1,31 @@
 import { css } from '@emotion/react';
-import { neutral, success, textSans } from '@guardian/source-foundations';
-import type { LabelPosition } from './ToggleSwitch';
+import type { SerializedStyles } from '@emotion/react';
+import type { ArticleFormat } from '@guardian/libs';
+import {
+	neutral,
+	space,
+	success,
+	textSans,
+} from '@guardian/source-foundations';
+import type {
+	LabelPosition,
+	ToggleSwitchFontSize,
+	ToggleSwitchFontWeight,
+} from './ToggleSwitch';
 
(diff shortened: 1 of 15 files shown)


ob6160 (approved):
😶‍🌫️

packages/@guardian/source-react-components-development-kitchen/src/toggle-switch/styles.ts
```diff
@@ -110,6 +111,11 @@ export const toggleStyles = (format?: ArticleFormat): SerializedStyles => {
 			left: 22px;
 			background: ${neutral[100]};
 		}
+
+		&:focus-visible {
+			outline: 5px solid ${format ? neutral[100] : brand[500]};
+			outline-offset: 3px;
```
sndrs, line 117: is there any chance this could overlap existing elements? if so, this might be considered a breaking change (consumers could need to account for some extra border)

or is the space it consumes already accounted for elsewhere?

joecowton1: Screenshot 2022-10-11 at 18 03 29

joecowton1: The ToggleSwitch component itself doesn't contain any padding but where it's consumed here in DCR liveblogs it fits ok within the space. I'm not sure what that means in terms of whether its a breaking change though?

PR #1591, opened 2022-10-11
Bump @guardian/libs to latest major version

tjsilver:

As part of our work to prepare Source packages for migration, we identified that `@guardian/libs` should be pinned in dev dependencies to the lowest latest major version and set in the peer dependencies to the latest version.

## What does this change?

- bumps version in eslint plugins
- bumps version in dev kitchen

### DIFF
--- a/.changeset/gold-keys-swim.md
+++ b/.changeset/gold-keys-swim.md
@@ -1,0 +1,7 @@
+---
+'@guardian/eslint-plugin-source-foundations': major
+'@guardian/eslint-plugin-source-react-components': major
+'@guardian/source-react-components-development-kitchen': major
+---
+
+Bump @guardian/libs to latest major version
--- a/.changeset/gold-keys-swim.md
+++ b/.changeset/gold-keys-swim.md
@@ -4,4 +4,4 @@
 '@guardian/source-react-components-development-kitchen': major
 ---
 
-Bump @guardian/libs to latest major version
+Bump @guardian/libs to `^9.0.0`
--- a/packages/@guardian/eslint-plugin-source-foundations/package.json
+++ b/packages/@guardian/eslint-plugin-source-foundations/package.json
@@ -28,7 +28,7 @@
 	"devDependencies": {
 		"@guardian/eslint-config": "^1.0.0",
 		"@guardian/eslint-config-typescript": "^1.0.0",
-		"@guardian/libs": "8.0.0",
+		"@guardian/libs": "9.0.0",
 		"@types/estree": "^0.0.51",
 		"rollup": "^2.75.4",
 		"rollup-plugin-ts": "^2.0.7",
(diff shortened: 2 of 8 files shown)


packages/@guardian/eslint-plugin-source-foundations/package.json
```diff
@@ -28,14 +28,14 @@
...
-		"@guardian/libs": "^3.0.0",
+		"@guardian/libs": "8.0.0",
 		"@types/estree": "^0.0.51",
 		"rollup": "^2.75.4",
 		"rollup-plugin-ts": "^2.0.7",
 		"typescript": "~4.6.4"
 	},
 	"peerDependencies": {
-		"@guardian/libs": "^3.0.0",
+		"@guardian/libs": "^8.0.5",
```
sndrs, line 38: did you mean to go with `8.0.5` here? 

the version we pin to in dev deps should be bottom of this range, so i think this should either be `"^8.0.5"` or devdeps pinned to `"8.0.0"`

tjsilver: I think I did pin dev deps to "8.0.0", with peer deps being "^8.0.5", but in any case there's a new version of libs now so they're now all set to v9, with dev deps pinned and peer deps on "^". Is that what you'd expect?

sndrs: 👍👍👍

packages/@guardian/source-react-components-development-kitchen/src/editorial-button/styles.ts
```diff
@@ -160,6 +161,10 @@ export const decideFont = (
 					return css`
 						color: ${specialReport[400]};
 					`;
+				case ArticleSpecial.SpecialReportAlt:
+					return css`
+						color: ${palette.specialReportAlt[200]};
```
tjsilver, line 166: @ioannakok @jamie-lynch @HarryFischer Is this what you expected to see? Please also see the quote icon as well

sndrs (dismissed):
great if @ioannakok @jamie-lynch @HarryFischer are happy!

HarryFischer:
Im happy - thanks @tjsilver

.changeset/gold-keys-swim.md
```diff
@@ -0,0 +1,7 @@
+---
+'@guardian/eslint-plugin-source-foundations': major
+'@guardian/eslint-plugin-source-react-components': major
+'@guardian/source-react-components-development-kitchen': major
+---
+
+Bump @guardian/libs to latest major version
```
sndrs, line 7: might be useful to be explicit about the version

```suggestion
Bump @guardian/libs to `^9.0.0`
```

sndrs (dismissed):

PR #1592, opened 2022-10-12
CSS variable 

ob6160:

As part of the ongoing work to remove our dependency on the Emotion CSS-in-JS library in the `source-foundations` package, this change removes the use of the parent selector in a plain CSS string export from the typography module.

## What does this change?

We replace our use of the parent selector for link hover underline emphasis at the `source-foundations` level with a CSS variable called: `--source-hover-text-decoration-thickness`.

This is scoped to the element that the Typography CSS is applied to, meaning that the Emotion CSS hover behaviour can be moved into the `<Link />` component. The primary consumer of this rule can then access the variable to know how large the underline thickness should be.


joecowton1 (approved):
Nicely done!

packages/@guardian/source-foundations/src/typography/font-styles.ts
```diff
@@ -128,12 +133,15 @@ export const fontStyleToStringFunction =
...
-			}
+			/*
+			 * Child elements (e.g. <Link/>) can use this variable
+			 * to set the thickness of their underline on hover.
+			 *
+			 * The thickness for each font type and weight is defined
+			 * in the underlineThickness object.
+			 */
+			--source-hover-text-decoration-thickness: ${
+				textDecorationThickness ? `${textDecorationThickness}px` : 'auto'
```
padolsey, line 144: Will people ever want to set `textDecorationThickness` to zero? And would this come through as a `Number`? If so, best to do `textDecorationThickness == null ? ...` perhaps to avoid falsey zeros.

packages/@guardian/source-foundations/src/typography/font-styles.ts
```diff
@@ -128,12 +133,15 @@ export const fontStyleToStringFunction =
...
-				};
-			}
+			/*
+			 * Child elements (e.g. <Link/>) can use this variable
+			 * to set the thickness of their underline on hover.
+			 *
+			 * The thickness for each font type and weight is defined
+			 * in the underlineThickness object.
+			 */
+			--source-hover-text-decoration-thickness: ${
```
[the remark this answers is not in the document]
ob6160: thinking about it, I think it's for general thickness, just used on hover in this case 👍 will rename

packages/@guardian/source-foundations/src/typography/font-styles.ts
ob6160: good spot — right now the text decoration thickness is pulled from [this mapping](https://github.com/guardian/source/blob/main/packages/%40guardian/source-foundations/src/typography/data.ts#L118) so we can customise it for different font/weight combinations.

that mapping doesn't return 0 right now - but could in the future, so I think specifically checking for `undefined` here makes sense 👍

packages/@guardian/source-react-components/src/link/styles.ts
```diff
@@ -26,6 +26,14 @@ export const link = css`
...
 		${focusHalo};
 	}
+
+	/* If the hover text decoration thickness is not set, we default to the initial value. */
+	&:hover {
+		text-decoration-thickness: var(
+			--source-hover-text-decoration-thickness,
+			auto
+		);
+	}
```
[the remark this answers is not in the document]
ob6160: moved the comment slightly further up (above the rule) because having it inline with the `var()` rule broke syntax highlighting

.changeset/tall-knives-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-react-components': patch
+---
+
+- The `<Link />` component now expects the `--source-text-decoration-thickness` CSS variable to be provided to specify the underline thickness on hover.
```
sndrs, line 5: I think this makes it sound like users of `Link` need to provide this. how about:
```suggestion
- `<Link />` text has variable underline heights, based on the size of the link text
```
The technique we use to accomplish it is only really an implementation detail. Consumer's don't need to know about it.

.changeset/chilly-shrimps-accept.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-foundations': patch
+---
+
+- Introduced a CSS variable: `--source-text-decoration-thickness`, provided by the typography API
```
sndrs, line 5: How do you feel about reordering this sentence?
```suggestion
- Typography functions now include a `--source-text-decoration-thickness` CSS variable
```
This refines the scope as you go, rather than starting with the micro level (the change) and building up to the macro level (the area that includes the change): 

area (Typography functions) > type of change (now include) > the change (a `--source-text-decoration-thickness` CSS variable)

sndrs (dismissed):
one suggestion if you like it, otherwise 👍

joecowton1 (approved):

PR #1595, opened 2022-10-13
Remove `!important` from `visuallyHidden`

joecowton1:

CSS containing `!important` [doesn't pass AMP validation on DCR](https://github.com/guardian/dotcom-rendering/actions/runs/3242851146/jobs/5316727377). 

This PR suggests a (potentially temporary) version of `visuallyHidden` that doesn't use `!important`.

If we decide we **need** to use `!important` we can always reintroduce that or discuss further, for now though the changes in this PR will unblock us so we're able to [bump Source in DCR](https://github.com/guardian/dotcom-rendering/pull/6178).

### DIFF
--- a/.changeset/large-mice-attack.md
+++ b/.changeset/large-mice-attack.md
@@ -1,0 +1,5 @@
+---
+'@guardian/source-foundations': patch
+---
+
+Remove !important from visuallyHidden
--- a/.changeset/large-mice-attack.md
+++ b/.changeset/large-mice-attack.md
@@ -2,4 +2,4 @@
 '@guardian/source-foundations': patch
 ---
 
-Remove !important from visuallyHidden
+Remove `!important` from `visuallyHidden`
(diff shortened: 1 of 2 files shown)


joecowton1:
> i think my feeling is that we should use something amp-appropriate for amp, rather than change our more general approach to fit one case.
> 
> this is the recommendation from DAC, i trust their motivations much more than AMPs.
> 
> appreciate this is to get source working with amp, but what if we didn't use it in amp instead? for this bit at least. is that possible?

We discussed this in CSTI standup and agreed that a reasonable approach would be to merge these changes for now and consider the issue outlined by Alex and in the description in isolation. 

https://github.com/orgs/guardian/projects/61/views/1

.changeset/large-mice-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+'@guardian/source-foundations': patch
+---
+
+Remove !important from visuallyHidden
```
sndrs, line 5:
```suggestion
Remove `!important` from `visuallyHidden`
```

sndrs (dismissed):

PR #1602, opened 2022-10-18
Use CSS from `focusHaloSpaced` for `ToggleSwitch` focus styles

joecowton1:

Uses edited version of `focusHaloSpaced` from `source-foundations` on `focus` rather than setting styles on `focus-visible` selector.

This brings the ToggleSwitch component focus styling into line with the rest of the Source components.


ob6160 (approved):

sndrs (approved):
why not

PR #1608, opened 2022-10-25
Storybook upgrade and tidyup in Source

ob6160:

- To upgrade and simplify Storybook so that it's easier to migrate into CSNX.

## What does this change?

- Upgrades Storybook to latest version v.6.5.13
- Removes the idea of a 'Playground' as this was brittle and broke when upgrading Storybook
  - Removes `asPlayground` and `asChromaticStory` methods
  - Update contributor docs to reflect changes
- Localises palette and typography Storybook config 
- Removes type fix for Emotion 10 as latest version us

## Screenshots

**Before**
image

**After**
image


package.json
```diff
@@ -18,12 +18,13 @@
...
 		"release": "yarn clean && yarn build && changeset publish",
-		"storybook": "start-storybook -p 6006",
 		"test": "nx run-many --all --target=test",
 		"test:coverage": "nx run-many --target=test --all --coverage",
 		"tsc": "tsc",
 		"validate": "npm-run-all tsc lint test build e2e",
-		"e2e": "nx run-many --target=e2e --all"
+		"e2e": "nx run-many --target=e2e --all",
+		"storybook": "start-storybook -p 6006",
+		"build-storybook": "build-storybook"
```
sndrs, line 27: think you need this name for the chromatic workflow to run correctly (or update the workflow)

```suggestion
		"build:storybook": "build-storybook"
```

packages/@guardian/source-foundations/src/colour/palette.stories.mdx
```diff
@@ -1,6 +1,6 @@
 import { Meta } from '@storybook/addon-docs';
 import { palette as colours } from './palette';
-import { ColorPalette, ColorItem } from '/.storybook/colorPalette';
+import { ColorPalette, ColorItem } from './storybookColorPalette';
```
joecowton1, line 3: What's the thinking behind moving these out of the `./storybook` folder?

ob6160: ah - so these are just being used in the `color` and `typography` mdx story respectively but are being included in the global storybook configuration.

the thinking was that if they are only consumed as a set of helper components/methods by those two mdx stories, they should probably live alongside them instead of at the global configuration level

joecowton1: My hunch is that this might be one to leave until you move source across to csnx - if `ColorPalette` and `ColorItem` are Storybook specific it makes sense for them to be in the `./storybook` folder. In the new setup there will be a `./storybook` folder per project so would make sense for it to live in there?

ob6160: I'm in two minds about whether they should live in `.storybook` though even if it is just scoped per project. I've always thought that directory was a place for just storybook configuration files to live 🤔 

These feel less like configuration and are more helpers just for those two specific stories, so in my mind they should be local to the two stories that consume them.

package.json
ob6160: ah - looks like we already have a `build:storybook` here: https://github.com/guardian/source/pull/1608/files/49ec530f5c1622432dff60004178d7dd8c491fce#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R12 

so I think safe to remove this line completely

packages/@guardian/source-foundations/src/colour/palette.stories.mdx
joecowton1: my feeling is that if they're only used in Storybook they should go in the `.storybook` folder - otherwise you're implying that they're used in the code itself. not a game changer either way though.

ob6160: I see your point there, we called it `storybookColorPalette` to indicate that it belongs to storybook. I suppose individual stories have `.stories.tsx` to differentiate them from just source code, so perhaps we could give the file its own custom extension as well to further strengthen that differentiation?

The thinking in my mind is that we don't locate stories in `.storybook`, so why should we locate their helper methods in there? It feels like we're losing something by not making the two files local to the two stories that they directly support.

joecowton1 (approved):

packages/@guardian/source-foundations/src/colour/palette.stories.mdx
joecowton1: thats a fair point about .stories files. i was think basically making them local is something that could be done down the line Source is in csnx but not really worth debating too heavily.

PR #1628, opened 2022-12-01
Expanding Wrapper should manage the focus of child nodes when the wrapper is collapsed

sophie-macmillan:
Previously the expanding wrapper wasn't managing the focus of any children, which meant this had to be managed by the parent. If this was not managed, it was possible to tab through the interactive elements within the collapsed ExpandingWrapper, which was a bad ux experience (see video).

Now, these elements are non-tabbable when the wrapper is collapsed by default (although the disableTabbingWhenCollapsed=false prop can be passed to prevent this).

**Before**
This was possible as the component wasn't managing focus (the button isn't scrolled into view, but is focused and on pressing space triggers the alert)

https://user-images.githubusercontent.com/26366706/205097996-ac1a0bb2-95b7-41a6-bbbe-581cb3ebe3b5.mov

**After**
By default, the component manages focus so that the children are only tabbable when the component is expanded

https://user-images.githubusercontent.com/26366706/205098316-7c6589fd-ffab-456f-9c4c-ca9ea9fba450.mov

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -60,7 +60,7 @@ module.exports = {
 			(rule) => String(rule.test) === String(/\.(mjs|tsx?|jsx?)$/),
 		).exclude = nodeModulesExclude;
 
-		config.resolve.plugins ||= [];
+		config.resolve.plugins = [];
 		config.resolve.plugins.push(
 			new TsconfigPathsPlugin({
 				configFile: path.resolve(__dirname, '..', 'tsconfig.json'),
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -60,7 +60,7 @@ module.exports = {
 			(rule) => String(rule.test) === String(/\.(mjs|tsx?|jsx?)$/),
 		).exclude = nodeModulesExclude;
 
-		config.resolve.plugins = [];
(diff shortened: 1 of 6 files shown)


.storybook/main.js
```diff
@@ -60,7 +60,7 @@ module.exports = {
 			(rule) => String(rule.test) === String(/\.(mjs|tsx?|jsx?)$/),
 		).exclude = nodeModulesExclude;
 
-		config.resolve.plugins ||= [];
+		config.resolve.plugins = [];
```
joecowton1, line 63: Do we want to be changing this here?

tkgnm:
A couple of minor design bits on the "show more" button: 

 Below I've added a screenshot of the figma from Ben for comparisons sake

image

image

1) The show less button needs less vertical padding (and maybe a TINY bit less horizontal padding but I am not sure)
2) The button needs a left margin (8px or so I think)

.storybook/main.js
sophie-macmillan: We do not! No idea how that got in there!

tkgnm (approved):

tjsilver (approved):
👍

